PR #38, opened 2019-05-27
added check for vtk dir variable

jessdtate:


cchriste:
merged with consolidate_submodules_master

PR #40, opened 2019-07-08
Fix VTK include, add ccache.

akenmorris:
Fixes VTK include (revert to what's in master).

Also add ccache support (where available).


cchriste:
Hi Alan, I added instructions and fixed things to work with latest OSX (Mojave). Please review and go ahead and merge if it looks good and still works for the older version of OSX you're using. Thanks!

PR #42, opened 2019-07-11
Changes needed for build for windows (using linux subsystem for windows)

akenmorris:
Included is a superbuild file for windows using the linux subsystem for windows as well as changes needed to build.


cchriste:
It looks like the build script added is for the old version w/ submodules, not this consolidated version.

akenmorris:
That's right.  Until it's in master, I'm not sure what to do.  Should I change this script to check out your merge_improv_etc branch?

PR #79, opened 2019-08-22
Superbuild, build instructions, conda_installs, all significantly improved

cchriste:


cchriste:
--no-gui. Thanks for catching the conflict in the help/build instructions. Fixed.

cchriste:
apt-install wasn’t complete, so I removed it so as not to create confusion, and instead added deps.txt so people can install whatever it necessary. Also, there are multiple methods that can be used to manually install dependencies, such as apt-get, zypper, homebrew (on osx), etc.
I also added a link to deps.txt in the BUILD.md. It’s not super legible in markdown, but I didn’t want to simply cut and paste these deps into the markdown since that will invariably go out of sync. It’s readable and the file can be loaded from the command line.
Also, if the deps are actually different when installed manually I would be surprised, so what apt-installs actually installed just happened to be what was missing from its creator’s machine, not the actual required deps for ShapeWorks.
This also relates to the Windows instructions, but for these I don’t have a system with which to experiment, so I didn’t modify anything.

PR #86, opened 2019-08-23
ShapeWorksStudio merge

akenmorris:
This PR bring ShapeWorksStudio as it was into the regular ShapeWorks build.  This means upgrading it to VTK 8, ITK 5, and integrating into the ShapeWorks build system.

The CMAKE option to build it is -DBuild_Studio=ON

The superbuild option to build it is --with-studio

The next step will be to start swapping out the backend stuff.


cchriste:
Looks good.

PR #90, opened 2019-08-25
Surface reconstruction shared libraries

sheryjoe:
This PR moves the commons libs for particle-based reconstruction to Libs so that they can be used as shared libs for both command line tools and viz tools (View2 and Studio).


cchriste:
I'll compile this with and without Studio just to make sure it works.

akenmorris:
I don’t think Studio is using this yet.

Alan

cchriste:
A pull request should first merge master:

cchriste:
This doesn't compile on my system because it doesn't know what language an hxx file indicates and the cmake addition is apparently ignored on OSX. Furthermore, let's not use .hxx files at all, but instead consistently use .h for our headers. They add an unnecessary layer of confusion and the compiler treats them the same. 

Another issue is that the files in Transforms were cut-and-pasted from ITK more than a decade ago and ITK has changed quite a bit, likely adding whatever functionality wasn't there that inspired their copying in the first place. If it's still not there but we need the primary functionality, let's create a derived class from the main ITK class (e.g., itk::AdvancedTransform) instead.

PR #91, opened 2019-08-25
Reorg part 1

cchriste:
- move non-ShapeWorks Libs to ExternalLibs
- remove ITKVTK and instead use version included with ITK, remove duplicate TinyXML
- remove Libs/fim_v4 CMakeFiles (generated by cmake, shouldn't be in git)
- remove duplicate PreViewMeshQC library (identical diffs)
- cleanup a bunch of CMakeLists stuff, but plenty more to go

There is plenty left to do, but I would suggest even if we don't merge this branch that everyone start working based on this branch rather than master, mostly due to the deleted duplicate code and demonstration of how to properly include a library (e.g, we should never explicitly tell cmake where to find its includes).


cchriste:
Merging this branch at this time seems like a positive step forward, but it is definitely not a finished reorg. If we do not merge, please base your changes on this branch until it's ready. Otherwise, it will introduce unnecessary (and painful) complications when we try to merge. Reorganization _is_ an incremental process you know :)

PR #126, opened 2019-09-04
Superbuild fixes and improved build instructions

cchriste:
Discovered some issues when one of our users was unable to build successfully. Now it checks that the anaconda Qt isn't being used, properly uses the build dirs, and requires qmake to be in the path.


akenmorris:
Why does Qt have to be 5.10? Are you just trying to exclude anaconda?

Personally, I've been using Qt 5.9.8 (latest release of last LTS version) on most projects.  I have found that newer Qt versions have lots of bugs on Mac, at least.  I'm not sure about 5.10, but 5.12 is terrible.

cchriste:
Unfortunately, that's exactly why I set the minimum version to 5.10 (as noted in the comment in the code).

PR #165, opened 2019-09-26
Update SurfaceReconstruction in Studio

akenmorris:
This PR Updates Studio to use the new Libs/SurfaceReconstruction instead of its own older copy of SurfWorks.


aogs:
Is this related to replacing Studio’s SurfWorks with Libs/SurfaceReconstruction?  Or is it new functionality?  If so, let’s merge this and make a separate issue.

sheryjoe:
In part, yes as SurfaceReconstruction is giving this good/bad particles. Yet, it is worth to have it on a different branch as it would require adding new interface components to Studio.

PR #175, opened 2019-09-26
Use case la

AtefehKashani:
This branch is ready to merge to master. All files are unified for both use cases.


AtefehKashani:
> This relies on:
> 
> filename="/home/sci/atefeh.gk/Public/leftatrium.zip"
> 
> Until we have girder setup, should we place this on sci.utah.edu somewhere (or github) and do a "wget/curl" to get the file?

We can have two options here, 1. Since the data is open access, I can add a ReadData module to download the data from the link and prepare it for the use case. 
2. As you mentioned we can place it somewhere and download files. 
which one do you recommend?

aogs:
ReadData?  Do you mean have python download it?  I think that’s the better option as Windows won’t have wget/curl.

AtefehKashani:
> ReadData? Do you mean have python download it? I think that’s the better option as Windows won’t have wget/curl.

Yes, a python module for download the data and provide it for the use case.

aogs:
Yep, that sounds good.

AtefehKashani:
> Yep, that sounds good.

I will add this module by end of today and commit after testing.

PR #202, opened 2019-10-04
Move along pca fixes

sheryjoe:


Prep/Source/MeshTools/ParticleBasedSurfaceReconstruction/source/ReconstructSamplesAlongPCAModes.cxx
```diff
@@ -77,31 +77,44 @@ int DoIt(InputParams params)
...
+                {
+                    NumberOfModes = n;
+                    found = true;
+                }
+            }
 
-    if(!found)
-        NumberOfModes = percentVarByMode.size();
+            if(!found)
+                NumberOfModes = percentVarByMode.size();
```
prafulag: why is this needed after line 101 ?

prafulag: is it just a fail safe for 101 ?

Prep/Source/MeshTools/ParticleBasedSurfaceReconstruction/source/ReconstructSamplesAlongPCAModes.cxx
```diff
@@ -130,17 +143,28 @@ int DoIt(InputParams params)
     {
```
prafulag: This loop may fail when we provide the particular mode index, unless the mode to be generated is very first mode, since NumberOfModes is set to 1.

Prep/Source/MeshTools/ParticleBasedSurfaceReconstruction/source/ReconstructSamplesAlongPCAModes.cxx
sheryjoe: yes

Prep/Source/MeshTools/ParticleBasedSurfaceReconstruction/source/ReconstructSamplesAlongPCAModes.cxx
sheryjoe: fixed

prafulag: I don't think it is fixed yet, coz the NumberOfModes is set to 1, so loop runs only one time. The continue condition was definitely needed, but the NumberOfModes need to be set to max_possible_number_of_modes. Since we already have the flag for singleModeToBeGen, number of modes need not be set to 1.

PR #216, opened 2019-10-14
Consolidate psm

akenmorris:
This PR pulls out the ITKParticleSystem out of ShapeWorksRun, moves it to Libs/ParticleSystem.

Studio's Optimization code has been re-targeted to use this library rather than the ITKv4 PSMFilter based optimization code.  Removed now non-existent options from UI and added new options (such as optimization iterations). 

ShapeWorksRun5.0 is renamed to ShapeWorksRun

SW_USE_MESH and SW_USE_FEAMESH have been removed and the relevant sections have been removed or preserved.

Removed unused Run 'Utilities' library.

Removed ShapeWorksCLT.


prafulag:
Few things -
1. Renaming the binary - As discussed earlier, it could be better to keep some kind of version number associated with the binaries OR we can tag the commits/merges to master based on some versioning.

2.Removing SW_USE_MESH and SW_USE_FEA_MESH - the main thing to note here is reading the inputs. I would like this to be tested before being merged into master. The simple tests to do can be - run shapeworks on sphere/ellipsoid data with a) xyz b) normals c) xyz and normals. Sphere data will be recommended, since normals can establish a good correspondence in spheres without using xyz.

aogs:
Definitely tag releases in GitHub

Can you help me with testing those options?  I’m happy to run it, but might need some direction.

prafulag:
Sure, do you have access to sphere one mode data ?

aogs:
Doesn’t ring a bell.

prafulag:
Please find the prepped data at /home/sci/prafulag/Public/ForAlan

akenmorris:
Got it.  I see, spheres of different sizes.

So I use combinations of "use_normals" and "use_xyz" flags?

prafulag:
Yes, also turn on the mesh_based_attributes flag.
The goal is to verify that you should not need to provide meshes as input for these options as long as no other feature_attributes are being used.

cchriste:
I agree with Praful about versioning using tags rather than the name of the executable. We can and should tag as often as we like. Major versions, minor fixes, specifically-named "update-the-whizbanger" changes, etc.

cchriste:
I have built this using a fresh run of superbuild.sh (maybe my last!), ran Studio successfully, and executed the Python ellipsoidTest script (which required installing the SimpleITK conda module).

prafulag:
> I have built this using a fresh run of superbuild.sh (maybe my last!), ran Studio successfully, and executed the Python ellipsoidTest script (which required installing the SimpleITK conda module).

Thanks Cameron. Did you have mesh_based_attributes turned on in your ellipsoid experiments ?

cchriste:
> > I have built this using a fresh run of superbuild.sh (maybe my last!), ran Studio successfully, and executed the Python ellipsoidTest script (which required installing the SimpleITK conda module).
> 
> Thanks Cameron. Did you have mesh_based_attributes turned on in your ellipsoid experiments ?

No idea. So probably not. :)

PR #221, opened 2019-10-16
Cmake superbuild

cchriste:
Now cmake brings in the external libraries and packages the Studio app rather than using a separate script.

I anticipate this should be merged with a prior pull request (consolidate_psm) first, but if that one is held up we may as well push this one first.


akenmorris:
cmake I was using on one of the SCI linux machines complained:

CMake Error at CMakeLists.txt:108 (cmake_policy):
  Policy "CMP0072" is not known to this version of CMake.

Do we need to change the cmake_minimum_required?

akenmorris:
On kouroush, I am getting:

CMake Error at Post/source/ShapeWorksPost-V1/external/nanogui/ext/glfw/CMakeLists.txt:259 (message):
  The RandR library and headers were not found


-- Configuring incomplete, errors occurred!


I had re-run "source ./conda_installs.sh" and started with a clean build.  I am not sure why it doesn't find it.

AtefehKashani:
--These are some minor editions:
1. List of dependencies should be updated to have needed libraries for python scripts; SimpleITK and requests 
2. BUILD.md is not completed for CMAKE and needs some edition:
""mkdir build
  cd build
  cmake ..
  make
""
it could be helpful if we have an example for how to make with options. The same as what we have for superbuild.

Issue #190, opened 2019-10-03
All test/use case datasets should be downloaded from Girder rather than hard coded path


akenmorris:

cchriste:
Examples currently requiring this:

- LA Use Case needs to download data rather than hard coded path #190
- ellipsoid test use case
- sphere use case  

Oleks will help to ensure data is accessible via Girder.

strikerz76:
@AtefehKashani Is the data currently in /home/sci/atefeh.gk/Public/leftatrium.zip the namic data? If so can you add a README with a link to the source of the data and a copy of the license info?

strikerz76:
@cchriste Where is the data for the sphere use case? Also I'm not seeing a sphere use case script.

AtefehKashani:
> @AtefehKashani Is the data currently in /home/sci/atefeh.gk/Public/leftatrium.zip the namic data? If so can you add a README with a link to the source of the data and a copy of the license info?

@strikerz76 yes, this data is the NAMIC data, that I have downloaded and prepared for the use case. http://insight-journal.org/midas/collection/view/197
what should be in the README file?

strikerz76:
I'm just going to copy the README on their website into the data folder.
New structure of data folder:
leftatrium.zip 
-> segmentation_LGE
-> README.txt

AtefehKashani:
> I'm just going to copy the README on their website into the data folder.
> New structure of data folder:
> leftatrium.zip
> -> segmentation_LGE
> -> README.txt

Sounds good!

strikerz76:
#226

PR #226, opened 2019-10-17
Girder Connection for ellipsoids, leftatrium, sphere data

strikerz76:
Added rough python script for querying the girder api (GirderConnector.py)
Modified ellipsoidMain.py and LAMain.py to download the data from girder using the connector if the data is not found locally.
The connection to girder currently uses an account with username shapeworks and the account hashed credentials are currently hardcoded in the script.
The authentication is Basic http auth but should eventually move to oauth.

#190


AtefehKashani:
I have tested the ellipse and LA examples. They work fine.
These scripts need "requests" and it is not in the list of dependencies, it should be added to conda_install.

PR #239, opened 2019-10-24
Fix conda_installs.sh to work cross platform

strikerz76:
Updated if statement that checks if conda is installed to work on Linux, OSX, WSL.
Get requests from conda instead of pip

Test that running source ./conda_installs.sh the first time installs miniconda and running it the second time skips the miniconda install.
WSL(Ubuntu): tested
Suse Linux: TODO
OSX: TODO


cchriste:
Note that this also changes conda_installs to use matplotlib from pip instead of conda so qt isn't installed. And it includes instructions and fixes to use CMAKE_PREFIX_PATH for when you want to specify your own installation of some library such as Qt, vtk, ITK, etc.

PR #240, opened 2019-10-24
Use case la

AtefehKashani:
The LAMain is changed to be consistence with ellipsoid example and also the ellipsoid optimization parameters are changed.


cchriste:
tested ellipsoidMain and LAMain and they both worked

PR #241, opened 2019-10-25
Use case la

AtefehKashani:
The subprocess.check_call is edited for all executable commands. 
Run pipeline function is created for ellipsoid and left atrium use cases, and the Main Scripts are edited accordingly.
These changes are tested for both use cases.


cchriste:
This hadn't yet been merged with origin/master. I merged and tested and it looks great.

Issue #89, opened 2019-08-25
upgrade to latest trimesh2 library using conda


cchriste:
This external library should (probably) be updated and moved to our external libs directory.

Used by fim_v4.
Original source found here: https://gfx.cs.princeton.edu/proj/trimesh2/

cchriste:
Instead of moving to ExternalLibs, this library should be added using cmake's ExternalProject_Add command like vtk, ITK, etc.

sheryjoe:
Agreed. But we need to extract our changes/additions to this lib to a class (i.e. internal lib). This also entails ensuring that this internal lib is compatible with the latent trimesh lib. @prafulag can help/advise on this.

cchriste:
> Agreed. But we need to extract our changes/additions to this lib to a class (i.e. internal lib). This also entails ensuring that this internal lib is compatible with the latent trimesh lib. @prafulag can help/advise on this.

Yes, this is already mentioned in the comment from Aug 26.

cchriste:
If the map/reduce classes are our creation vs a presumably upgraded part of trimesh, here are instructions on how to update threading with ITK 5: https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/ITK5MigrationGuide.md

cchriste:
Instead of providing a copy of the library, we should install the conda version:
`https://anaconda.org/conda-forge/trimesh`

(title updated accordingly)

0leks:
[trimesh](https://github.com/mikedh/trimesh) is different from [trimesh2](https://gfx.cs.princeton.edu/proj/trimesh2/)
I don't think trimesh2 can be installed via conda. I don't even know if there is a git repo for it... (I haven't found one)

cchriste:
Ah. Okay then, return to upgrade.

cchriste:
This version might be the best. It's updated regularly and contains Windows support:
https://github.com/Forceflow/trimesh2

0leks:
That looks nice, I'll  check it out

cchriste:
What is in ExternalLibs/Trimesh2? 
If this is a clone of the Forceflow repo, can we add a README and close this issue? 
Or are there ShapeWorks-specific modifications? 
Do we still have another version (I don't see one)?
Are the changes from our originally-modified version still here, or were they moved to another library?
Thanks!

0leks:
[...]

PR #242, opened 2019-10-28
Reorganize and consolidate libraries

cchriste:
Consolidated libraries as we discussed at our meeting last Thursday, now we have:
**Alignment/
Analyze/
Image/
Mesh/
Optimize/
Utils/**

Have built from scratch to ensure everything works, tested our guis and latest python demonstrations. Let's please get this merged ASAP since any work may be on files that have moved and this could be confusing or more challenging to merge as time goes on. It's a solid, stable branch with appropriate directory structure and clarity. Thanks!


cchriste:
Thanks, Shireen. I did `make clean` but apparently it still keeps things
around like old libs. For example, -lFoo will look for the library named
Foo and fail if it doesn't exist. I thought I'd ensured our old libs were
no longer being requested but `make clean` wasn't sufficient. So now I've
completely deleted the build/shapeworks directory to ensure nothing is
there that's not supposed to be. Ready to merge!

sheryjoe:
To be addressed before the merge:
- TriMesh (fim_v4) should be in Libs till we extract and isolate our additions to this library.
- Examples should have Scripts (containing .sh scripts currently in Examples) and Python (containing use cases) folders.
- Another name for Utils.* to avoid conflict with other libs utils
- Inconsistencies in Libs/XX/"sublibs" - example, Libs/Mesh/PreviewMeshQC verus Libs/Reconstruction.* 

The below could be addressed after this merge in different branch(s):
- File extensions consistency (txx vs cxx vs cpp)
- Get rid of "itk" prefix in file names and classes (esp the particle system)
- Command line tools in Libs to be converted to classes with calling APIs for cmd/studio/cloud tools.
- consolidate shapeworks executable
-

cchriste:
> To be addressed before the merge:
> 
> * TriMesh (fim_v4) should be in Libs till we extract and isolate our additions to this library.
Nothing should be in ExternalLibs, so everything there is meant to be pulled in by cmake's ExternalProject_Add. I'm aware of our Trimesh modifications and putting this lib in ExternalLibs is the first step towards resolving issue #89, mentioned in several other places, including the commit that moved trimesh.

> * Examples should have Scripts (containing .sh scripts currently in Examples) and Python (containing use cases) folders.
Done.

> * Another name for Utils.* to avoid conflict with other libs utils
Can this please be done after the merge? I need to commit time to another project and the Utils is as it has been all along. Just like the itk-prefixed files mentioned below, not really a reorganization task.

> * Inconsistencies in Libs/XX/"sublibs" - example, Libs/Mesh/PreviewMeshQC verus Libs/Reconstruction.*
I would argue all files for a given library should be in its top-level directory, but some of these consolidations involved lots of files (another example is the [itk]ParticleSystem files. They could be moved up, but perhaps also reduced to fewer files. As it stands, this is an incremental process and this first step seems reasonable. I would strongly argue _against_ creating a zillion subdirectories within each lib, but a few might be useful. Again, this is a good start and can be incrementally improved. 

> 
> The below could be addressed after this merge in different branch(s):
> 
> * File extensions consistency (txx vs cxx vs cpp)
Totally agree! Same for cxx and cpp, which should all be cpp.

> * Get rid of "itk" prefix in file names and classes (esp the particle system)
I started this and it is a big task so I feel it should be addressed incrementally after this merge is complete.

> * Command line tools in Libs to be converted to classes with calling APIs for cmd/studio/cloud tools.
> * consolidate shapeworks executable
Yup!

cchriste:
> * UseCaseDocumentation should be renamed to Python and only include shapeworks python library. Use cases should be moved to Examples/Python and use the lib in ShapeWorks/Python. Pdfs in UseCaseDocumentation should be moved to Documentation.

I wasn't sure their purpose since those pdfs were added recently. I agree they should go in Documentation.

I agree we should have a Python folder for the ShapeWorks Python API, which doesn't yet exist. So all our examples are in Examples as they should be. If a different organization is desired, perhaps it would best be done by the people closer to these scripts _after the merge_.

akenmorris:
Regarding CMAKE_BUILD_TYPE. I found inconsistent values for this across the different CMake files.

cchriste:
I've seen Alan's comments on Slack (ellipsoid, studio and view2 are all working). Can we go ahead and merge now?

cchriste:
Everything seems good to go so I’m going to merge now.

Issue #243, opened 2019-10-29
ITK5 threading migration issues


akenmorris:
We missed a couple of things in the ITKv5 migration, in particular, the threading:

https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/ITK5MigrationGuide.md

As a result, we are getting crashes in GenerateFidsFilesFromMeshes in either MapFilter.h or ReduceFilter.h in trimesh.  I can't tell which one because someone incorrectly called both of them "MeanFilter" in the itkTypeMacro.

ShapeWorks/install/include/ITK-5.0/itkImageSource.hxx:280:
itk::ERROR: MeanFilter(0x10ea1b10): Subclass should override this method!!! If old behavior is desired invoke this->DynamicMultiThreadingOff(); before Update() is called. The best place is in class constructor.

PR #244, opened 2019-10-29
Disable ITK5 dynamic multithreading for MapFilter and ReduceFilter.

akenmorris:
Also fix ITK names.

This should fix the crash in GenerateFidsFilesFromMeshes (issue #243)


akenmorris:
I have confirmed that GenerateFidsFilesFromMeshes no longer crashes.

PR #245, opened 2019-10-30
Studio Rearrangement

akenmorris:
This PR updates ShapeWorksStudio in several ways:

1) Rename Analysis to Analyze
2) Change stylesheets so that side toolbar items have consistent size
3) Move Surface Reconstruction from Optimize to Analyze
4) Reorganize Analyze tool user interface to match View2.  Added remaining View2 UI elements (not all enabled yet).


AtefehKashani:
I have tested the studio with groomed data(distance transform) and unprepped data for both ellipsoid and NAMIC data. The only thing that I faced is that, the studio cann't optimize without saving the project. Alan found that it is because of ITKParticleSystem. It works fine after saving the project.

akenmorris:
Atefeh, with the latest commits, you should be able to optimize without saving now.

AtefehKashani:
I have tested the new commits  with groomed data(distance transform) and unprepped data for both ellipsoid and NAMIC data. It works with and without saving the project.

PR #247, opened 2019-11-03
Reorg girder connector

strikerz76:
If you rerun conda_install.sh on a current build it should install the DatasetUtils package as a local package.
Can someone also try installing everything from scratch on linux to make sure the DatsetUtils package gets installed properly.


akenmorris:
Confirmed that "DatsetUtils package gets installed properly."

akenmorris:
Worked for me, downloaded from Girder after deleting zips.

PR #249, opened 2019-11-04
Studio: Live optimize, legacy reconstruction and view2 support

akenmorris:
* Adds live particle position tracking
* Adds "stop optimization" button
* Fixes crashes on exit during optimizations (clean shutdown of threads)
* Adds legacy surface reconstructor
* Adds support for legacy View2 files
* Fix camera clipping range issues


akenmorris:
@cchriste , that's only in the windows branch right now.

Issue #263, opened 2019-11-21
Procrustes is not working


akenmorris:
<img width="765" alt="Screen Shot 2019-11-21 at 1 47 35 PM" src="https://user-images.githubusercontent.com/1693349/69375424-8755d180-0c65-11ea-98af-49098db6b14d.png">

These spheres should all be the same size.

akenmorris:
Fixed in "fix_procrustes" branch:

<img width="762" alt="Screen Shot 2019-11-21 at 1 53 18 PM" src="https://user-images.githubusercontent.com/1693349/69375793-5033f000-0c66-11ea-95df-7c5a1396f8d4.png">

PR #264, opened 2019-11-21
Fix procrustes

akenmorris:
The diffs associated show all uses of SetComputeTransformationOn/Off since I have removed them and it all compiles.

### DIFF
--- a/Libs/Optimize/ParticleSystem/itkParticleProcrustesRegistration.cxx
+++ b/Libs/Optimize/ParticleSystem/itkParticleProcrustesRegistration.cxx
@@ -21,11 +21,6 @@ template<>
 void
 ParticleProcrustesRegistration<3>::RunRegistration(int d)
 {
-    /* PRATEEP */
-    // Do not compute any transformation if already computed.
-    if(m_ComputeTransformation == false) return;
-
-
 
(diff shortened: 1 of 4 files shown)


akenmorris:
Results shown here:

https://github.com/SCIInstitute/ShapeWorks/issues/263

For multiple domains, do you suggest the same spheres for each domain?  Can you be more specific?

Procrustes should only operate on each domain separately, correct?

prafulag:
> For multiple domains, do you suggest the same spheres for each domain? 
> 
Yes
> Procrustes should only operate on each domain separately, correct?
> 
Yes

akenmorris:
It seems to work fine.  I ran with differing numbers of particles to be sure I was getting the results I wanted (separate domains), but View2/Studio cannot currently handle this situation, so viewing them separately, procrustes works fine:

First domain, 128 particles:

Screen Shot 2019-11-22 at 12 26 18 AM

Second domain, 64 particles:

Screen Shot 2019-11-22 at 12 26 05 AM

cchriste:
It builds fine and the code looks cleaner, but I'm sure about why the pieces removed were there in the first place.

prafulag:
Results look good to me! Great work Alan!

PR #273, opened 2019-12-10
Fix #272 - Fix non-Qt build.

### LINKED ISSUE #272: CMake configure requires Qt even if GUI not specified
CMAke is requiring Qt even if no GUI components are specified.

akenmorris:
Praful, can you try this branch?


prafulag:
> Praful, can you try this branch?

I cloned the branch (fix_non_qt_build) and tried to build with default (GUI OFF) and still getting the same issue:

CMake error at CMakeLists.txt: 238 (get_target_property):
 get_target_property() called with non-existent target "QT5::Widgets".

CMake error at CMakeLists.txt: 239(get_filename_component):
 get_filename_component called with incorrect number of arguments

akenmorris:
Thanks for checking Praful.  I'm not sure why my linux system without Qt doesn't complain.  In any case, I've moved those remaining steps that use Qt inside a IF guard.

Can you please try again?

prafulag:
It now builds successfully!

PR #276, opened 2019-12-12
Femur use case

jadie1:
Adding complete femur use case to /Examples/Python (normal samples only)


Examples/Python/RunFemur.py
```diff
@@ -0,0 +1,574 @@
...
+import sys
+import csv
+import argparse
+import glob
+
+from GroomUtils import *
+from OptimizeUtils import *
+from AnalyzeUtils import *
+
+from GirderConnector import downloadUseCaseData
```
akenmorris: I believe this has been replaced with:

from DatasetUtils import datasets
datasets.downloadDataset(filename)

I get an error looking for GirderConnector.  I am assuming it works for you because have the older one installed with pip.

akenmorris:
Next, I get:

  File "FemurMain.py", line 27, in <module>
    Run_Femur_Pipline(args)
  File "/Users/amorris/sci/shapeworks/Examples/Python/RunFemur.py", line 260, in Run_Femur_Pipline
    resampledFiles = applyIsotropicResampling(parentDir, fileList_seg, None, 1)
UnboundLocalError: local variable 'fileList_seg' referenced before assignment

Should I be specifying either `--start_with_prepped_data` or `--start_with_image_and_segmentation_data`?

akenmorris:
I tried with `--start_with_prepped_data` and got the same error.

Examples/Python/RunFemur.py
jadie1: Hi Alan, yes this should be replaced. I apologize I've had the zip file downloaded for a while so I didn't realize the Girder code had been changed.

akenmorris: Yep, no problem.  I've committed a fix.

sheryjoe:
A side note, can we consolidate md files in a documentation folder and only maintain the readme.md in the repo's root dir with links to those individual md files? 

@jadie1 please update the main readme to provide the link to the femur use case md file.

jadie1:
I am adding some functionality for the cutting plane, I will reopen this pull request when that is done.

PR #280, opened 2019-12-20
added versions to all the prerequisites installed by conda_installs.sh

cchriste:
This will ensure old versions will still work instead of making it a guessing game to get the correct (old) versions of the prerequisites.

(didn't modify .bat version since I don't have Windows handy to check availability of versions)


akenmorris:
I'm going to guess the windows ones will be the same.

PR #282, opened 2019-12-30
Only allow centering when on original meshes.

### LINKED ISSUE #279: Studio visualization artifact with procrustes
When procrustes is on for optimization, the visualization of the samples are rendering incorrectly in studio analyze. Shireen suspects this is because the scaling has changed the coordinates. 
The same samples render correctly in View2, here are some examples:

Studio_Sample0
View2_Sample0
Studio_Sample1
Vew2_Sample1

akenmorris:
Fixes #279.

### DIFF
--- a/Studio/src/Application/Visualization/Visualizer.cc
+++ b/Studio/src/Application/Visualization/Visualizer.cc
@@ -112,7 +112,7 @@ void Visualizer::display_samples()
     annotations << "";
     object->set_annotations(annotations);
 
-    if (this->center_ && mesh) {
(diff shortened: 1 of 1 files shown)


jadie1:
Now when I open studio on the analyze step, the centering checkbox is still visible but it appears to be disabled, I don't see a change in the samples whether it is checked or unchecked. Was this what was intended or was the checkbox supposed to be removed entirely?

akenmorris:
I decided to leave the checkbox, but just have it ignored when on groomed/reconstructed surfaces.

Sounds like it's working ok.

PR #293, opened 2020-01-10
Mac installer

akenmorris:
This PR modifies the Mac build to create both a zip file and a pkg installer.  A new README describes how to use the pkg installed version.

I tested both methods on my machine, but it would be great if anyone else could as well.

### DIFF
--- a/.github/workflows/build-mac.yml
+++ b/.github/workflows/build-mac.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - master
-      - mac_installer
   pull_request:
 
 jobs:
--- a/.github/workflows/build-mac.yml
+++ b/.github/workflows/build-mac.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - mac_installer
   pull_request:
 
 jobs:
--- a/Examples/Python/LAMain.py
+++ b/Examples/Python/LAMain.py
@@ -17,6 +17,8 @@ import platform
 binpath = "../build/shapeworks/src/ShapeWorks-build/bin:../../bin"
 if platform.system() == "Windows":
     binpath = "C:\\Program Files\ShapeWorks\\bin"
+if platform.system() == "Darwin":
+    binpath = binpath + ":/Applications/ShapeWorks/bin"
 
 parser = argparse.ArgumentParser(description='Example ShapeWorks LA Pipeline')
 parser.add_argument("--interactive", help="Run in interactive mode", action="store_true")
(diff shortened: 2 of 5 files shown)


akenmorris:
The two downloads can be found here:

https://github.com/SCIInstitute/ShapeWorks/suites/394889156/artifacts/1058037

cchriste:
ShapeWorksStudio is not included with the packaged installer.

cchriste:
Takes maybe 20-30s to start each executable the first time, presumably because every single linked .dylib must be independently approved (under the hood in this case, but still so slow).

cchriste:
...and not all executables are working due to (you guessed it) missing Qt libs:
```(shapeworks) cam@ananda:/Applications/ShapeWorks/bin$ ./SmoothMesh 
dyld: Library not loaded: @rpath/QtSql.framework/Versions/5/QtSql
  Referenced from: /Applications/ShapeWorks/bin/./SmoothMesh
  Reason: image not found
Abort trap: 6
```

cchriste:
The .zip version still requires all the approvals mentioned earlier, so basically impossible to use.

All I did was click the link from Safari, which unzips as it downloads (which I hate), then unzipped the .zip file in the archive from the command line. So not sure if gatekeeper is that smart or not.

akenmorris:
Try here:

http://www.sci.utah.edu/~amorris/ShapeWorks-v5.2.2-rc2-54-g3557ea58-mac.pkg

I confirmed uncompressing with xar that it contains ShapeWorksStudio.app, also installs on my machine.  If it's refusing to install on yours (Catalina?) we'll have to figure out why or abandon the installer idea.

akenmorris:
I just tried it on a Mojave system and had no issue.  I don't have access to any systems running Catalina, so I'm not sure if that's the issue.

akenmorris:
You might have to take over this task.  I don't have a Catalina system with the enhanced GateKeeper to try anything on.

cchriste:
I think the installer is a great idea, especially for ShapeWorksStudio, which can be installed directly into the /Applications folder, and which works "out of the box" since it uses the Qt bundler and doesn't have the RPATH issues of the individual executables. 

What I think we can do for the command line tools is to add is a link that opens Terminal with the conda environment activated.

When I run `xar -x -f ./ShapeWorks-v5.2.2-rc2-54-g3557ea58-mac.pkg` is creates `Bom`, `PackageInfo`, and `Payload`, but I haven't looked into how to use them. `PackageInfo` does list studio as part of the contents.

As far as taking over this task, my list of priorities has many things ahead of it, so if you want to create an issue and assign me that's fine, but please don't hold your breath it will get done anytime soon (probably not even before the next users meeting). Also, I'll need some instructions on what you've done so far to create the installer.

In the meantime I'll just merge this pull request since it's a step in the right direction.

akenmorris:
To open the payload file:

cat Payload | gunzip -dc |cpio -i

Support for Catalina will have to wait until either I get access to such a system (can't say I'm interested in upgrading any machine I care about) or you have some time to look at it.

PR #294, opened 2020-01-10
Fix errors in ellipsoid example

medakk:
Small changes for missing variables in Python script.

Probably unrelated here, but the Python script crashed at a later step because ShapeWorks Studio was missing: the default option in CMake is to not build Studio.


akenmorris:
Great!

PR #295, opened 2020-01-11
More testing

akenmorris:
Reorganize Testing structure into Testing folder, switch to git-lfs for nrrd files, add ImageTests with test for resample.


cchriste:
I'm trying to test and git-lfs us used to store the testing data. Unfortunately I can't remember how to use it and I can't find any documentation to explain it. Is there something I'm missing? Due to this, the tests themselves fail.

cchriste:
```git lfs fetch

akenmorris:
We can switch to operator== if you want.  Will it be confusing if someone wants to use it with a (smart) pointer to an Image?

akenmorris:
I'm not a super git lfs expert by any means.  I believe you only have to do those steps because you're adding it to an existing clone.  If the git-lfs stuff was merged and you had it installed, I believe it would be seamless.

PR #298, opened 2020-01-13
updates to progress with executable

cchriste:
- read/writeimage (and read/writemesh) are now themselves subcommands, significantly simplifying the use of the shapeworks command both direction and in the Python examples.
- separated recenter and isbinary portions of resample command so that users can perform these actions deliberately if desired
- make all arguments lowercase (still deliberately using CamelCase for Command classes themselves)
- added doxygen-style docs to the functions in Image class
- updated all examples
- added DEBUG_CONSOLIDATION flag to CMakeLists.txt for continuing work on these commands


sheryjoe:
isbinary should not be dropped from resampleimage,  resampleimage should handle both binary (segmentations) and nonbinary (ct/mri) images.

cchriste:
I fixed an issue with `Image::resample` that resulted in an output image from the resample with a row of pixels considered to be outside the image (the bottom row). Instead of setting the default pixel value of the resample filter to make believe those pixels are inside the image, I set the output image size smaller to ignore that row of pixels (just like it already does for the top layer of the 3d volume). 

With that change I've requested the review of a few of you who know the resample tool better to be able to confirm such a fix is appropriate. Thanks

cchriste:
As I mentioned on Slack, I reverted back to the original behavior of this function (sans binarization at the end). The issues I observed may require investigation, but they already existed in the standalone executable. I'm not even sure if they require an issue, so I'll just let the users of the function consider it in practice (one reason I'm going to merge this branch).

PR #305, opened 2020-01-15
Fix windows

akenmorris:
This fixes the shapeworks part of the superbuild on Windows.

Also fixes conda_installs.bat (hopefully).  This PR will create some binary build artifacts that we can try on various windows machines.


0leks:
At the end of the superbuild, it outputs the following:

akenmorris:
I don't think we need that for any platform anymore

PR #308, opened 2020-01-19
Documentation

AtefehKashani:
* Using itk to read the images for finding the reference image.
* added Optimize and use case documentations.


akenmorris:
I get:

ModuleNotFoundError: No module named 'itk'

Do we need to add this to conda_installs.sh until we are building python bindings ourself?

AtefehKashani:
Yes, it is the temporary solution for using itk. We can remove the SimpleITK from conda and add itk with pip. I will commit this changes for your review.

PR #311, opened 2020-01-23
Preview mesh qc changes

akenmorris:
This is not final answer, but a start.  Right now, you can use the coverage command with:

shapeworks readmesh --name=pelvis.vtk coverage --second_mesh=femur.vtk writemesh --name=shapeworks_out.vtk

Since we don't have any multi-data commands implemented, I have a temporary side solution in just adding a secondary mesh as a parameter to the coverage command.

This way we can get a build out to users to try in the short term.

This PR also adds FEMesh import and export directly from vtkPolyData.  No longer needing to read/write from disk.  Reconstruction has been modified to use this.


akenmorris:
I sent a Windows build of this to Joe and he was able to use the coverage command with success.

PR #319, opened 2020-01-25
New mesh from dt

jadie1:
MeshFromDistanceTransforms no longer works. NewMeshFromDT does, but it does not do smoothing or mesh decimation like MeshFromDistanceTransforms was supposed to. If this is added it could replace MeshFromDistanceTransforms.


akenmorris:
I think a file is missing:

CMake Error at Libs/Utils/CMakeLists.txt:46 (ADD_EXECUTABLE):
  Cannot find source file:

    NewMeshFromDT.cxx

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .m .M .mm .h .hh .h++ .hm
  .hpp .hxx .in .txx

PR #321, opened 2020-01-28
Executable padvolume

archanasri:


akenmorris:
The automated python test (e.g. ellipsoid) is failing with:

3:   File "ellipsoidMain.py", line 45, in <module>
3:     Run_Ellipsoid_Pipeline(args)
3:   File "/home/runner/work/ShapeWorks/ShapeWorks/Examples/Python/RunEllipsoid.py", line 105, in Run_Ellipsoid_Pipeline
3:     resampledFiles = applyIsotropicResampling(parentDir + "resampled", fileList)
3:   File "/home/runner/work/ShapeWorks/ShapeWorks/Examples/Python/GroomUtils.py", line 46, in applyIsotropicResampling
3:     if binary:
3: NameError: name 'binary' is not defined

There seems to be a mismatch between "binary" and "isBinary"

Applications/shapeworks/Commands.cpp
```diff
@@ -263,6 +287,34 @@ int SmoothMesh::execute(const optparse::Values &options, SharedCommandData &shar
   return sharedData.mesh.smooth(/*maxRMSErr, numIter*/);
 }
 
+///////////////////////////////////////////////////////////////////////////////
+// Coverage
+///////////////////////////////////////////////////////////////////////////////
+void Coverage::buildParser()
+{
+  const std::string prog = "coverage";
+  const std::string desc = "coverage between two meshes";
```
cchriste: This is not a sufficient description. What does "coverage between two meshes" mean? Maybe _compute_ coverage between two meshes?

cchriste: What is the result? Another mesh?

archanasri: I did not create the coverage command

Applications/shapeworks/Commands.cpp
```diff
@@ -263,6 +287,34 @@ int SmoothMesh::execute(const optparse::Values &options, SharedCommandData &shar
...
+  parser.prog(prog).description(desc);
+  parser.add_option("--second_mesh").action("store").type("string").set_default("").help("Second mesh to apply coverage.");
```
cchriste: What is the "second_mesh"? Is it a target to which "coverage" is applied? Is it therefore the result of the command? There just isn't enough information here for the command to be understood. I'm creating an issue about this (add detailed documentation shown only for <subcommand> --help), but for now maybe better to add a bit more description.

Applications/shapeworks/notes.txt
```diff
@@ -63,7 +65,12 @@ Image commands:
 [] *ExtractGivenLabelImage
 [] *FastMarching
 [] *FindLargestBoundingBox
-[] *PadVolumeWithConstant
+[] *PadVolumeWithConstant -> shapeworks padvolume
```
cchriste: If it's done let's put an 'x' in here. 👍

Applications/shapeworks/Commands.cpp
akenmorris: I will ask Praful & Steve how this should be described.

Examples/Scripts/Tools/PadVolumes.sh
```diff
@@ -124,9 +124,9 @@ do
     EchoWithColor  "imgfilename_padded $imgfilename_padded" "yellow"
     EchoWithColor "-------------------------------------------------------------------------------------------------" "yellow"
     
-    PadVolumeWithConstant --inFilename $segfilename --outFilename $segfilename_padded --paddingSize $padding_size --paddingValue $padVal
+    shapeworks readimage --name $segfilename padimage writeimage --name $segfilename_padded
```
cchriste: $padVal is no longer used. Please add it back to the commands here.

Libs/Image/Image.cpp
```diff
@@ -286,10 +286,29 @@ bool Image::resample(float isoSpacing, PixelType defaultvalue, Dims outputSize)
 ///////////////////////////////////////////////////////////////////////////////
 bool Image::compare_equal(const Image &other)
 {
+  // we use the region of interest filter here with the full region because our
```
cchriste: Is there another comparison function? How do we now test functions like `recenter` which do not modify the image, only its origin?

akenmorris: I believe it will still work for a function such as `recenter`.

Libs/Mesh/Mesh.cpp
```diff
@@ -1,66 +1,80 @@
 #include "Mesh.h"
+#include <PreviewMeshQC/FEAreaCoverage.h>
+#include <PreviewMeshQC/FEVTKImport.h>
+#include <PreviewMeshQC/FEVTKExport.h>
+
+#include <vtkPolyDataReader.h>
+#include <vtkPolyDataWriter.h>
 
 //#include <vtkVersion.h>
```
cchriste: This is great that the Mesh class implementation is starting. Please clean it up: no commented code like was in there, and either implement or completely drop the functions that were simply placeholders before. Thank you!

Libs/Mesh/Mesh.h
```diff
@@ -21,11 +20,14 @@ class Mesh
   /// \param filename
   bool write(const std::string &outFilename);
 
+  /// coverage
```
cchriste: Please emulate Image.h/cpp and move the function comments/descriptions to the .cpp file.

Testing/ImageTests/ImageTests.cpp
```diff
@@ -19,6 +19,43 @@ TEST(ImageTests, resample_test) {
...
+  Image ground_truth(test_location + "1x2x2.nrrd");
+
+  ASSERT_TRUE(image.compare_equal(ground_truth));
+}
+
+TEST(ImageTests, binarize_test) {
+
+std::string test_location = std::string(TEST_DATA_DIR) + std::string("/binarize/");
+
+Image image(test_location + "1x2x2.nrrd");
```
cchriste: As we were learning the hard way before, let's limit these tests to one per operation. Or we can figure out some way to indicate to the developer whose test fails which part of the test that failed (maybe googletests has some way to indicate "subtests" within a test).

Applications/shapeworks/notes.txt
archanasri: Done

PR #323, opened 2020-01-28
New mesh from dt

jadie1:
MeshFromDistanceTransforms no longer works. NewMeshFromDT does, but it does not do smoothing or mesh decimation like MeshFromDistanceTransforms was supposed to. If this is added it could replace MeshFromDistanceTransforms.

I named the file incorrectly in the last pull request.


jadie1:
Riddhish and I figured out how to fix the original MeshFromDistanceTransforms.cxx, so I updated that and removed NewMeshFromDT.cxx.

sheryjoe:
> 1. Can we add a google test for this?
Agreed ... @jadie1 @riddhishb 

> 2. Why is this in "Utils"?  Although I don't know if this belongs in Mesh or Image, anyone?
Following our convention thus far, this should be in Image as the tool is processing a distance transform (which is an image) - regardless that it produces a mesh. 

@riddhishb @jadie1, please make sure to move it to Image, remove redundant tools (if any, VTKFromDT, MeshFromXX) and work on adding the "command" (e.g. shapeworks dt2mesh) and the associated unit test (@archanasri and/or @cchriste can help you get this started).

jadie1:
This branch is caught up with Master and MeshFromDistanceTransforms has been moved to the Image dir. It is ready to merge.

PR #324, opened 2020-01-28
Hong documentation

HeavenlyBerserker:
Build Documentation. Changed structure and added Linux stuff.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,9 +1,7 @@
 ---
----
 
 # Building ShapeWorks from source
 
----
 ---
 
 This document contains detailed instructions on building ShapeWorks from scratch. This guide will consist of three main steps. The steps will be summarized, but you can click on the links to access further information.
--- a/BUILD.md
+++ b/BUILD.md
@@ -33,6 +33,9 @@ Follow these instructions to install the dependencies required to build shapewor
     * You can also install these [dependencies](deps.txt) manually if you prefer not to use Anaconda.
 
 5. **Qt5**: Download and install the latest version of [Qt5](https://download.qt.io/archive/qt/), selecting the LGPL (free) license. (at least version 5.10 required)
+
+   * [[OSX]](https://download.qt.io/archive/qt/5.13/5.13.0/qt-opensource-mac-x64-5.13.0.dmg) [[Linux]](https://download.qt.io/archive/qt/5.13/5.13.0/qt-opensource-linux-x64-5.13.0.run) [[Windows]](https://download.qt.io/archive/qt/5.13/5.13.0/qt-opensource-windows-x86-5.13.0.exe) 
+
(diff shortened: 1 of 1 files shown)


cchriste:
It looks like a lot of the problems I pointed out above were added before now. Sorry I hadn't had a chance to look things over earlier. But these are serious issues, in particular trying to get the user to manually install a whole bunch of stuff they don't need to because they're installed by conda, _including conda itself!_

akenmorris:
There is a little bit of a chicken and egg issue here with something like git-lfs.  The user can't run conda_install.sh until after they've cloned the repo, so they may or may not have git-lfs installed when they clone it.

cchriste:
For SuSE Leap 15.1 (the current default OS installed by SCI Support):
```
sudo zypper addrepo https://download.opensuse.org/repositories/devel:tools:scm/openSUSE_Leap_15.1/devel:tools:scm.repo
sudo zypper refresh
sudo apt-get install git-lfs
```

akenmorris:
What error do you get if it’s not installed? I’ve never gotten an error before, it just checks out the stubs instead of the actual files. I guess we could have conda installs install it and then do the two commands to enable it for the currently checked out repo.

cchriste:
I found the exact error in some other post investigating today, but
basically it just says it's not installed and therefore can't check out
master (I was on SuSE when I tested this), and `git status` showed tons of
files as deleted, but they couldn't be undeleted in the usual fashion of
`git checkout -- <file>`. Installing git-lfs is kind of a pain, so having
the conda version would be handy. We could provide conda_installs.sh
independently, so if git-lfs is needed they could simply run that first,
then clone. Or do the two-step like you mentioned.

akenmorris:
Is this what you saw?

git clone git@github.com:SCIInstitute/ShapeWorks.git shapeworks
Cloning into 'shapeworks'...
remote: Enumerating objects: 172, done.
remote: Counting objects: 100% (172/172), done.
remote: Compressing objects: 100% (123/123), done.
remote: Total 21025 (delta 82), reused 103 (delta 48), pack-reused 20853
Receiving objects: 100% (21025/21025), 100.49 MiB | 19.45 MiB/s, done.
Resolving deltas: 100% (13058/13058), done.
git-lfs filter-process: 1: git-lfs filter-process: git-lfs: not found
fatal: The remote end hung up unexpectedly
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry the checkout with 'git checkout -f HEAD'

akenmorris:
That's annoying.  Well, if it's just too onerous to have the user install both git and git-lfs, then we will need to find another solution for the test data.

cchriste:
Yes, that was the error. And good idea! There isn’t too much test data yet,
and I’ve been reducing the resolution of the data for new tests, so maybe
just keep it small for now.

akenmorris:
Ok, should we just move the images into the repo for now?  I think it will add about 100MB.

0leks:
conda_installs.bat currently requires anaconda to already be installed on windows

akenmorris:
Yes, it does. That should be in the README for Windows.

PR #329, opened 2020-01-31
Optimize library

akenmorris:
Do not merge this PR yet, but I am putting it up for feedback and to get people looking at it.

This PR makes the following changes:

1. Rename ShapeWorksRunApp to Optimize, remove template.  
2. Move all parameter file reading to OptimizeParameterFile.  OptimizeParameterFile takes a filename and an Optimize object and sets parameters.
3. Retarget Studio to use Optimize rather than "ShapeWorksOptimize" that was trimmed version of ShapeWorksRunApp.
4. Split out public API of Optimize vs internal methods.  All public methods are now truly meant to be public and have at least the start of Doxygen comments for each one.

The Optimize class is close to being able to be wrapped by Python and used directly.  One no longer needs to create a parameter file to use it.


sheryjoe:
Why do we have the ParticleSystem in Libs/Optimize rather than in Libs?

akenmorris:
ParticleSystem really is the Optimize Library. Perhaps we should dispense with the ParticleSystem directory and move this files up a directory to Optimize.

sheryjoe:
> ParticleSystem really is the Optimize Library. Perhaps we should dispense with the ParticleSystem directory and move this files up a directory to Optimize.


Would the particle system have functionality (APIs) that could be used in Analyze? for example, shape statistics? 

P.S. please check out the updated comment (with more questions).

akenmorris:
Yes, it's true that ParticleSystem could be used by Analyze.  Perhaps we should just rename Optimize to ParticleSystem and not have an Optimize library.  The optimization piece is almost entirely within the ParticleSystem code.

sheryjoe:
Agreed. Optimize could rather be a command <shapeworks optimize param.xml>

akenmorris:
What should we rename ShapeWorksRun to?

sheryjoe:
> What should we rename ShapeWorksRun to?


ShapeWorksOptimize or simply Optimize -- along these lines.

akenmorris:
I agree that a more generic parameter file class would probably be useful.  That's not the task I was trying to address here though.  I think we need more thought and design on universal parameter file support.

akenmorris:
I would vote for ShapeWorksOptimize for now.

sheryjoe:
> I agree that a more generic parameter file class would probably be useful. That's not the task I was trying to address here though. I think we need more thought and design on universal parameter file support.


Definitely -- let's finalize this task first with the OptimizeParameterFile as a prelude to a more generic input parameter file.

sheryjoe:
> I would vote for ShapeWorksOptimize for now.


Yep ... I second this choice.

PR #334, opened 2020-02-02
Adding Doxygen

akenmorris:
This PR adds a BUILD_DOCUMENTATION cmake flag to run Doxygen for C++ documentation.


akenmorris:
Also, we have some name clashes.  For example, the recently added Mesh class clashes with the Mesh class in Studio.

sheryjoe:
Until we refactor Studio, can we use/rely on the shapeworks namespace for the new Mesh class to resolve this?

akenmorris:
You are correct, they are already differentiated.

PR #335, opened 2020-02-03
Add group differences

akenmorris:
This PR adds group support similar to View2 in Studio.

At the moment, this must be in the parameter file in a View2 like parameter file.  We don't really have any way of specifying groups to a list of input images.  As we move to allowing more than two groups, we should consider how one should specify groups upon import into Studio.

To fix the longstanding issue in View2 where the group slider always affects the PCA slider, I've added radio buttons for PCA and Group sliders, so only one is active at a time.  The two no longer interact when groups are available.


akenmorris:
I think we should create a release from this point for the upcoming meeting.

PR #342, opened 2020-02-05
Add scalar bar to group difference display.

akenmorris:
Screen Shot 2020-02-04 at 11 43 22 PM


akenmorris:
I've added a separate issue about units for this scalar bar:

https://github.com/SCIInstitute/ShapeWorks/issues/347

PR #344, opened 2020-02-06
Femur use case

jadie1:
Full femur use case and use case reorganization and simplification.

Running: 
python RunUseCase.py --use_case femur
Will have the user select the cutting plane in the beginning of grooming on a sample of their choice

Running:
python RunUseCase.py --use_case femur --interactive 
Will have the user select the cutting plane in the middle of grooming on the reference


jadie1:
I think this is failing the tests because they are built around the old use case structure, it is giving the error:

Running command: python ellipsoidMain.py --tiny_test /home/runner/work/ShapeWorks/ShapeWorks/build/bin
4: python: can't open file 'ellipsoidMain.py': [Errno 2] No such file or directory

But now this test should be run as:
python RunUseCase.py --use_case ellipse --tiny_test

akenmorris:
I tried the femur use case and it died on Step 8 for me:

Step 8. Analysis - Reconstruct dense surface for samples along dominant PCA modes.

No number_of_correspondences_in_warp provided, all good correspondences will be used to build the warp ...!

...
...
...

#iterations: 3096
estimated error: 608.888
608.888
solving sparse system failed ...!
General exception caught.
	Returncode: -11
	Output: None
21:55:23.92 (512725.79s user, 16834.81s system, 670% cpu), running "python RunUseCase.py --use_case femur "

This was after almost 22 hours of processing (on kourosh)

jadie1:
I fixed the analyze set so it just opens studio. I also made a tiny test so when you run it with the option --tiny_test, it runs on only three of the samples (similar to the ellipse use case). So if you use that to test with that it'll run much faster.

akenmorris:
Great, I'll give it a try.

akenmorris:
I ran it and got:

  File "RunUseCase.py", line 49, in <module>
    module.Run_Pipeline(args)
  File "/Users/amorris/sci/shapeworks/code/Examples/Python/femur.py", line 190, in Run_Pipeline
    COM_filehandler = open(COM_filename, "r")
UnboundLocalError: local variable 'COM_filename' referenced before assignment

PR #345, opened 2020-02-06
Updated use case documentation with new way of running use case.

jadie1:
This should be merged after the femur use case branch is merged so that the documentation is consistent with the changes.


0leks:
I really like the gifs!

Issue #348, opened 2020-02-06
Turn on file compression for NRRD writing at every stage of our pipelines


akenmorris:
The output from our demo processing pipelines is unnecessarily large.  I compressed a 17GB left atrium dataset down to 6GB with a simple script.  11GB is a lot of wasted space.

FYI, this is not even a time/space tradeoff.  I have found that it's faster to save with compression on than it is to save the (very large) uncompressed image volumes.

cchriste:
The —compressed=true flag is set by default for writeimage. Otherwise, just
modify the standalone executables if they haven’t been consolidated into
the libraries yet.

akenmorris:
Done.

PR #354, opened 2020-02-14
Enable ITK writer compression on itk::ImageFileWriters

akenmorris:
Enabled compression on ITK image writers wherever I could find them in the code.  This should save time and space.


akenmorris:
Fixes  #348

PR #360, opened 2020-02-15
La changes

AtefehKashani:
I have added a function to CommonUtils , which takes the input files and runs k-mean clustering, then picks one sample from each cluster as the sub-sample data, by this approach we will get roughly the same set of samples for different runs.


akenmorris:
47 hours, but it finished!

PR #378, opened 2020-02-19
Attempting to fix eigen build on windows.

akenmorris:
We're having a problem that the eigen build is wrong for windows:

2020-02-19T18:58:26.9726289Z -- Configuring done
2020-02-19T18:58:31.6609916Z -- Generating done
2020-02-19T18:58:31.7035254Z -- Build files have been written to: D:/a/bdeps/eigen/build
2020-02-19T18:58:31.9449568Z mingw32-make: *** No rule to make target 'install'.  Stop.

However, for some reason this isn't aborting the github action and furthermore, the compile of shapeworks actually works.  But it does not cache correctly, so subsequent builds using the cache will fail.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,7 +44,6 @@ if (BUILD_TESTS)
   endif()
 endif(BUILD_TESTS)
 
-
 project(ShapeWorks)
 
 # use ccache if available
--- a/superbuild.sh
+++ b/superbuild.sh
@@ -209,8 +209,14 @@ build_eigen()
   if [[ $BUILD_CLEAN = 1 ]]; then rm -rf build; fi
   mkdir -p build && cd build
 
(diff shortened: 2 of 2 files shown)


medakk:
Re-running Linux, it failed on Install qt:

```
E: Failed to fetch https://packages.microsoft.com/ubuntu/16.04/prod/dists/xenial/main/binary-amd64/Packages  Writing more data than expected (670139 > 669553)
E: Some index files failed to download. They have been ignored, or old ones used instead.
##[error]The process 'sudo' failed with exit code 100
```

akenmorris:
Sometimes the Qt download fails. I usually just re-run.  It looks like it’s still going though.

akenmorris:
Oh, I also accidentally cancelled a Linux build and the restarted it.

PR #388, opened 2020-02-24
Fix conda_installs.sh for zsh (default on Mac Catalina)

akenmorris:
Fix conditionals for zsh.  Still works with bash.


cchriste:
I know I’m late to review, but these changes look fine, and they work on
both bash and zsh.

PR #390, opened 2020-02-24
Removed unused QualifierEntropyGradientFunction

0leks:
This PR is waiting on #329 to be merged first.

As per discussion with Shireen, Alan, Praful

Tested on ellipse use case

adaptivityMode is always 0


akenmorris:
Let’s wait until after the release to merge. This will also cause conflicts with the Optimize Library PR that’s been waiting for the release as well.

0leks:
makes sense

PR #396, opened 2020-02-26
Fix typos and layout of 'shapeworks' usage

akenmorris:
Fixed typo, moved second line of long description to match indent, added newline between command groups.

Before:

shapeworks-before

After:

shapeworks-after


cchriste:
Looks good to me.

PR #398, opened 2020-02-26
tested successfully with a use case 

prafulag:
ALso updated CMakeLists.txt to move the Find QT to occur only if GUI is on


prafulag:
@akenmorris : Any idea why QT install failing on windows but passed in Linux ?

akenmorris:
Yes, the install-qt action is broken. I’m trying to deal with it.

prafulag:
okay, thanks

akenmorris:
I've fixed it and merged to master.  If you merge from master and push again it should work now.

prafulag:
@akenmorris: thanks!

PR #399, opened 2020-02-27
Try qt action 2.4.1

akenmorris:


akenmorris:
This fixes github action qt-install for Windows.  I had to fork the action code and fix the problem myself.

PR #405, opened 2020-02-27
Use forward slash instead of os.sep to split string path up in groom utils

0leks:
Was getting the following error:
```
Continuing to groom.
  File "RunUseCase.py", line 54, in <module>
    module.Run_Pipeline(args)
  File "D:\Workspace\ShapeWorks\ShapeWorks\Examples\Python\femur.py", line 143, in Run_Pipeline
    reflectedFiles_mesh, reflectedFile_img = anatomyPairsToSingles(parentDir + 'reflected', files_mesh, files_img, reference_side)
  File "D:\Workspace\ShapeWorks\ShapeWorks\Examples\Python\GroomUtils.py", line 536, in anatomyPairsToSingles
    img_name = img.rsplit(os.sep, 1)[1]
IndexError: list index out of range
```
The file paths look like this: `TestFemur/femur/images/m03_1x_hip.nrrd`
But we are trying to split using \ as a delimiter on Windows.
Should just always use / since windows allows using either for file paths.
I fixed it by changing line 536 from `img_name = img.rsplit(os.sep, 1)[1]` to `img_name = img.rsplit('/', 1)[1]`
Another option is to do `img = img.replace(os.sep, '/')` before hand in case a \ sneaks in

### DIFF
--- a/Examples/Python/AnalyzeUtils.py
+++ b/Examples/Python/AnalyzeUtils.py
@@ -10,7 +10,7 @@ from termcolor import colored, cprint
 from CommonUtils import *
 
 
-def create_View2_xml(xmlfilename, dtFiles, localPointFiles, worldPointFiles):
+def create_analyze_xml(xmlfilename, dtFiles, localPointFiles, worldPointFiles):
     root = ET.Element('sample')
     input_points = ET.SubElement(root, 'point_files')
     input_points.text = "\n"
@@ -47,8 +47,8 @@ def create_View2_xml(xmlfilename, dtFiles, localPointFiles, worldPointFiles):
 
 
 def launchShapeWorksStudio(parentDir, dtFiles, localPointFiles, worldPointFiles):
-    xmlfilename = parentDir + '/shapeworksview2.xml'
-    create_View2_xml(xmlfilename, dtFiles, localPointFiles, worldPointFiles)
+    xmlfilename = parentDir + '/analyze.xml'
+    create_analyze_xml(xmlfilename, dtFiles, localPointFiles, worldPointFiles)
     create_cpp_xml(xmlfilename, xmlfilename)
     execCommand = ["ShapeWorksStudio" , xmlfilename ]
     subprocess.check_call(execCommand )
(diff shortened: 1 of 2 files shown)


jadie1:
Thanks for looking into this Oleks, I ran the femur tiny test with your fix on linux and the paths looked correct, but  I think we use os.sep for splitting in pretty much every function in GroomUtils, I am assuming they all need to be changed? 

Here is the os.sep documentation:
os.sep
    The character used by the operating system to separate pathname components. This is '/' for POSIX and '\\' for Windows. Note that knowing this is not sufficient to be able to parse or concatenate pathnames — use os.path.split() and os.path.join() — but it is occasionally useful. Also available via os.path.

Sounds like maybe we need to switch to using os.path.split() which would involve some thorough testing to make sure all the path names are right. Maybe for now we can just replace all the uses of os.sep with img = img.replace(os.sep, '/') and img_name = img.rsplit('/', 1)[1] as you did.

prafulag:
@jadie1: that seems like the ideal solution!

akenmorris:
Correct, we should be using os.path.split() (and dirname, basename).  See for example:

https://github.com/SCIInstitute/ShapeWorks/commit/ff56780d75689b12bba651d0a8e988509541e202

akenmorris:
Interestingly, os.path.split on windows can handle both "/" and "\\", where on mac/linux, it will only split on "/"

akenmorris:
@0leks , @jadie1 

I've replaced as much of the path handling as I could in GroomUtils.py.  See the diff.  We need to test all the use cases on both mac/linux and windows now.

jadie1:
Thanks Alan, I can test all the use cases on Linux

akenmorris:
Thanks @jadie1.  The tougher one will be Windows.  Unless we can modify the kmeans, we will need a Windows machine with ~64GB of RAM.  I will just assume that if it works on Linux that it works on Mac.  These changes should be entirely compatible between the two.

0leks:
works on ellipsoid and ellipsoid_fd, femur and left_atrium hasnt finished yet

akenmorris:
@0leks on windows?

Issue #406, opened 2020-02-27
DICOM to nrrd conversion command


farshaduou:
Please add a command to ShapeWorks for converting from DICOM to nrrd files.

akenmorris:
This request should probably be to add DICOM reading support to the existing read-image command. That way the conversion is something like “shapeworks read-image —dicom-dir=<path> write-image —name=out.nrrd”

akenmorris:
This was completed in #415

PR #415, opened 2020-03-03
Add image series (e.g. dicom) support to read-image command

akenmorris:
So I tried to add this option as "--dicom-dir", but the option parser we're using doesn't like the dash in the middle, so I changed it to --dicom_dir.  I didn't really intend to inject underscore into this, so I'm open to whatever suggestions people have.  I don't like "--dicomdir".

To convert a dicom to nrrd:

shapeworks read-image --dicom_dir=/path/to/dicom write-image --name=out.nrrd


akenmorris:
Resolves #406

cchriste:
It looks like there are types of images composed of a subdirectory full of images of that type. Do we intend to create separate commands in shapeworks::Image to read every one of these types of images? 

Another option would be to add a '/' to the image filename and then look for the first file in that directory to determine what type of image set is being read. I saw this was attempted [here](https://github.com/SCIInstitute/ShapeWorks/pull/415/commits/304e2bc69c2d4be816d4521ad1c5afc65eb1b946), but the Image::read function wasn't modified to handle directories as filenames.

akenmorris:
Now detects if path given to --name is a directory and if so, uses ImageSeriesReader if so.

cchriste:
This is great 👍 Is there a new test added to verify this works (I’m on my phone and haven’t kept up with the other commits for this task).

akenmorris:
Yes, here is the test:

https://github.com/SCIInstitute/ShapeWorks/blob/839cc822a93febdfdb8b8e98a7b3c7459c7d1b30/Testing/ImageTests/ImageTests.cpp#L94

The "/" is not necessary, I had just happened to add it when I was trying to figure out why the action wasn't finding the files.

I will update it to use the constructor rather than read()

cchriste:
Merged with latest master and tested. As long as the GitHub actions test work, I approve this pull request.

PR #418, opened 2020-03-04
Remove itk stuff from Domain classes

0leks:
Removes a bunch of itk clutter from domain classes, making them much easier to read through.

Tested on Windows: ellipsoid, left_atrium, femur
Tested on nonWindows: ellipsoid


akenmorris:
FYI, ellipsoid runs via github actions on all three platforms.

PR #419, opened 2020-03-04
use conda for gtest rather than downloading it every time cmake is run

cchriste:
You don't have to re-run conda_installs (but it's in there), just run:
`conda install gtest`
It's very fast and dependency-free.

This will break the GitHub actions since I don't think they use conda_installs.sh.

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -47,6 +47,18 @@ jobs:
         version: "5.9.9"
         cached: ${{ steps.cache-qt.outputs.cache-hit }}
 
+    - name: Setup Conda
+      uses: goanpeca/setup-miniconda@v1
+      with:
+          miniconda-version: 'latest'
+          
+    - name: Conda info
+      run: conda info
+
+    - name: conda installs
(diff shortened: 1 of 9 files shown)


cchriste:
Okay, so I noticed it's including conda, but not till after building for the Windows GitHub action. @akenmorris I'm not sure why this is the case so I'll leave it to your expertise to fix.

akenmorris:
The action is fixed.  Someone needs to test the distributable to make sure it hasn't picked up conda libraries that are not included in the distribution.

akenmorris:
FYI, CMake does not download Gtest every time you run CMake, only on a new directory.

cchriste:
Yes it does, or at least it requires internet access to check the currently downloaded version every time cmake is run, which is equally impossible when such access is not available. 

The gtest conda package only includes gtest, so if that is included then the distributions will have everything they need.

akenmorris:
I’ve had no problem continuing to run CMake with no network access. I just tried turning the WiFi off and it works just fine. 

I am not opposing this change, I am just correcting that it does not require network access every time you run CMake.

PR #420, opened 2020-03-05
Fix "fixed domain" flag initialization.

### LINKED ISSUE #416: Windows Issue with ellipsoid_fd use case
So I started encountering crashing on the ellipsoid_fd use case.
Can someone confirm if this issue happens on linux?
I narrowed down the pull request that causes this crash to #329 
Here is the command line output:
```
Step 4. Optimize - Particle Based Optimization
Reading inputfile: TestEllipsoidsFD/Ellipsoids_ExistingDT/seg.ellipsoid_0.isores.pad.com.aligned.cropped.tpSmoothDT.nrrd...
... reading more nrrd files ...
Reading inputfile: TestEllipsoidsFD/PrepOutput/distance_transforms/seg.ellipsoid_21.isores.pad.com.aligned.cropped.tpSmoothDT.nrrd...
General exception caught.
        Returncode: 3221225477
        Output: None
```
When building and running from #405 there are no problems.

Side note, there is an unrelated issue with directory paths in ellipsoid_fd.py introduced in #404 that I will have a pull request for shortly.

akenmorris:
Fixes #416

This should fix the problem you were having with elllipse_fd

### DIFF
--- a/Libs/Optimize/Optimize.cpp
+++ b/Libs/Optimize/Optimize.cpp
@@ -465,6 +465,16 @@ void Optimize::InitializeSampler()
   m_sampler->Initialize();
 
   m_sampler->GetOptimizer()->SetTolerance(0.0);
+
+  // These flags have to be set after Initialize, since Initialize will set them all to zero
+  for (unsigned int i = 0; i < this->m_domain_flags.size(); i++) {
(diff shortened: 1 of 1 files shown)


0leks:
femur was able to get through the optimize step!

PR #433, opened 2020-03-11
Cleaned Optimizer

0leks:
Added comments and cleaned up ParticleGradientDescentPositionOptimizer while trying to keep functionality exactly the same. (This included deleting unused Jacobi and non adaptive opimization functions)

Removed unused things from Domain classes such as the constraints_enabled flag (was always true) and moved sphere constraints code to a separate function for storage (A bunch of code that was after a return statement so never accessed)

Tested on:
Windows: ellipsoid
NonWindows:


Libs/Optimize/ParticleSystem/itkParticleGradientDescentPositionOptimizer.txx
```diff
@@ -17,10 +11,8 @@
 
 #ifdef SW_USE_OPENMP
 #include <omp.h>
-const int global_iteration = 1;
-#else /* SW_USE_OPENMP */
-const int global_iteration = 1;
```
akenmorris: I am not sure the current state of the OpenMP stuff in ShapeWorks.  But the reason this "global_iteration" const exists is to allow the OpenMP version of the code to run the parallel, per-shape part of the iteration multiple times per global part (e.g. m_GradientFunction->BeforeIteration()).  You can see that part is outside the OpenMP parallel loop, so the program runs in serial for that call.

Back when I added this, I was experimenting with allowing multiple passes of the parallel section.  The result was that the code ran quite a bit faster, but the result is not the same as the serial version.  At some point for safety, I left the checked-in code with "global_iteration = 1" for both cases, and would just change that number to '20' or whatever in my own local branch.

At this point, I have no idea if that would even work.  I have not had a chance to re-evaluate the parallelization opportunities.

0leks: Makes sense. I just notices that it was getting set in both the if and the else so I moved it outside of the ifelse

Libs/Optimize/ParticleSystem/itkParticleGradientDescentPositionOptimizer.txx
```diff
@@ -17,10 +11,8 @@
...
 #endif /* SW_USE_OPENMP */
+const int global_iteration = 1;
```
0leks: Here is where I moved it so it doesnt appear twice

0leks:
> I think you may also need to double-check the code in SurfaceSampler where the object is initialized. And maybe remove the associated xml tag, that becomes obsolete with this update.
Which object are you referring to being initialized?

prafulag:
The object for Optimizer class.

0leks:
Tested on ellipsoid with OpenMP.
Finished removing obsolete optimizer type from Optimizer and the XML settings

akenmorris:
I ran this with OMP at full threads and at one (disabled) on kourosh.  I got:

51 minutes vs 34 minutes.  The results were essentially the same.

I need to run the master branch for comparison.

akenmorris:
Results seem the same against master w.r.t. OpenMP

PR #435, opened 2020-03-11
Refactor ImageDomain

medakk:
This PR adds a few extra APIs to ParticleImageDomain so that the Optimizer and other classes don't do `domain->GetImage() ...`. This is a subset of the changes I've made for OpenVDB, and would be required to get the Optimizer to work with arbitrary domains(such as meshes).

(Note that the UpdateSurfaceArea function does not work currently. It is needed by ModifiedContangentGradientFunction, which hasn't been tested recently.)

@0leks does this cover everything you need?

### DIFF
--- a/Libs/Optimize/ParticleSystem/itkParticleImageDomain.h
+++ b/Libs/Optimize/ParticleSystem/itkParticleImageDomain.h
@@ -14,7 +14,7 @@
 #include <itkZeroCrossingImageFilter.h>
 #include <vtkContourFilter.h>
 #include <vtkMassProperties.h>
-#include <debug-install/include/ITK-5.0/itkImageRegionConstIteratorWithIndex.h>
+#include <itkImageRegionConstIteratorWithIndex.h>
 
 namespace itk
 {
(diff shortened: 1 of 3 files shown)


Libs/Optimize/ParticleSystem/itkParticleImageDomain.h
```diff
@@ -11,6 +11,10 @@
 #include "itkImage.h"
 #include "itkParticleRegionDomain.h"
 #include "itkLinearInterpolateImageFunction.h"
+#include <itkZeroCrossingImageFilter.h>
+#include <vtkContourFilter.h>
+#include <vtkMassProperties.h>
+#include <debug-install/include/ITK-5.0/itkImageRegionConstIteratorWithIndex.h>
```
0leks: This include is weird. I replaced it with #include <itkImageRegionConstIteratorWithIndex.h>

medakk: Oops, thanks.

akenmorris:
Looks good.

PR #436, opened 2020-03-11
Encapsulate Image Domain Functionality

0leks:
Finish making Optimizer use only ParticleDomain instead of ImplicitSurfaceDomain.
Once I start adding MeshDomain the Optimizer code will need a bit more refactoring.

dependent on #433 and #435


akenmorris:
Let's merge #433 and #435 first.  It's hard to see the diffs until they are merged.

akenmorris:
I believe I'm merging in the correct order, but this now has conflicts.  Can you check on this, @0leks ?

akenmorris:
Looks pretty simple, but I'll let you do it.

PR #439, opened 2020-03-17
Test adding of XLNT dependency

akenmorris:


akenmorris:
Yeah, in the 'studio' branch, there is a new library called "Project" that uses this (it's already been moved out of Studio).   This PR was really just to get the XLNT build working.  I don't see any need to merge it and force people to rebuild dependencies if they aren't used yet.

PR #441, opened 2020-03-17
Executable fastmarch, extractlabel, closeholes, threshold

archanasri:


akenmorris:
Does pybind11 need to be added to conda_installs?

archanasri:
> Does pybind11 need to be added to conda_installs?

Yes if we are using pybind but I don't want that change to happen in this branch.

Issue #160, opened 2019-09-20
rename 207 ITK-prefixed files to no longer have that prefix since they are not part of ITK


cchriste:

cchriste:
The `rename_itk-vtk-prefixed_files` branch starts this task, which includes the following:
- drop itk prefix from these files
- change associated includes
- take their classes out of namespace itk (this one means that anything in the class that assumed itk now needs to explicitly use itk::thing for that thing)

I was going to just remove this branch because it's gotten old, but the work so far may be useful to merge. You can see if by first checking out the branch and then diffing from the point before its changes started:
`git checkout rename_itk-vtk-prefixed_files`
`git diff e08336af9ace77fe27ae1888c242fc9ad53c30a6`

cchriste:
Another handy thing is that you can replace all the header include ifdefs with a simple `#pragma once`
For example:
```
#ifndef HEADERNAME_H
#define HEADERNAME_H
... (the header)
#endif 
```
becomes just
```
#pragma once
... (the header)
```

cchriste:
🎉

PR #452, opened 2020-03-25
make the build quieter by silencing some warnings which otherwise mak…

cchriste:
…e it almost impossible to notice more important warnings or errors

One modified file here also demonstrates how to silence the warnings that say a function lacks `override` in its declaration.

The modified trimesh CMakeLists.txt also shows how to set a compiler flag for only a single target rather than modifying CMAKE_CXX_FLAGS for the entire project. But the warning came up every time the file was included and so this change also simply removes the offending line of code, the argument for which is that the default value of that parameter is already 1.

### DIFF
--- a/ExternalLibs/trimesh2/include/map.h
+++ b/ExternalLibs/trimesh2/include/map.h
@@ -48,8 +48,8 @@ public:
         typedef MapFilter<FType,InType,OutType> MF;
         typename MF::Pointer mf = MF::New();
         mf->SetFunctor(&functor); // handles output
-        if(numThreads>0)
-            mf->SetNumberOfThreads(numThreads);
+        // if(numThreads>0)
+        //     mf->SetNumberOfThreads(numThreads);
         mf->SetInput(in);
         if(outRegion.GetSize()[0] > 0)
         {
@@ -96,8 +96,8 @@ public:
     typedef ReduceFilter<FType,InType,OutType> RF;
     typename RF::Pointer rf = RF::New();
     rf->SetFunctor(&functor);
-    if(numThreads>0)
-      rf->SetNumberOfThreads(numThreads);
+    // if(numThreads>0)
+    //   rf->SetNumberOfThreads(numThreads);
     rf->SetInput(in);
 
     rf->Update(); // throws
--- a/ExternalLibs/trimesh2/include/map.h
+++ b/ExternalLibs/trimesh2/include/map.h
@@ -2,6 +2,9 @@
  * Copyright (c) 2013 University of Utah
  */
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+
 #ifndef __map_H
 #define __map_H
 
@@ -48,8 +51,8 @@ public:
         typedef MapFilter<FType,InType,OutType> MF;
         typename MF::Pointer mf = MF::New();
         mf->SetFunctor(&functor); // handles output
-        // if(numThreads>0)
-        //     mf->SetNumberOfThreads(numThreads);
+	if(numThreads>0)
+            mf->SetNumberOfThreads(numThreads);
         mf->SetInput(in);
         if(outRegion.GetSize()[0] > 0)
         {
@@ -96,8 +99,8 @@ public:
     typedef ReduceFilter<FType,InType,OutType> RF;
     typename RF::Pointer rf = RF::New();
     rf->SetFunctor(&functor);
-    // if(numThreads>0)
-    //   rf->SetNumberOfThreads(numThreads);
+    if(numThreads>0)
+      rf->SetNumberOfThreads(numThreads);
     rf->SetInput(in);
 
     rf->Update(); // throws
@@ -110,3 +113,5 @@ public:
 
 
 #endif
+
+#pragma clang diagnostic pop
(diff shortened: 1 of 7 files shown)


ExternalLibs/trimesh2/include/map.h
```diff
@@ -48,8 +48,8 @@ struct map
         typedef MapFilter<FType,InType,OutType> MF;
         typename MF::Pointer mf = MF::New();
         mf->SetFunctor(&functor); // handles output
-        if(numThreads>0)
-            mf->SetNumberOfThreads(numThreads);
+        // if(numThreads>0)
```
akenmorris: I saw some discussion about this on issue #89, but I'm not sure what this has to do with compile time warnings.  

This doesn't change the behavior at all?

akenmorris:
I don't see why, but it is failing on Windows now:

CMake Error at C:/Miniconda3/envs/shapeworks/Library/share/cmake-3.15/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES)

ExternalLibs/trimesh2/include/map.h
cchriste: It won't change behavior, though it could slow some computation down. But I'm not even sure if this code is used, and it accounts for half of the warnings when we compile. Those warnings hide other important warnings -- there are so many I can't even find build errors when using Xcode! 

I commented on #89 in association with this commit yesterday (with a link to how to upgrade this deprecated multithread functionality) because that issue is the reason we still have this ancient code in the first place. This piece of our version of trimesh may not even be part of trimesh (I couldn't find it anywhere in the latest version).

cchriste:
> I don't see why, but it is failing on Windows now:
> 
> CMake Error at C:/Miniconda3/envs/shapeworks/Library/share/cmake-3.15/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
> Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES)

No idea why this is. CMake will forever be a mystery.

akenmorris:
My only guess would be that MSVC doesn't support -Wno-inconsistent-missing-override, so compile checks are failing.

ExternalLibs/trimesh2/include/map.h
akenmorris: @prafulag, @sheryjoe can you comment on this section of code?  I'm all for reducing compile warnings, but I'm not sure we should trade execution time for it, if that is truly the case. (I have no idea)

cchriste:
Can we please renew this pull request? 
I fixed the Windows build (override warnings will still be displayed there).

As for OSX and Linux: don't worry, there are still _plenty_ of other warnings.
As incentive to fix them, we'll re-enable the override warnings once the others are addressed!

Note the majority of the override warnings come from the itk files in Optimize/ParticleSystem, which aren't itk files at all (see #160). Another huge class of warnings comes from Trimesh, which we can't update since there is another set of files in there which apparently don't have much in common with trimesh and just happened to get mixed into that directory (see #89).

akenmorris:
Looks like it builds on Windows again!

I'm fine with everything on here except for the apparent trade of computation time (threads) for compiler warning.  If threads aren't really being used here, then it's moot, but I haven't heard an answer on it.

cchriste:
@Praful Agrawal <prafulag@cs.utah.edu>  @sheryjoe, will you please respond
to Alan's question? From what I can see, numthreads is never passed to
these functions, thereby using the default of one (i.e., serial
computation). In addition to being severely outdated, this code doesn't
seem to actually be used, so I modified it to remove the distracting
warnings (there are actually valid warnings but most people don't see them
due to the ones generated here). If it's okay with the two of you
(or @riddhish if he has an opinion), we'd like to temporarily mute these
warnings in order to eliminate that noise which distracts for noticing
other more important warnings, but we need you to validate that this is
indeed the case and the code here isn't being used multithreaded anyway, or
that it'll be updated to use modern techniques soon
Thanks ya'll!

prafulag:
@cchriste @akenmorris I think this code is from another PhD student in the past (not me or Prateep), @sheryjoe should be able to comment. 
From what I understand, this code is used to set the number of threads for functor based processing. It is used in the last steps of FIDS computation. This is a crucial multithreaded part which significantly reduces the FIDS computations.

cchriste:
Thank you, Praful! Can you or someone else please provide an example of the
FIDS processing? We would like to test this to see what the performance
differences are and what it might take to update the multithreaded
functionality (it's not super trivial or I would have just done it, but
maybe it's worth an hour cuz these warnings are costly by hiding other
potentially important issues). Thanks

akenmorris:
Can I suggest this solution to the warning instead of disabling multithreading?  (FYI it seems the number of threads is set to 6 in GenerateFidsFiles.cxx:217):

https://stackoverflow.com/questions/14065844/how-to-suppress-the-deprecation-warning-dispatch-get-current-queue-is-depreca/14065939

Just add 

#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wdeprecated-declarations"

to the top of map.h and then:

#pragma clang diagnostic pop

to the bottom.  This silenced the warning for me.

cchriste:
Thanks, Alan. Do you mind committing that to the branch I created for this?

Best,
Cameron

PR #458, opened 2020-04-04
Set C++ standard to C++14

akenmorris:
Seeing if ShapeWorks will build with C++17 on all the GitHub action platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.11)
-set(CMAKE_CXX_STANDARD 17)             # available options are [98, 11, 14, 17. 20]
+set(CMAKE_CXX_STANDARD 14)             # available options are [98, 11, 14, 17. 20]
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/CMake")
 
 #
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.11)
-set(CMAKE_CXX_STANDARD 11)             # available options are [98, 11, 14, 17. 20]
+set(CMAKE_CXX_STANDARD 17)             # available options are [98, 11, 14, 17. 20]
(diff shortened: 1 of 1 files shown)


cchriste:
😃

akenmorris:
Well, that seemed to work, but I ran into a very annoying error when trying to actually use it:

error: 'path' is unavailable: introduced in macOS 10.15
  std::filesystem::path path;

Despite the compiler supporting it, you apparently have to run OSX 10.15 to use it :(

cchriste:
Any chance of resolving this? I'm trying to use some stuff that's only available in C++17.

akenmorris:
I can't see any way that doesn't require all Mac developers and users to upgrade to 10.15.  I'm not entirely sure about Windows or Linux.

akenmorris:
Note: as far as I know, only the filesystem module requires 10.15, there could be other C++17 features that work on older Mac OS.  Unfortunately, it seems that you don't find out until you actually try to include and use a particular feature.

cchriste:
I was able to get things working using C++14, so I'll let you decide what to do about upgrading to 17.

akenmorris:
I'd really like to use std::filesystem, but it doesn't even work on my own system, so it's a no-go.

We're currently set to 11, do you want to try 14?

cchriste:
Yes, let's try 14 please.

akenmorris:
Just set CMAKE_CXX_STANDARD to 14 and then use a couple of the features you want (as I found out it's not enough to just set the standard) and let's see if it builds on all the GitHub platforms.  I don't think it will be a problem.  GCC has supported it since 6.1.

cchriste:
I got what I was doing working find with C++11, but I went ahead and
changed it to 14 and pushed so you can try the features that weren't
working for you.

akenmorris:
std::filesystem doesn't exist until C++17.

https://en.cppreference.com/w/cpp/header/filesystem

There are some Qt things in C++14 that are nice.  I've used them on all three platforms before.

cchriste:
I am in favor of moving to at least C++14 then (I actually thought we were
already using it :).

PR #463, opened 2020-04-07
Image warp

riddhishb:
Landmark driven image warp command written in executable manner inside the Image API


akenmorris:
The windows failure is just at installing Qt.  If this every happens, just restart the build.  The fetches for Qt seem to be flaky from time to time.

Applications/shapeworks/Commands.cpp
```diff
@@ -296,4 +296,34 @@ int Coverage::execute(const optparse::Values &options, SharedCommandData &shared
   return sharedData.mesh.coverage(second_mesh);
 }
 
+///////////////////////////////////////////////////////////////////////////////
+// Warp Image
```
cchriste, line 300: Excellent new command!

Libs/Image/Image.cpp
```diff
@@ -391,4 +393,82 @@ bool Image::pad(int padding, PixelType value)
 #endif
   return true;
 }
+
+bool Image::warp(std::string &source_file, std::string &target_file, int pointFactor)
```
cchriste, line 397: This function can just live in ImageUtils, produce the Transform, and call Image::applyTransform. 
This function has been in place for some time, but not yet in master.

Libs/Image/Image.cpp
```diff
@@ -391,4 +393,82 @@ bool Image::pad(int padding, PixelType value)
...
+  tps->SetTargetLandmarks(targetLandMarks);
+  tps->ComputeWMatrix();
+  // Set the resampler params
+  ResamplerType::Pointer resampler = ResamplerType::New();
+  InterpolatorType::Pointer interpolator = InterpolatorType::New();
+  resampler->SetInterpolator( interpolator );
+  ImageType::SpacingType spacing = image->GetSpacing();
+  ImageType::PointType   origin  = image->GetOrigin();
+  ImageType::DirectionType direction  = image->GetDirection();
+  ImageType::RegionType region = image->GetBufferedRegion();
```
cchriste, line 446: Image::applyTransform by default applies it to the whole image. GetBufferedRegion is unnecessary and only affects threading, which is also used by default if possible for applying the filter. For the way this function is written, no threading would be possible anyway, but for our Image::applyTransform maybe there's a chance (I'm still skeptical).

Libs/Image/Image.h
```diff
@@ -4,6 +4,10 @@
 #include "ImageUtils.h"
 #include <limits>
 #include <itkTranslationTransform.h>
+#include "itkThinPlateSplineKernelTransform.h"
+// Software Guide : EndCodeSnippet
```
cchriste, line 8: tsk-tsk!

Libs/Image/Image.h
```diff
@@ -13,6 +17,13 @@ class Image
   static const unsigned dims = 3;
   using PixelType = float;
   using ImageType = itk::Image<PixelType, dims>;
+  typedef double CoordinateRepType;
```
cchriste, line 20: All of these should be only in the ImageUtils function itself, not in the outer namespace.

Libs/Image/Image.h
```diff
@@ -27,8 +38,8 @@ class Image
   bool isoresample(double isoSpacing = 1.0f, Dims outputSize = Dims());
   bool pad(int padding, PixelType value);
   // bool nextfunction(...);
-
   bool compare_equal(const Image &other);
+  bool warp(std::string &source_file, std::string &target_file, int pointFactor);
```
cchriste, line 42: Make this a static function of the ImageUtils class.

Libs/Image/Image.cpp
```diff
@@ -14,6 +14,8 @@
 #include <itkImageSeriesReader.h>
 #include <itkGDCMImageIO.h>
 #include <itkGDCMSeriesFileNames.h>
+#include "itkThinPlateSplineKernelTransform.h"
```
cchriste, line 17: actual ITK includes should use braces, not quotes (I think these aren't the versions in our tree)

Libs/Image/Image.h
cchriste: no need for the same includes to exist in both the header and the cpp

cchriste:
Hi @riddhishb, I'm making the changes above to demonstrate how things can be added.

Is this a function for "reverse warp"? Inside the function it says "reverse warp thats why reverse order of p1 p2".  For a general purpose function, would it be okay to simply reverse the order of source and target?

akenmorris:
@cchriste, the transform would be set up to be reverse so that when the reference (fixed) space is sampled (into the moving), it transforms the moving into reference.

akenmorris:
Do we have a test for this command?  If not, we should probably add one.

cchriste:
> @cchriste, the transform would be set up to be reverse so that when the reference (fixed) space is sampled (into the moving), it transforms the moving into reference.

Let's make a single function for which the order of parameters can be reversed.

cchriste:
> @cchriste, the transform would be set up to be reverse so that when the reference (fixed) space is sampled (into the moving), it transforms the moving into reference.

I reversed the arguments and clarified in the command help that it transforms from source to target. If that's not what we want, go ahead and change it, but just keep it very clear and maybe even mention that simply reversing the order or the arguments will go the other direction.

cchriste:
I made these changes based on executable and they will be merge to master from there. 
Closing this pull request.

PR #479, opened 2020-05-07
Separate centering resampling

jadie1:
I removed the recenter call from applyIsotropicResampling in GroomUtils.py and had the use cases call the "center" function in GroomUtils if they needed it. I also fixed the girder naming issue Alan mentioned for the femur use case.
I tested this on all of the use cases and this change did not effect the results.

### DIFF
--- a/Examples/Python/GroomUtils.py
+++ b/Examples/Python/GroomUtils.py
@@ -29,7 +29,7 @@ def rename(inname, outDir, extension_addition, extension_change=''):
     print("######################################\n")
     return outname
 
-def applyIsotropicResampling(outDir, inDataList, isoSpacing=1.0, recenter=True, isBinary=True):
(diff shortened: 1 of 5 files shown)


cchriste:
It looks fine, conforming with the current groom processing patterns.

That said, those patters create i/o bottlenecks. So while it's nice to separate the various steps, what each does is, one file at a time, reads the file, runs a command on that file, and writes the result. For most of these cases, a file could be read and all commands executed while it's in memory, and the result written at the end. Furthermore, the process could be carried out in parallel (especially if significant processing is required for any step and it the step has no function-level parallelism). There is also sometimes usefulness in looking at mid-process results. We should consider a way to accomplish both application-level parallelism as well as how one might consider mid-process results. 

I'll create a high level issue for all this. It's not urgent, but should be on our radar as. There are details even going down to the individual subcommands.

PR #482, opened 2020-05-13
Reduce memory usage in inverse covariance matrix

medakk:
This reduces memory usage in itkParticleGeneralMeshBasedGeneralEntropyGradientFunction when mesh_attributes is enabled. Avoids some copies, some allocations, and reordered/expanded matrix operations to run faster. In the pelvis case, memory requirement for the particle system has reduced from ~14GB to ~5GB. Iteration time seems to have reduced by about 50% in the same use case.

Same fixes implemented in itkParticleEnsembleEntropyFunction.

### DIFF
--- a/Libs/Optimize/CMakeLists.txt
+++ b/Libs/Optimize/CMakeLists.txt
@@ -12,6 +12,7 @@ target_link_libraries(Optimize
   trimesh2
   Mesh
   tinyxml
+  Utils
   Particles)
 
 target_include_directories(Optimize PUBLIC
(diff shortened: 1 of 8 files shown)


medakk:
@akenmorris Any idea why the Mac build can't find ITK?

medakk:
Mac build fails because ITK didn't build, but that step in the CI(Build dependencies) was green anyway.

I'll revert to the working version of VXL.

medakk:
@akenmorris Yes, the final shape models are correct

PR #484, opened 2020-05-14
Add Warp Image command using ImageUtils::computeWarp transform function.

cchriste:
This is the function previously added directly to the Image class, but that can now be used as an example of an ImageUtils function to create a transform then applied using Image::applyTransform.


cchriste:
Note the pull request is to go into executable, which has some broken unit tests. It should be merged nonetheless since it contains other improvements which were required to maintain elegance (see commit log for details).

PR #487, opened 2020-05-18
Update evaluation commands

medakk:


Libs/Particles/ParticleSystem.h
```diff
@@ -9,7 +9,7 @@ class ParticleSystem
 public:
   ParticleSystem();
 
-  bool LoadParticles(const std::vector<std::string> &paths);
+  void LoadParticles(const std::vector<std::string> &paths);
```
cchriste, line 12: Is ParticleSystem like Image in the sense that the only valid instantiation has loaded a file? If so, why not emulate Image and put the argument of file names in the constructor, making a default constructor private?

Libs/Particles/ParticleSystem.cpp
```diff
@@ -5,12 +5,10 @@ ParticleSystem::ParticleSystem()
...
 }
 
-bool ParticleSystem::LoadParticles(const std::vector<std::string> &_paths)
+void ParticleSystem::LoadParticles(const std::vector<std::string> &_paths)
 {
-
   if (_paths.empty()) {
-    std::cerr << "No filenames passed to readparticlesystem; returning false." << std::endl;
-    return false;
+    throw std::runtime_error("No filenames passed to readparticlesystem");
```
cchriste, line 11: This reinforces my previous comment since loading particles is now a non-optional operation.

Libs/Particles/ParticleSystem.h
medakk: Done. I've made a new PR, closing this
https://github.com/SCIInstitute/ShapeWorks/pull/522

Libs/Particles/ParticleSystem.cpp
medakk: Loading is non-optional. This check is just to ensure that we haven't been passed an _empty_ list of files

medakk:
Made a new one
https://github.com/SCIInstitute/ShapeWorks/pull/522

PR #488, opened 2020-05-19
Reduce fixed domain memory usage

akenmorris:
This PR reduces memory usages for fixed domains.  It eliminates the loading of distance transforms for fixed domains (though they must still be specified in the xml file), and eliminates the computation of gradients, curvature and hessians for fixed domains.

### DIFF
--- a/Libs/Optimize/Optimize.h
+++ b/Libs/Optimize/Optimize.h
@@ -222,6 +222,8 @@ public:
   //! Set Domain Flags (TODO: details)
   void SetDomainFlags(std::vector<int> flags);
 
+  std::vector<int> GetDomainFlags();
+
   //! Set if file output is enabled
   void SetFileOutputEnabled(bool enabled);
 
--- a/Libs/Optimize/Optimize.h
+++ b/Libs/Optimize/Optimize.h
@@ -222,7 +222,7 @@ public:
   //! Set Domain Flags (TODO: details)
   void SetDomainFlags(std::vector<int> flags);
 
-  std::vector<int> GetDomainFlags();
+  const std::vector<int>& GetDomainFlags();
 
   //! Set if file output is enabled
   void SetFileOutputEnabled(bool enabled);
--- a/Libs/Optimize/ParticleSystem/itkParticleImageDomainWithCurvature.h
+++ b/Libs/Optimize/ParticleSystem/itkParticleImageDomainWithCurvature.h
@@ -41,10 +41,14 @@ public:
 
   /** Set/Get the itk::Image specifying the particle domain.  The set method
       modifies the parent class LowerBound and UpperBound. */
-  void SetImage(ImageType *I)
+  void SetImage(ImageType *I, bool minimal)
   {
     // Computes partial derivatives in parent class
-    Superclass::SetImage(I);
+    Superclass::SetImage(I, minimal);
+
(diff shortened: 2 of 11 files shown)


Libs/Optimize/ParticleSystem/itkParticleImageDomainWithCurvature.h
```diff
@@ -80,21 +84,25 @@ class ParticleImageDomainWithCurvature : public ParticleImageDomainWithHessians<
...
     //this->DeletePartialDerivativeImages();
     
+    m_CurvatureInterpolator = ScalarInterpolatorType::New();
     m_CurvatureInterpolator->SetInputImage(m_CurvatureImage);
   } // end setimage
   
   double GetCurvature(const PointType &pos) const
   {
-    return m_CurvatureInterpolator->Evaluate(pos);
+    if (m_CurvatureInterpolator) {
```
medakk, line 93: Is this using `m_CurvatureInterpolator` as a proxy for whether or not we are in `minimal` mode?

Libs/Optimize/Optimize.h
```diff
@@ -222,6 +222,8 @@ class Optimize
   //! Set Domain Flags (TODO: details)
   void SetDomainFlags(std::vector<int> flags);
 
+  std::vector<int> GetDomainFlags();
```
medakk, line 225: Maybe return a const reference?

Libs/Optimize/ParticleSystem/itkParticleSystem.h
```diff
@@ -361,11 +361,25 @@ class ITK_EXPORT ParticleSystem : public DataObject
...
   { m_DomainFlags[i] = false; }
   bool GetDomainFlag(unsigned int i) const
-  { return m_DomainFlags[i]; }
+  {
+    if (i >= m_DomainFlags.size()) {
+      // not set
+      return false;
+    }
+    return m_DomainFlags[i];
+  }
```
medakk, line 382: There's a m_DomainFlags here and in Optimize.h?

Libs/Optimize/Optimize.h
akenmorris: Sure

Libs/Optimize/ParticleSystem/itkParticleSystem.h
akenmorris: There is.  It's sort of redundant.  The parameters need cleaning up for sure.  The domain vector of ints of domain flags in Optimize.h are a list of flagged domains (e.g. 2, 8, 87, 124) Whereas the particle system m_DomainFlags[i] indexes the actual domain, e.g. m_DomainFlags[2] = true, etc.

Libs/Optimize/ParticleSystem/itkParticleImageDomainWithCurvature.h
akenmorris: Sort of.  I'm skipping the generation of the gradient/hessian images for the fixed domains since the data will never be used.  Unfortunately, there is a still a callback during AddPosition that ends up asking for the curvature.  Without this check, it will crash.  The curvature doesn't matter for the fixed domains.

The initialization structure needs to be reorganized for sure, this is sort of a temporary kludge to skip some parts for now.

PR #493, opened 2020-05-23
Check for nvidia-smi, use to pick version of pytorch to install.

akenmorris:
Please try this out on any cuda-enabled machines you have access to.  

conda_installs.sh will check for nvidia-smi and attempt to use it to pick an appropriate pytorch package to install, falling back to cpuonly.

This might work on windows for developers using msys, but it occurs to me that we aren't requiring end users to have bash/msys/etc installed, they only run conda_installs.bat and I have no idea how to do similar sed/grep mumbo-jumbo in a windows batch script.  

Suggestions, anyone?

### DIFF
--- a/conda_installs.sh
+++ b/conda_installs.sh
@@ -9,6 +9,27 @@ if [[ "$sourced" == "0" ]]; then
   exit 1
 fi
 
+# PyTorch installation
+function install_pytorch() {
+  echo "installing pytorch"
+  PYTORCH="cpuonly"
+  if ! [ -x "$(command -v nvidia-smi)" ]; then
+    echo 'Could not find nvidia-smi, using cpu-only PyTorch'
+  else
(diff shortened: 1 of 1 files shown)


jadie1:
I tried this on my linux machine which has an old GPU (needs CUDA version 9.0) and got:
`Found CUDA Version: 
CUDA version not compatible, using cpu-only`

akenmorris:
The pytorch website only describes support for 9.2, 10.1 and 10.2. Are there conda packages for 9.0?

jadie1:
I think they only actively support 9.2, 10.1, and 10.2 but there are older ones you can still use and install the same way. I've tried on the SCI Kepler machines and those have even older GPU's and need version 8.0...

akenmorris:
Closing this for now.

PR #494, opened 2020-05-24
Fixed Domain optimizations

akenmorris:
Merge the fixed domain memory optimization into openvdb branch instead.

I got tired of fighting with the MaximumEntropySurfaceSurfaceSampler's inheritance of InplaceImageFilter and just changed it to be a DataObject.  This also fixes the issue of the first image being held by SetInput.

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -14,7 +14,7 @@ env:
 jobs:
   build:
 
-    runs-on: ubuntu-16.04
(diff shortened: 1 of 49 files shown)


akenmorris:
Should we add it as a parameter to AddImage and SetImage and remove it from sampler and domain?

medakk:
> Should we add it as a parameter to AddImage and SetImage and remove it from sampler and domain?

That's a good idea.

PR #495, opened 2020-05-26
Refactor datasetutils

0leks:
Throw exceptions instead of returning false
Clean up functions and function naming


akenmorris:
@0leks , should this be merged for the 5.4.0 release?

0leks:
Yes. I will also soon be making a PR for the updated datasets

PR #496, opened 2020-05-26
add link to data portal to top of datasets.md and top of usecases.md

0leks:
Should wait to merge this until the permissions on shapeworks-dev allow random people to send emails. If this isn't feasible we should make a shapeworks-support email list that does


sheryjoe:
I initiated shapeworks-dev-support email list (public list) to support dev community. It is pending activation.

PR #500, opened 2020-05-27
Fix la memory

AtefehKashani:
I have edited the "FindReferenceImage" function to take care of different image sizes and prevent loading all the images into memory.  I have tested on LA data.
For sampling a subset of data the "sampledata" is changed to just read two images at a time and run spectral clustering based on the distance matrix of input images. 
@jadie1  Could you please test this two functions for the femur data as well.

### DIFF
--- a/Examples/Python/CommonUtils.py
+++ b/Examples/Python/CommonUtils.py
@@ -38,34 +38,8 @@ def create_cpp_xml(filename, outputfilename):
     file = open(outputfilename,"w")
     file.write(xml_text)
     file.close()
-    
+
 def sampledata(inDataList, num_sample):
-    IMG = []
-    DIM = []
-    for i in range(len(inDataList)):
-        image = itk.imread(inDataList[i], itk.F)
-        tmp = itk.GetArrayFromImage(image)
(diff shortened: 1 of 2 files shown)


akenmorris:
Thanks, @AtefehKashani , I'm trying this out now.

This will be very important for our release as it will allow people to run the use case who couldn't before.

akenmorris:
@jadie1 , is there also a way we could add an option to run the femur case with a pre-defined cutting plane?  It's very difficult to get X11 forwarding and GLX working to run it on remote machines.

jadie1:
@akenmorris I could add this. It would take a little time and we'd have to think through how have the user specify which way to do it since there'd be three options now.

jadie1:
@AtefehKashani Does this need to be merged with the executable branch? I am trying to test it with the femur use case but I keep getting command not found for things like "isoresample" and 'binarize" because the names have been changed.

AtefehKashani:
@jadie1 Please check the path to the binaries that you built. I modified the bin path to my shape works bin directory to make it work. I am not sure that the executable branch is merged to the master yet.

PR #519, opened 2020-05-30
fix the VTK_DIR for the itk cmake command in build_dependencies.sh

0leks:


akenmorris:
So this changes VTK_DIR from:

-DVTK_DIR="d:\a\ShapeWorks\deps\lib\cmake\vtk-8.2"

to: 

-DVTK_DIR="d:/a/ShapeWorks/deps/lib/cmake/vtk-8.2"

Which makes CMake fail:

 when parsing string

    d:\a\ShapeWorks\deps/lib/cmake/vtk-8.2

  Invalid character escape '\a'.

akenmorris:
Finally works.

PR #537, opened 2020-06-03
Add pre-defined cutting plane to femur use case

jadie1:
Added a pre-defined cutting plane to femur use case that is used when the use case is run without the interactive tag.
When the interactive tag is used, the user decides between defining the plane on a sample of their choice or the reference.
Updated Femur.md accordingly. 
Also fixed a small typo in GroomUtils that was causing problems with path names.


akenmorris:
I get:

  File "RunUseCase.py", line 53, in <module>
    module = __import__(args.use_case)
  File "/home/csafe/amorris/sci/shapeworks/Examples/Python/femur.py", line 63
    extract the zipfile
              ^
SyntaxError: invalid syntax

---------------

Seems like this should be in a comment

Issue #530, opened 2020-06-02
Organize Documentation and update GettingStarted 


sheryjoe:
- should be the top level for SSM workflow
- should introduce the three phases with links to their md
- should be in the documentation folder
- fix the link in the readme
- link to use cases

sheryjoe:
Shireen to take the first crack on the getting started.

akenmorris:
Images is already capitalized:

https://github.com/SCIInstitute/ShapeWorks/tree/master/Images

I don't think there is an 'install' folder.  Maybe inside dependencies from build_dependencies.sh?

sheryjoe:
I was referring to Documentation/images and Documentation/install (which is now capitalized).

sheryjoe:
This has been addressed in the org-doc branch. To be merged soon.

PR #541, opened 2020-06-04
Organize documentation and add getting started

sheryjoe:
This is to address #530

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ Installation
 Users
 ---------------------
 
-ShapeWorks provides official user releases, the features of which can be seen at [ChangeLog](ChangeLog).
+ShapeWorks provides official user releases, the features of which can be seen at [ChangeLog](Documentation/ChangeLog.md).
 
 First, download the latest ShapeWorks binary release:  
 https://github.com/SCIInstitute/ShapeWorks/releases/latest


Examples/Python/RunUseCase.py
```diff
@@ -42,7 +42,7 @@
 parser.add_argument("--start_with_image_and_segmentation_data", help = "use images and segmentations data for preprocessing", action="store_true")
 parser.add_argument("--use_single_scale", help="Single scale or multi scale optimization", action="store_true")
 parser.add_argument("--tiny_test", help="Run as a short test", action="store_true")
-parser.add_argument("shapeworks_path", help="Path to ShapeWorks executables (default: "+binpath+")", nargs='?', type=str, default=binpath)
+parser.add_argument("--shapeworks_path", help="Path to ShapeWorks executables (default: "+binpath+")", nargs='?', type=str, default=binpath)
```
akenmorris, line 45: This change breaks the PythonTests that assumes the last optional argument is the shapeworks path.  This argument did not previous need "--shapeworks_path"

jadie1:
I noticed a few things:

- The Q.png image isn't linked properly at /Documentation/Workflow/Optimize.md
- The xml referenced at /Documentation/Workflow/Analyze.md is now called analyze.xml instead of shapeworksview2.xml
- The links to groom, optimize, and analyze.md are now broken in the individual use case .md's like ellipse.md and femur.md

Other than that I tried to check all of the links and I think they look good. I can make these changes tomorrow if that's easiest. 
It would be good to merge [PR #537 ](https://github.com/SCIInstitute/ShapeWorks/pull/537) first if possible. That is the predefined cutting plane one and I updated femur.md for it

sheryjoe:
I have fixed those. In general, please don't use explicit GitHub links, unless needed. The groom, optimize, and analyze links were broken because they were linked to the master. These links should be relative to the documentation folder.

akenmorris:
When we merge this, can we do a 'squash and merge' so that the unused large GIF files don't enter into the history?

sheryjoe:
Yes. We can.

sheryjoe:
@jadie1 @AtefehKashani @akenmorris how about this double dash for shapeworks binpath. this should also be relfected in the RunUseCase documentation.

akenmorris:
I don't know about the double dash for the path.  @cchriste , @jadie1 , @AtefehKashani , was there a reason it was an un-named optional final argument instead?

sheryjoe:
In general, we should maintain consistency for input arguments. Double dashes are used for use case input arguments. It makes not sense to have one of those arguments without the double dashes. Let's either use the double dashes for the shapeworks bin path or otherwise remove the argument if it has no particular use/justification.

AtefehKashani:
I would say it is better to have the shapeworks binpath as an argument, since it may be installed in different path and setting  the default path will break the script.

cchriste:
I agree completely that it should be a double-dash argument. It was
originally added when there were far fewer args and no standalone
executable.

On a side note, single dash, single letter arguments are common shortcut
arguments (ex: —foobar/-f). They exist in many of our executables and while
I wouldn’t recommend using them for examples, I don’t believe their use
should be discouraged.

akenmorris:
Sounds good.  I just didn't know if there was some rationale for the current behavior.

sheryjoe:
Agreed! @jadie1 @AtefehKashani  you can do something like below. Not only for consistency across different tools/interfaces we provide, but also single letter args could be very convenient short cut for some users especially when running these scripts multiple times.

parser.add_argument('--verbose', '-v',
                    action='store_true',
                    help='verbosely list files processed')

AtefehKashani:
what is this verbose argument pointing to? the level of verbosity in the pipeline (like optimization)?

akenmorris:
What else do we need to merge this?

sheryjoe:
> what is this verbose argument pointing to? the level of verbosity in the pipeline (like optimization)?

That was meant to be an example.

akenmorris:
Are we ready to merge?

sheryjoe:
I think so

akenmorris:
What about the gifs?

sheryjoe:
For now, How about dividing this video into three gifs to demo groom, optimize and analyze. Do you think This will reduce their individual sizes? We could also rely on YouTube video for now till we agree on the way to move forward.

akenmorris:
Yes, that sounds good, but we will need to split the others as well as I see several GIFs that are almost 60MB in the repo.

akenmorris:
This is a real pain, it looks like the limit is more like 5MB:

http://progsharing.blogspot.com/2018/06/gifs-on-github-pages-content-length.html

sheryjoe:
Let's reduce the resolutions of these gifs and have a link to a high res youtube video. I think we should move to a static site generator hosted on SCI sometime soon to mitigate these content limitations.

AtefehKashani:
In the Build/BUILD.md the link to "Adding to PATH" is not working I assume it to be in the "How to" section of the getting started but it is missed.  This was a part of Gettingstarted.md .

sheryjoe:
> In the Build/BUILD.md the link to "Adding to PATH" is not working I assume it to be in the "How to" section of the getting started but it is missed. This was a part of Gettingstarted.md .

Fixed. Good catch!

akenmorris:
> Let's reduce the resolutions of these gifs and have a link to a high res youtube video.

I don't have access to the original videos except for studio.

sheryjoe:
@AtefehKashani @jadie1 could you upload the original videos for your gifs/mp4 to the below folder?

https://drive.google.com/drive/folders/101pBP085XNKIQe7HZw_gOvMvJTl47vxC?usp=sharing

sheryjoe:
@akenmorris for any video on youtube, please add a brief description on what the video is show casing.

akenmorris:
Sounds good.  Are you sure you want GIFs as well as these will not be obvious that you can click on them to load a video.  What they suggested in those threads was to put the video on youtube, then take a screenshot of the video with the red play button over it.  Then it is obvious to the user that it's a video they can click and watch.

sheryjoe:
This is even better. The red button will entice the user to click and play! and without keeping gifs on our repo. I like this. Let's do it.

akenmorris:
How's this as an example:

https://github.com/SCIInstitute/ShapeWorks/blob/org-doc/Documentation/GettingStarted.md

sheryjoe:
Looks good!

PR #545, opened 2020-06-05
Datasets v1

0leks:
Fixed folder naming for left_atrium, ellipsoid, ellipsoid_fd
uploaded all datasets to use-case-data-v1
Switch to downloading from use-case-data-v1


akenmorris:
I get this when I run it:

Downloading the left_atrium dataset from the ShapeWorks Portal
Collection: use-case-data-v0
WARNING finding collection: use-case-data-v0
Found 2 collections with name = use-case-data-v0.
Using collection with id = 5e4b7c000a02fb02ba2427b9

PR #560, opened 2020-06-10
Domain init cleanup

0leks:
Bunch of non-breaking cleanup that doesn't change functionality.

TODO run the image use cases and verify that they still work compared to master branch
- ~ellipsoid~
- ~ellipsoid_fd~
- ~left_atrium~
- ~femur~


akenmorris:
When dealing with changing the order of initialization, we will have to check examples like "fixed_domains" and stuff like that.

Libs/Optimize/Optimize.cpp
```diff
@@ -228,6 +228,24 @@ int Optimize::GetDomainsPerShape()
   return this->m_domains_per_shape;
 }
 
+//---------------------------------------------------------------------------
+void Optimize::SetDomainType(std::string& domain_type)
```
akenmorris, line 232: Should we use an enum or something rather than a std::string?

Libs/Optimize/Optimize.h
```diff
@@ -81,6 +82,9 @@ class Optimize
   //! Return the number of domains per shape
   int GetDomainsPerShape();
 
+  void SetDomainType(std::string &_domain_type);
```
akenmorris, line 85: Is that a leading underscore?

Libs/Optimize/Optimize.cpp
0leks: This is the string it loads from the params file. At every point after this it uses the DomainType enum.

akenmorris: Oh, you have a DomainType.  So it's a question of where the conversion from the parameter file's string takes place...

akenmorris: Should we move it to OptimizeParameterFile.cpp?  I guess the mismatch of get/set is bothering me.

Libs/Optimize/Optimize.h
0leks: Oops

Libs/Optimize/Optimize.cpp
0leks: Mismatch fixed

akenmorris:
All looks ok to me now.  Ok to merge pending the test results.  @0leks you are running LA?

0leks:
> All looks ok to me now. Ok to merge pending the test results. @0leks you are running LA?

Yeah Im running LA single and multiscale to compare to master branch

0leks:
Ill also run ellipsoid_fd after that which will be very quick

akenmorris:
Yes, please check FD. I need to add the automated test for it.

0leks:
ellipsoid_fd looks good

0leks:
This PR does not encounter the narrow band issue for femur multiscale. It is ready to go.

akenmorris:
@0leks, was a a different branch you were having the problem with femur?

0leks:
Yes the femur problem is on the mesh_domain_newest branch.

PR #585, opened 2020-06-16
Mesh domains!

0leks:
This PR may look big but don't worry, nothing too crazy going on here.
Adds Mesh Domains.
Either remove casting or move it inside of an if statement.
Curvature not figured out yet for meshes so can't have use_normals=1 for mesh domain yet.

TODO run the image use cases and verify that they still work compared to master branch
- ~ellipsoid~
- ~ellipsoid_fd~
- ~left_atrium~
- ~femur~


AtefehKashani:
I have tested this branch for three use cases. The femur use case crashed in multi scale optimization due to :"Description: itk::ERROR: DataObject(0x1d11e60): Attempt to sample at a point outside the narrow band: [30.5882, 2.03564, -36.3751]. Consider increasing the narrow band " Error. 
The left atrium single /multi scale and ellipsoid are done without any error, although the particle outputs aren't equal to the master output. I am not sure is this because of the changes in the optimize library or some randomness is involved.

0leks:
I encountered the same exact issue for Femur multiscale on Windows

Libs/Optimize/ParticleSystem/itkParticleImplicitSurfaceDomain.h
```diff
@@ -71,6 +75,16 @@ class ParticleImplicitSurfaceDomain : public ParticleImageDomainWithCurvature<T,
                                       const PointType &pos) const override;
 
 
+  inline PointType UpdateParticlePosition(PointType &point, vnl_vector_fixed<double, VDimension> &update) const override {
+    PointType newpoint;
+
+    vnl_vector_fixed<float, 3> negativeUpdate;
+    for (unsigned int i = 0; i < 3; i++) { negativeUpdate[i] = -update[i]; }
+    for (unsigned int i = 0; i < 3; i++) { newpoint[i] = point[i] + negativeUpdate[i]; }
```
akenmorris, line 83: What's the purpose of negativeUpdate?  Why not just subtract the update?

akenmorris: @0leks , this is one source of numerical difference between this branch and master.  You have narrowed the update vector from **double** to **float** before the update occurs.  I know the point is only float, but this is affecting the result ever so slightly.

0leks: This is a remnant from debugging when I was trying to compare image optimization vs mesh optimization

0leks: Good point on the double to float cast

akenmorris:
Femur working now!

Issue #549, opened 2020-06-09
Improve use case tiny tests


jadie1:
Improve and note runtime for all use case tiny tests for next release after 5.4

sheryjoe:
Has to be addressed so that we can merge the executable branch @AtefehKashani @cchriste @archanasri

AtefehKashani:
addressed in PL #587

sheryjoe:
@AtefehKashani I am running the left atrium with  --tiny_test and redirecting console output to a log file. It seems that it starts with particle optimization then grooming is done at the end. attached is the log. Any idea what is going on here?
[left_atrium_log.txt](https://github.com/SCIInstitute/ShapeWorks/files/4808417/left_atrium_log.txt)


branch: exec-auto-doc (up to date with master after resolving conflicts)

sheryjoe:
@AtefehKashani  And it also takes some time (may be 10 mins) before printing anything to the console.

sheryjoe:
@AtefehKashani and getting the below error when running --tiny_test with --start_with_image_and_segmentation_data

TypeError: applyCropping() got an unexpected keyword argument 'processRaw'

branch: exec-auto-doc (up to date with master after resolving conflicts)

AtefehKashani:
> exec-auto-doc

There is nothing to do with the tiny test. The use case is not reflected the changes in the GroomUtils in this branch. You will get the same error for the other use cases as well with/out tiny test when start_with_image_and_segmentation_data is on.

AtefehKashani:
> @AtefehKashani I am running the left atrium with --tiny_test and redirecting console output to a log file. It seems that it starts with particle optimization then grooming is done at the end. attached is the log. Any idea what is going on here?
> [left_atrium_log.txt](https://github.com/SCIInstitute/ShapeWorks/files/4808417/left_atrium_log.txt)
> 
> branch: exec-auto-doc (up to date with master after resolving conflicts)

I'm not sure what is happening on your end. Since the optimization can't be run without distance transforms, It first runs the grooming and then the optimization. I guess there is something wrong with printing in the console output.
 I reran the tiny test and it ran in couple of minutes. It started to write out in the console immediately (without 10 mins delay that you mentioned), and also in the order of the pipeline.

sheryjoe:
> > exec-auto-doc
> 
[...]

PR #587, opened 2020-06-17
Usecases

AtefehKashani:
addressed issue #549 
**left atrium, femur** :  addressed for running with less number of particles and less number of  iterations.
**left atrium**: added running with prep data using the downloaded distance transforms.

### DIFF
--- a/Examples/Python/femur.py
+++ b/Examples/Python/femur.py
@@ -101,7 +101,8 @@ def Run_Pipeline(args):
         if args.tiny_test:
             files_img = files_img[:3]
             files_mesh = files_mesh[:3]
-
+            args.use_single_scale = True
+            args.interactive = False
         # run clustering if running on a subset
         if args.use_subsample:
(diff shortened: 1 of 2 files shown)


cchriste:
This looks good. Does it still test everything? For example, when using single scale? How long do the tiny tests now take to run?

jadie1:
This looks good to me, I added the option to skip grooming in the Femur use case with --start_with_prepped_data and run the tiny test

AtefehKashani:
@cchriste if you run with prep_data it will be done in less than 2 minutes. If you want to test the grooming it takes longer(because of the rigid alignment step), but still will not take longer than 5 minutes.

cchriste:
Please delete this branch if you're done with it.

sheryjoe:
@cchriste can we merge master to exec-auto-doc (branched from exectuable_improv? I am trying to run tiny test but taking significant time.

sheryjoe:
exec-auto-doc  is now up to date with master.

PR #596, opened 2020-06-23
Fix cutting plane

akenmorris:
Merge after #591 

This adds an automated test for the cutting plane feature.  It also fixes the feature since at some point some init reorganization broke the order of loading planes and inputs.


cchriste:
How does this cutting plane feature compare to Image::clip? Here is the header, and you can see the implementation here: https://github.com/SCIInstitute/ShapeWorks/blob/executable/Libs/Image/Image.cpp

```
  /// sets values on the back side of cutting plane (containing three non-colinear points) to val (default 0.0)
  Image& clip(const Point &o, const Point &p1, const Point &p2, const PixelType val = 0.0);

  /// sets values on the back side of cutting plane (normal n containing point p) to val (default 0.0)
  Image& clip(const Vector &n, const Point &q, const PixelType val = 0.0);
```

akenmorris:
> How does this cutting plane feature compare to Image::clip?

It’s pretty unrelated. These cutting planes impose a constraint on particle movement within the optimizer.

PR #601, opened 2020-06-24
Rescale update vector to original magnitude after projection to surface tangent

akenmorris:
This is something I discussed with @prafulag.

The goal here is to restore the original update vector magnitude while still projecting to the tangent plane.

This is not ready to merge, but I'm making the PR for people to look at it.

We may need to adjust the number of iterations on some of the use cases.


sheryjoe:
This makes sense to me. I assume this could reduce the number of iterations. It is worth running some benchmarks to assess this hypothesis.

akenmorris:
Yes, it should restore the previous speed of particle spread.

I will have to experiment with all of the use cases to adjust the iterations.

prafulag:
I am certainly curious to know if that improves speed. Thanks Alan for taking this on.

Regards,
Praful Agrawal

sheryjoe:
@akenmorris are we still planning to merge this?

akenmorris:
Eventually, yes.  It's going to require a lot of testing and likely re-tuning of all of the use cases.

medakk:
@sheryjoe  I don't believe we intend to merge this branch. The rescaling for meshes has already been merged to release. I tested that and it works well.

akenmorris:
@medakk , correct.  @sheryjoe , this requires more testing for image based optimization.

akenmorris:
Close for now.

PR #605, opened 2020-06-25
Optimize utils cleanup

0leks:
moved some copy pasted code to a helper function


sheryjoe:
@0leks where are distances called/computed? sampling rely on particle-to-particle distances/interactions.

PR #609, opened 2020-06-27
Correspondence mode enum

0leks:
Replace magic numbers with enum.

Not that there was one case where the mode would be set to 2.
There are no comments describing what 2 means so I named it UnknownEnergy.

Tested on ellipsoid


akenmorris:
What's the effect of the unknown energy (mode == 2)?  

It looks like in SetCorrespondenceMode that it is the same as mode == 0, which is mean energy

But there is a difference in Optimize::ComputeEnergyAfterIteration:

```
 if (m_sampler->GetCorrespondenceMode() > 0) {
    corrEnergy = m_sampler->GetLinkingFunction()->EnergyB(0, 0, m_sampler->GetParticleSystem());
  }
```

Any ideas @sheryjoe, @prafulag  ?

It looks like it's only set to 2 when `<use_shape_statistics_in_init>` is on and the number of attributes per domain (m_attributes_per_domain) is > 0, and mesh based attributes is off.

This doesn't make sense to me.  I thought you had to use m_mesh_based_attributes if you wanted to use attributes.

akenmorris:
Should we just get rid of this mode=2?  I think at a certain point we just need to remove stuff that no one understands and no one knows how to use.

0leks:
I'm all for removing mode 2

prafulag:
+1

Regards,
Praful Agrawal

PR #610, opened 2020-06-29
anatomyPairsToSingles throws exception

0leks:
anatomyPairsToSingles throws exception instead of printing error and continuing to try to run.
Also moved param checking to top of function instead of inside for loop.


akenmorris:
I merged master into this branch and fixed the conflicts.  It looks good to me.  Any comment, @cchriste ?

cchriste:
I looked at it but have no experience with the work so not sure what effect
it might have. Seems superficially fine to me though.

PR #617, opened 2020-07-05
Optimize multiscale

akenmorris:
Multiscale using shape statistics internal to the optimizer.

At this point, I made several modifications to ensure identical execution to the mutli-run based multi-scale.  This includes adding operations that intentionally drop precision and add and remove particles from neighborhoods to ensure that later stages run in the same way that separate runs would.  After this PR is vetted and merged, I recommend that we remove these steps.

Needs full multiscale verification on:

- [x] Ellipsoid
- [x] Left Atrium
- [x] Femur
- [x] Ellipsoid Mesh


akenmorris:
Ellipsoid mesh is not identical between the two branches on a diff of the particles, but it is equivalent:

ellipsoid_mesh_multi

akenmorris:
Left atrium is also not identical (points are very close, but not identical).  The model seems equivalent:

left_atrium_multi

akenmorris:
Femur results are identical (diff on particle files)

akenmorris:
I've removed the stuff that was only added to make the internal multiscale more like the python re-run based approach.  Now comparing the differences:

akenmorris:
ellipsoid mesh:

image

akenmorris:
Left Atrium:

image

akenmorris:
Femur:

image

akenmorris:
Ellipsoid:

image

akenmorris:
Yes, it runs some initialization and some optimization iterations for each split.  This replaces the previously python script functionality where ShapeWorks was run multiple times, feeding the previous output as the input.

akenmorris:
What do you mean you don't see this behavior happening?

HeavenlyBerserker:
Just curious if this happens by default, or do we have to set it to true as a parameter?

akenmorris:
It's off by default.  You need to manually enable it and choose a number of particles to start the multiscale operation on.  (e.g. 32, 64, etc).  It can't be done at 2-8 particles since there are no shape statistics yet.

HeavenlyBerserker:
Ah, I see. I was confused cause Shireen said it was enabled by default, which I wasn't seeing from the code nor had I observed it before. This makes more sense. Thanks!

akenmorris:
Well, I believe that's the default operation for most of the python use cases.

PR #623, opened 2020-07-13
Studio: optimization parameters

### LINKED ISSUE #544: Add remaining optimizer parameters to Studio interface
Starting with all the simpler settings like <use_normals>.  Other parameters like supplying meshes with each need to be extended through the Project class.

### LINKED ISSUE #606: Expose procrustes scaling trigger in Studio

### LINKED ISSUE #607: Expose use_normals in Studio
Use normals flag as well as normals strength (default 10 probably).

### LINKED ISSUE #620: Studio: multiscale option
Now that the optimize library can run in a multi-scale mode where shape statistics are used after N particles, this option should be available in studio.

### LINKED ISSUE #621: Studio: when procrustes is in used, local particles need to be displayed with original/groomed
I think right now we are showing the world particles regardless.  So when the user is showing the original or groomed image, the world particles don't sit on the surface:

world

akenmorris:
This PR adds a number of common optimization parameters to the Studio control box:

Before:

controls_old

After:

image

Resolves #620 
Resolves #607 
Resolves #606 
Resolves #544 

This also fixes #621 

Before:

world

After:

particles_fixed


AtefehKashani:
I have tested this branch for LA for different setting of parameters. It is consistent with master. 
The only thing I noticed is that, the maximum value for "multiscale Start" parameter is 99 and I couldn't set it to any value greater than 99.

akenmorris:
I've adjusted the ranges for all of the optimization parameters.

PR #639, opened 2020-07-23
Scale/Rotate command/function fixes, new resize command/function

cchriste:
biggest items:
- added resize command
- simplified resample command
- fixed scale and rotate commands
- added use cases as options, not merely strings to python example
- updated and added and improved unit tests for image and executable

other:
- un-tabified conda_installs.sh
- improved readability of info command output
- added text to shapeworks --version where release version can be added (and corresponding github issue)

### DIFF
--- a/Applications/shapeworks/Commands.cpp
+++ b/Applications/shapeworks/Commands.cpp
@@ -1066,7 +1066,7 @@ void Compare::buildParser()
   parser.add_option("--name").action("store").type("string").set_default("").help("Compare this image with another");
   parser.add_option("--verifyall").action("store").type("bool").set_default(true).help("Also verify origin, spacing, and direction matches [default: true]");
   parser.add_option("--tolerance").action("store").type("double").set_default(0.0).help("Allowed percentage of pixel differences [default: 0.0]");
-  parser.add_option("--precision").action("store").type("double").set_default(1e-12).help("Allowed difference between two pixels for them to still be considered equal [default: 0.0]");
+  parser.add_option("--precision").action("store").type("double").set_default(1e-12).help("Allowed difference between two pixels for them to still be considered equal [default: 1e-12]");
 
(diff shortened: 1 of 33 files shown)


cchriste:
This is the original image.  
original_image
These images show the fixes to scale and rotate.  
The first column shows the results of scale, with the old version on top and the fixed on the bottom. Note the original image isn't quite centered, so in the bottom the white part is still a little to the left, but it is correctly scaled from the actual center of the original image.  
The second column shows a rotation of -PI/3, old version on top, fixed version on bottom.
scale_rotate_fixes_comparison

akenmorris:
Can you explain a little more about what was broken before with scale and rotate?  Is the the point of reference?

cchriste:
> Can you explain a little more about what was broken before with scale and rotate? Is it the point of reference?

These operations are intended to be performed using the center of the image, and the way that was being set was incorrect. Instead, it now saves the current origin, uses Image::recenter() to change it (what it should have done anyway), and changes it back at the end. The evidence for the origin being set incorrectly is that the rotations and scales appear to be based on the wrong place in the image. This was because it was being set incorrectly.

akenmorris:
Sounds good.  Does this PR change the use cases at all or were they not using these?

cchriste:
> Sounds good. Does this PR change the use cases at all or were they not using these?

The use cases only use isoresample, which was a call to resample that only specifies new physical spacing and allowed the function to automatically compute the new dims, which is what the function will always do now. If explicit resizing is desired, there is now a separate function called Image::resize.

cchriste:
I just made another fix for the info command. The parser wasn't clearing old user-specified arguments, though it was still resetting them to their default values so it was a very invisible issue except for the one command for which we query the number of arguments that were set by the user.

cchriste:
Are we good to go on this?

PR #642, opened 2020-08-01
Refactor itkMaximumEntropySurfaceSampler and itkMaximumEntropyCorrespondenceSampler

akenmorris:
Refactored these Sampler classes into a new class called simply "Sampler".  The parent class (if anyone could ever remember which was which), was never instantiated or used, so there was no reason to have a hierarchy that split the functionality of what was effectively a single class (I'm sure it made sense at one point in early development).  

Also removed the template over ImageType.  It makes even less sense now since one can use it on meshes without images.

Also pulled this class out of ITK.  It's no longer inheriting from any ITK objects, no longer using all the class/pointer macros.  Replaced pointers to this class with std::shared_ptr.

Here's the change in build time (advantage to developer standpoint):

Both are parallel build, 8 cores:

master:
one line change to Optimize.cpp : 24 seconds build
one line change to itkMaximumEntropySurfaceSampler.txx : 32 seconds build

refactor_templates:
one line change to Optimize.cpp : 8 seconds build
one line change to Sampler.cpp : 22 seconds build


sheryjoe:
Does this refactoring affect functionality? Is this tested on a segmentation and mesh use cases (examples)?

akenmorris:
There should be no change to functionality.  I haven't yet tested all the use cases to be sure.

akenmorris:
Left Atrium yields identical results.

akenmorris:
ellipsoid_mesh is identical as well.

PR #442, opened 2020-03-17
API Consolidation: Images

archanasri:

PR #651, opened 2020-08-03
Bug fix for LA use case

archanasri:


Examples/Python/GroomUtils.py
```diff
@@ -292,10 +292,13 @@ def applyRigidAlignment(outDir, inDataListSeg, inDataListImg, refFile,
             print("CMD: " + " ".join(cmd))
         subprocess.check_call(cmd)
 
+        size = getInfo(ref_tpdtnrrdfilename, "dims")
+
         cmd = ["shapeworks", 
                "read-image", "--name", seginname,
-            #    matrix created by icp is not used
+            #    matrix created by icp is not used so we use resize since icp resizes all images to the reference image which is a requirement for cropping (next step)
```
cchriste, line 299: Can we please create an issue about icp not being used that points to these specific lines in GroomUtils (if we don't already have one)?

Examples/Python/left_atrium.py
```diff
@@ -68,8 +68,8 @@ def Run_Pipeline(args):
         fileList_seg = sorted(glob.glob(parentDir + "segmentations/*.nrrd"))
 
     if args.tiny_test:
-        fileList_img = fileList_img[:2]
-        fileList_seg = fileList_seg[:2]
+        fileList_img = fileList_img[:5]
```
cchriste, line 71: I like that these extra images are being added, but perhaps we should include a comment about why (because the first two images are the same resolution so the issue being addressed by this pull request only comes up when images are diff res are included).

Examples/Python/GroomUtils.py
```diff
@@ -304,16 +307,16 @@ def applyRigidAlignment(outDir, inDataListSeg, inDataListImg, refFile,
         if processRaw:
             cmd = ["shapeworks", 
                    "read-image", "--name", rawinname,
-                #    matrix created by icp is not used
+                #    matrix created by icp is not used so we use resize since icp resizes all images to the reference image which is a requirement for cropping (next step)
```
cchriste, line 310: Just to double check: we haven't modified icp in any way, correct? So it still does what it's supposed to do, just it's used.

Examples/Python/GroomUtils.py
archanasri: I'll create the issue

Examples/Python/GroomUtils.py
archanasri: Yes

sheryjoe:
@archanasri can you describe the bug? why is resize used instead of icp? both have different purposes.

@AtefehKashani can you review this?

Examples/Python/left_atrium.py
archanasri: Done

archanasri:
@sheryjoe The bug was that a boundingbox couldn't be calculated because image sizes were not matching. This bug was not present earlier because ICP used to set all image sizes to the reference image size even though the transformation matrix was not used. So we are using resize here instead to achieve the same result.

sheryjoe:
> @sheryjoe The bug was that a boundingbox couldn't be calculated because image sizes were not matching. This bug was not present earlier because ICP used to set all image sizes to the reference image size even though the transformation matrix was not used. So we are using resize here instead to achieve the same result.

@archanasri Resize should not be used instead of ICP as a quick-and-dirty way to make the bounding box work. This could introduce artificial variations (e.g. scale) that will be captured in the subsequent modeling (optimize) step. Shapes (segmentations) should be first aligned to make all shapes share the same physical coordinate system, hence all segmentations will inherit the same size/spacing/origin of the reference segmentation used for alignment.

archanasri:
> @archanasri Resize should not be used instead of ICP as a quick-and-dirty way to make the bounding box work. This could introduce artificial variations (e.g. scale) that will be captured in the subsequent modeling (optimize) step. Shapes (segmentations) should be first aligned to make all shapes share the same physical coordinate system, hence all segmentations will inherit the same size/spacing/origin of the reference segmentation used for alignment.

@sheryjoe Using the working version of ICP gives us undesirable results as seen in #442. If we don't want to use resize (which sets the size to the reference segmentation), the other option is to apply an identity matrix and set the size/spacing/origin/direction to the reference segmentation

sheryjoe:
@archanasri This could be related to input for ICP, please refer to Alan's comment in #653

akenmorris:
@cchriste , is this ready to be merged?

cchriste:
> @cchriste , is this ready to be merged?

This closely relates to our conversation in yesterday's meeting. In that meeting significant changes were proposed in order to avoid resampling until the last possible time, thus reducing errors inherent in that operation (see new issue #660). Currently Image::resize calls resample (also note new issue #659) and also changes physical pixel size (spacing), which will affect any downstream operations that assume/require spacing of different images to be the same. 

The ICP transform only rotates and translates (see
https://vtk.org/doc/nightly/html/classvtkLandmarkTransform.html#a74b9f22a62dc143194a2d8f6d06f6731 called by ImageUtils::createIcpTransform). It doesn't scale, but instead applies this transform along with forcing the result to have the same size/spacing as the target image, so after GroomUtils.applyRigidAlignment all the (resampled) images are the same size with the same spacing. 

If we want the same results, instead of calling resize we can call the Image::resample command that takes both new dims and new spacing, which isn't currently available as a subcommand.

In summary, to accomplish the intent of this fix while maintaining the _current_ grooming process, we must:
1. update shapeworks resample subcommand to enable specification of dims, origin, spacing, and coordsys and call Image::resample (see new issue #661)
 - note that it we want to mirror current functionality *identically* we'll need to call the version which uses linear interpolation if only new dims are specified, and otherwise use nearest-neighbor interpolation (#659 will make this more straightforward), a questionable difference that was likely made to keep things identical to differences to the old executables
2. change GroomUtils.applyRigidAlignment to call this function instead of resize

There are other minor changes/cleanups in this pull request that would be nice to merge.

cchriste:
I think this is ready to merge. It's slightly irrelevant since it's all merged with open_meshes, but it'd be nice to merge with master and close the pull request, plus it gives others an opportunity to ensure there's nothing wrong with this part separate from the open_meshes, a much bigger change. Anyone have issues with this?

akenmorris:
It sounds fine to me.  So with this PR, the LA use case is as it was before?

cchriste:
> It sounds fine to me. So with this PR, the LA use case is as it was before?

Unfortunately, it's _very_ difficult to say. The new icp command wasn't resampling the images as the old one was, and the old one wasn't performing the actual transform as the new one does. The old one has been out of the game (i.e., not in master) for some time. The only reason the tiny test use cases were succeeding was because they weren't including images of different sizes (that's now resolved by simply including a few more images than before). The first fix (after quite some time) was to resize images, but that also changed their spacing which seemed to cause trouble down the line. The final fix has been to provide the itk `resample` command directly, so the `applyRigidAlignment` function in GroomUtils now does _exactly_ what the old `ICPRigid3DImageRegistration` command did: a resample to ensure all the images have the same size and spacing as some selected "median" image.

Therefore I believe the LA use case is as it was "before" where "before" is defined as before the very big merge of shapeworks subcommands. But it's a very difficult fix to verify objectively since many other bugs have been fixed along the way, so simply checking out code from before this merge won't result in the same intermediate images. I'm afraid this puts it in the same category as *tons* of other changes to our code, and while I don't consider this to be the best way to work, I think the results seem correct and therefore same as "before."

cchriste:
This pull request includes the new, more general resample command, thus it also resolves issue #661.

Issue #614, opened 2020-06-30
Consistent groom output for all use cases


sheryjoe:
<img width="225" alt="image" src="https://user-images.githubusercontent.com/3253036/86161417-74905e80-baca-11ea-969d-0cf14cc497e5.png">

Also consider the "groom-optimize-analyze" naming convention rather than prepped output.

AtefehKashani:
addressed in PR #656

Issue #593, opened 2020-06-20
RunUseCase help does not show defaults for input args


sheryjoe:

jadie1:
Most of the input args are flag, they are on if included or off if not. Should we specify the default is off? 

The only args which require values are use_case, use_subsample, and  shapeworks_path.

- use_case does not currently have a default value. It prints the help if RunUseCase is called without a use case. Should we set a default value? I think that could lead to confusion.
- use_subsample also does not have a default currently. We could set a default value for when a user adds the tag but doesn't specify the number, but what should that be?
- shapeworks_path has a default value which is already specified in the help

![image](https://user-images.githubusercontent.com/54382613/91647338-f6433080-ea16-11ea-8704-0c3c02838fdf.png)

sheryjoe:
- Agreed on the use_case. It will be confusing to have a default.
- We could use 3 samples as a default for use_subsample if the user didn't specify the number but print a warning that this number was not given and this default will be used (probably with a "Press any key to continue ... " message to confirm that this is what the user wants).

cchriste:
- flags are only enabled if set. It's implicit they aren't used if they're not set. Not sure how much more clearly we could/should be about this... _leave as is_
- agreed: no default use case should be provided
- use_subsample, start_with_prepped_data, start_with_image_and_segmentation_data, and use_single_scale are all specific to subsets of use cases. I don't currently have a suggestion on how to clarify this to users, but it _is_ confusing and should be addressed
  - regarding use_subsample, it would be better for the help tio suggest a default value since it's not a flag, but a manual specification of... the number of samples to use? It's not even clear looking at the code what this is doing. Here's the code:

jadie1:
I added the default arg for use_subsample to pull request #656

PR #656, opened 2020-08-07
Usecases

AtefehKashani:
- updated LA parameters
- updated the parameters in documentation and new images and added the video files for pca
- changed the prepoutput and pointfiles to groomed and shape_models to be consistent with the data portal.
- changed the ellipsoid outputs file to be consistent with the other use cases issue #614


akenmorris:
Running now...

Examples/Python/ellipsoid.py
```diff
@@ -88,28 +88,28 @@ def Run_Pipeline(args):
     if int(args.interactive) != 0:
         input("Press Enter to continue")
 
-    parentDir = 'TestEllipsoids/PrepOutput/'
+    parentDir = 'TestEllipsoids/groomed/'
```
akenmorris, line 91: Do we have the wrong zip file?  On a clean system (GitHub Action), this is failing.

@AtefehKashani @0leks

akenmorris: I guess this isn't the right place.  But in any case, the automated test doesn't work while looking for some file.

akenmorris: Ok, it was the path that the test was looking for output.

AtefehKashani: @akenmorris  Sorry I missed you r massages earlier. Did this solve the problem?

akenmorris: I committed the fix.

akenmorris:
I've run it and the results look ok.  I'm not sure what we're going for here:

image

sheryjoe:
@jadie1 @AtefehKashani when start_with_prepped_data tag is used, which distance transforms are used? The ones that are packaged/released with the use case or the previously groomed data run by the user on his local machine?

jadie1:
@sheryjoe  When start_with_prepped_data is used it should use the distance transform released with the use case. It looks like that is the case for the Femur and the LA. But the Ellipse use case still calls the distance transform function when start_with_prepped_data is used, so that probably needs to be addressed.

jadie1:
I added fixes for issues #593 and issue #667 and I tested the ellipsoid, femur, and left_atrium use cases.
I think the ellipsoid_fd use case needs to be updated still both in the python code and the data portal.

jadie1:
@riddhishb Could you update the ellipsoid fixed domain use case on this branch to address issue #614 both in ellipsoid_fd.py and on the data portal? You can look at the ellipsoid use case to see how the folders have been renamed/restructured.

jadie1 (commented):
I believe the naming is now consistent across all use cases and I verified that all use cases are running as expected. @sheryjoe I think this is ready to merge.

sheryjoe:
@jadie1  Let's merge. Also, these changes should propagate to new use cases 
- lumps and femur_mesh @medakk
- sphere and femur with cutting planes @HeavenlyBerserker

akenmorris:
@jadie1 Can you help resolve this conflict?

<<<<<<< usecases
    meshFiles = meshFiles[:15]
    print(meshFiles)

    pointDir = parentDir + 'shape_models/'

=======
    pointDir = parentDir + 'PointFiles/'
>>>>>>> master


Is it PointFiles or shape_models?

jadie1:
It should be shape_models in all use cases now. Likewise it should be
"groomed" rather than "PrepOutput" now. This is consistent with the new
portal data.

cchriste:
Can this be merged now?

0leks:
Question: were all of the dataset versioning issues caused by the switch PrepOutput to groomed and PointFiles to models?

jadie1:
@0leks I'm not exactly sure which issues you are talking about but I think all current issues in master will be resolved when this gets merged in.

PR #658, opened 2020-08-13
Studio splash

akenmorris:
Adds a splash screen like Seg3D has.

studio_splash


akenmorris:
Obsolete, included in studio_feature_map

PR #689, opened 2020-09-07
complete femur mesh

medakk:


sheryjoe:
@medakk This PR only includes processing the full dataset. Can we also include the mesh grooming steps for this use case?

medakk:
There's [this CPP grooming file](https://github.com/SCIInstitute/ShapeWorks/blob/d2285e8473cfcaec26caa55a3163780bdab27f8c/Libs/Mesh/GroomFemur.cpp). @0leks says that the shapeworks API doesn't have support for all the mesh grooming operations he used yet.

sheryjoe:
> There's [this CPP grooming file](https://github.com/SCIInstitute/ShapeWorks/blob/d2285e8473cfcaec26caa55a3163780bdab27f8c/Libs/Mesh/GroomFemur.cpp). @0leks says that the shapeworks API doesn't have support for all the mesh grooming operations he used yet.

@archanasri @cchriste This is a good example for mesh grooming. I think it shares couple of steps with the femur use case that uses image domains (e.g., reflect mesh, align mesh .. etc).

@medakk For the release, we could consider only demonstrating optimization on meshes given groomed meshes and release the femur_mesh dataset (groomed) and document this in its readme.

cchriste:
Is there anything in particular holding up this pull request? I created an item noting the GroomFemur app in the ShapeWorks Consolidation google sheets so it can be prioritized in the creation of our concrete Mesh grooming class.

sheryjoe:
I think we can merge this.

PR #693, opened 2020-09-15
Femur no images

jadie1:
Updated femur use case so when --start_with_image_and_segmentation_data is used it grooms images and meshes and when it is not it grooms only meshes.
When not using image, origin and size are inferred from meshes and spacing is isotropic unless the user specifies otherwise.
I updated the documentation to reflect this.

I have already merged this with the usecases branch so it might be better to merge it after pull request #656


cchriste:
Can this be merged now (after #656)?

jadie1:
@cchriste It's ready to go but I don't know if anyone else has looked at it.

jadie1:
I have fixed the bugs @HeavenlyBerserker found and the femur use case on this branch is running except for in interactive mode. In interactive mode there is a version issue with VTK 9 and Python VTK 8.2.

HeavenlyBerserker (approved):
Seems to be in order.

Tests now with changes:
  femur
    Without images
       Test subsamples with interactive flag
    With images
       Test subsamples with interactive flag
          Option 2
          Option 1

akenmorris:
Ok, let’s merge the DeepSSM branch first that will fix the VTK stuff.

jadie1:
@sheryjoe @akenmorris I think this is ready to merge

PR #694, opened 2020-09-15
Upgrade to vtk9

akenmorris:
FYI, requires re-running build_dependencies.sh

We'll see if the GitHub Actions all works.


jadie1 (approved):
I built successfully and ran all of the use cases. It seems like there are no issues with the upgrade.

cchriste:
I'm 2/3 of the way finished with upgrade_deps, but I think merging this (and the unrelated meshusecases branch) first might be the best path forward. Any dissenting opinions?

cchriste (approved):
👍

PR #701, opened 2020-09-22
Pybind Optimize

akenmorris:
This adds a basic python API for the Optimize class.  This API works both ways in that the user can specify a python file as a parameter to the XML.  It also provides a callback interface on each iteration.


cchriste:
This fails to successfully build the shapeworks python bindings (which also happen to fail to build altogether on linux, but that's another issue). There is some error about the python GIL mutex failing to be acquired. It works for pybind branch, but not for pybind_optimize. Here's a supposed fix, so I'm tossing this one back over the fence for you:

CMakeLists.txt
```diff
@@ -145,6 +145,7 @@ include_directories("${CMAKE_CURRENT_SOURCE_DIR}/Libs/Common")
 include_directories("${CMAKE_CURRENT_SOURCE_DIR}/ExternalLibs/optparse")
 
 if(Build_PythonAPI)
+  SET(PYBIND11_PYTHON_VERSION 3.7)
```
medakk, line 148: I don't understand it, but according [to this](https://github.com/pybind/pybind11/issues/587), this should be

```
set(PYBIND11_PYTHON_VERSION 3.7 CACHE STRING "")
```

cchriste: Caching is optional, so this isn't important.
I'm more interested in how it finds the correct version of Python compared to the main pybind branch.

Regarding the failure itself, that's because external python code can only be called from the main thread, and it's likely the googletest code uses separate threads. I've dealt with this in other projects by painstakingly creating a queue and executing those items with the main thread. I don't know how it could be done with googletest.

cchriste: I can't remember where we're getting googletest, but here's some muck about using it single threaded:
https://stackoverflow.com/questions/29335446/command-line-options-to-force-googletest-to-run-in-single-thread

cchriste: So... it finds the correct python because I fixed it awhile back to do that, which is why there is the set pybind11 python version, etc, etc. 

The failure to acquire the python gil seems to have been a problem from the very beginning. Was this work ever tested from Python?

cchriste:
The failure to acquire the python gil seems to have been a problem from the very beginning. Was this work ever tested from Python?

```python
>>> import shapeworks
Fatal Python error: PyMUTEX_LOCK(_PyRuntime.ceval.gil.mutex) failed

Abort trap: 6
```

It's the same error even if you try to import something nonexistent, like `from shapeworks import Foo`.

cchriste (changes requested):
This doesn't work from Python, and it's not clear how to test it independently.

Please add unit tests to verify and demonstrate the functionality created here for both the shapeworks executable (or wherever it's used) and the Python module (using PythonTests.cpp, for the additions to ShapeworksPython.cpp).

akenmorris:
> Was this work ever tested from Python?

Yes.  It's interesting, it still works fine for me if I build against my system python.  It's only when I use python from conda that I get this error.  I will look into it.

Libs/Optimize/CMakeLists.txt
```diff
@@ -18,14 +18,17 @@ target_link_libraries(Optimize
   tinyxml
   Utils
   Particles
+  #pybind11::embed
```
archanasri, line 21: Why is it commented?

Libs/Optimize/CMakeLists.txt
```diff
@@ -18,14 +18,17 @@ target_link_libraries(Optimize
...
   ${ITK_LIBRARIES}
   ${VTK_LIBRARIES}
   vgl
   vgl_algo
+#  ${PYTHON_LIBRARIES}
```
archanasri, line 26: Why is it commented?

akenmorris:
This is ready to be looked at again.  I've added automated tests for both python->cpp and cpp->python.  After GH gets fixed, and merged back to here, we'll see if it runs other places like windows.

akenmorris:
Embedded python on windows leads us to this error:

> Fatal Python error: initfsencoding: unable to load the file system codec
> ModuleNotFoundError: No module named 'encodings'

Discussed here:

https://github.com/pybind/pybind11/issues/1930

https://stackoverflow.com/questions/58075887/pybind11-python-embed-on-windows-10-fatal-python-error-initfsencoding-unable

It seems we need to set PYTHONHOME as well.  Unless this issue gets resolved:

https://github.com/pybind/pybind11/issues/2369

akenmorris:
Working on windows now.

cchriste:
merged with pybind

cchriste:
Spoke too soon. I hadn't fetched latest.

PR #702, opened 2020-09-22
upgrade all conda and pip deps

cchriste:
This also upgrades existing windows deps (many items missing from conda_installs.bat, but if it ain't broke...).
We might want to merge this with the vtk 9 upgrade and merge to master all at once (that's what I'll try next anyway).


build_dependencies.sh
```diff
@@ -280,6 +280,7 @@ show_shapeworks_build()
   fi
 
   echo "cmake -DCMAKE_PREFIX_PATH=${INSTALL_DIR} ${OPENMP_FLAG} -DBuild_Studio=${BUILD_GUI} -Wno-dev -Wno-deprecated -DCMAKE_BUILD_TYPE=Release ${SRC}"
+  echo "Have a nice day!"
```
medakk, line 283: 🎉 🎉

cchriste:
Vtk is building its own Python bindings, and in order to get the tests to be able to load them from their non-standard install location, the testing setup was modified, and since at least one platform is succeeding (Linux), I'll share this notes about the outstanding issues related to this:
- it's tricky to import vtk libraries on OSX
 o not quite working for PythonTests or setupenv script, both of which are working for the other two platforms
 o best solution (for all platforms and for many other issues) would be to install to $CONDA_PREFIX
- the path to both the dependencies install and Qt can be different for every installation
 o it's currently set to work with the path used for github actions
- the ITK Python bindings are not so easy to get working
 o already a separate issue, so not blocking this pull request

We could just split out the vtk stuff so the conda updates can be merged. This would mean re-adding vtk to the conda installs, and I might be forgetting, but I think that may have resulted in a dependency conflict which was what drove me to do all this building and loading of vtk python bindings in the first place (well, that and idealism).

I'm working on getting the Windows test failure resolved. It's due to a forward-vs-backward slash issue in cmake. After that, the issues above are the ones still outstanding. I see loading the vtk python module on OSX as just a little more work munging paths, and then the issues of hard-coded paths in the build another challenge to overcome with cmake, and finally teaching people to run `setupenv` before using shapeworks and/or its Python module the final bit that will just need to be documented, and is already something of an issue, one that we're mostly facing by using the path to shapeworks bin as the argument to RunUseCase.py and nobody really using the command line directly (and therefore not noticing `shapeworks` isn't in their path).

PR #705, opened 2020-09-23
Add getFileList api to DatasetUtils

0leks:
Add API to get file list of a dataset


cchriste (approved):
Looks great, tested* and works great too. 
*tested independently; didn't try examples, but I could still download a dataset. Github actions will affirm they still work.

PR #718, opened 2020-09-29
Fix main/unit tests

### LINKED ISSUE #717: All unit tests are not being executed and thereby "succeeding"
Seems to be an issue that came from the open meshes branch, but not positive. It started about a month ago.
Assigning it high priority and absolutely imperative for next release. These tests are very important.

akenmorris:
This PR fixes the unit tests by removing the 'main' from GroomFemur.cpp from LibMesh.

Resolves #717

### DIFF
--- a/Libs/Mesh/CMakeLists.txt
+++ b/Libs/Mesh/CMakeLists.txt
@@ -4,7 +4,6 @@ set(Mesh_sources
  Mesh.cpp
  meshFIM.cpp
  MeshUtils.cpp
- GroomFemur.cpp
  )
 
 set(Mesh_headers
--- a/Testing/MeshTests/MeshTests.cpp
+++ b/Testing/MeshTests/MeshTests.cpp
@@ -6,7 +6,9 @@ using namespace shapeworks;
 
 //---------------------------------------------------------------------------
 TEST(MeshTests, coverage_test) {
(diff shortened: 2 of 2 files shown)


cchriste:
Here's the failing test:

cchriste:
This is the place the test is failing, in the Mesh::coverage function:
```c++
  FEMesh* surf1 = importer.Load(this->mesh);
  FEMesh* surf2 = importer.Load(other_mesh.mesh);
  if (surf1 == nullptr || surf2 == nullptr) { throw std::invalid_argument("Mesh invalid"); }
```
We could just omit this test from MeshTests.cpp for now (although it's the only mesh test we currently have in master). I would suggest also adding GroomFemur as a separate executable in Libs/Mesh/CMakeLists.txt -- not sure who might be using it, but it always helps to ensure things keep compiling.

cchriste:
Disabling the one broken unit test and created a separate issue for it: #722.

cchriste:
> We could just omit this test from MeshTests.cpp for now (although it's the only mesh test we currently have in master). I would suggest also adding GroomFemur as a separate executable in Libs/Mesh/CMakeLists.txt -- not sure who might be using it, but it always helps to ensure things keep compiling.

It turns out GroomFemur was _already_ an independent executable.

cchriste:
It's great to see all those unit tests running again. Merged.￼

PR #743, opened 2020-10-02
cleaning up dataset download/unzip logic for ellipsoid,lumps

0leks:
The data folder should now be considered read-only for these three use cases (the usecase won't re-unzip unless the data folder is missing)
Tested:
- [x] ellipsoid
- [x] ellipsoid_fd
- [x] lumps

### DIFF
--- a/Examples/Python/.gitignore
+++ b/Examples/Python/.gitignore
@@ -1,11 +1,12 @@
 TestEllipsoidsFD/
 TestEllipsoids/
 TestEllipsoidsMesh/
+TestLeftAtrium/
 TestFemur/
 TestFemurMesh/
 TestScapula/
 TestCan/
 TestLumps/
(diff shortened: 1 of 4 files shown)


jadie1 (commented):
I checked that the data folder is correctly downloaded for all three use cases and I verified that if it has already been unzipped it will not unzip again. Anything else to check?

0leks:
@jadie1 Nothing else to test

jadie1 (approved):

jadie1:
@sheryjoe @akenmorris Can we merge this? I have tested it and I am updating the other use cases to have these changes as well, I'll make a new PR for that soon.

sheryjoe:
i think we can merge this. @akenmorris ?

akenmorris (approved):

PR #752, opened 2020-10-06
Fix ellipsoid mesh

akenmorris:
This PR re-enables the ellipsoid_mesh use case and fix Studio for the output.


jadie1 (approved):
This is worked perfectly!

PR #775, opened 2020-10-13
Use case path fix

jadie1:
Fixies paths for use cases when --start_with_prepped_data is used
Fixes output paths of femur_cut use case


akenmorris:
ellipsoid --start_with_prepped_data works!

jadie1:
With the last change I just made updating the ellipsoid zip, all use cases should work with the --start_with_prepped_data tag (ellipsoid, ellipsoid_cut, femur, femur_cut, and left_atrium). I just tested all of these (at least tiny_test) again and they seem to be working.

PR #776, opened 2020-10-13
Fix ellipsoid evaluation

medakk:
Fixes paths, but group differences not working in Studio yet


sheryjoe:
> but group differences not working in Studio yet

why is that?

sheryjoe:
@medakk could you also update doc/new/ssm-eval.md ?

medakk:
I'm not sure. @akenmorris says it used to work at some point. I recall now that this is why I used View2 here instead of Studio.

akenmorris:
> why is that?

See #777

PR #781, opened 2020-10-13
Move zips to LFS as they keep changing and accumulating storage in the history

akenmorris:
Ok with everyone?


jadie1 (approved):

medakk (approved):

HeavenlyBerserker:
What exactly does this mean? You're moving nrrd, vtk, dcm and zip to LFS. This does not change anything for us, right? These files will still be available just the same, just replaced by LFS pointers?

akenmorris:
Just adding zip to the list. Yes, no difference to you since you are already using LFS.

HeavenlyBerserker (approved):

PR #783, opened 2020-10-15
Deploy gh pages

akenmorris:
This PR adds a github action workflow that will deploy the mkdocs on gh-pages for pushes to master branch.


sheryjoe:
@akenmorris how could i test this PR? should I run the deploy mkdocs script in Support?

akenmorris:
I don't think that script will work for you.  It's using the $GITHUB_TOKEN that's given by the GitHub Action.

I tested it by changing the branch to deploy_gh_pages and seeing that it updated the gh-pages.

https://github.com/SCIInstitute/ShapeWorks/runs/1261317769?check_suite_focus=true

sheryjoe:
We should also include  the python autodoc for shapeworks commands before mkdocs deploy.

akenmorris:
Yes, that's a good point.

akenmorris:
Ugh, well, I guess this needs to go in with the build actions since we need 'shapeworks' executable built in order to generate this documentation.  Shall we just put it on one platform?  Linux?

sheryjoe:
> Ugh, well, I guess this needs to go in with the build actions since we need 'shapeworks' executable built in order to generate this documentation. Shall we just put it on one platform? Linux?

Yeah, we can deploy on linux.

akenmorris:
I've moved it there and the checks are now running.

akenmorris:
I think this is ready to merge.

Issue #760, opened 2020-10-09
Concise but descriptive names for RunUseCase arguments 


sheryjoe:
e.g --start_with_image_and_segmentation_data seems too long and too specific to using segmentation.
start_with_prepped_data is not consistent with the groom terminology 

make sure to update the corresponding docs files.

PR #784, opened 2020-10-16
Use case arguments

jadie1:
To address issue #760 I made the following changes:

- Changed `--start_with_prepped_data` to `--skip_grooming`
- Changed `--start_with_image_and_segmentation_data` to `--groom_images`
- Updated RunUseCases.py to print a warning if  `--groom_images` is used with a use case that does not have images.
- Updated the use case documentation for these changes on the mkdocs branch. 

To test run the use cases with these new tags.


Examples/Python/femur.py
```diff
@@ -38,7 +38,16 @@ def Run_Pipeline(args):
     if args.interactive:
         input("Press Enter to continue")
 
-    if not args.start_with_prepped_data:
+    if args.skip_grooming:
```
sheryjoe, line 41: If getting the distance transforms file list is the same for all use cases, it is worth modularizing it into a utils function. For example, get_file_list with arguments directory, extension, is_tiny_test, num_samples.

Examples/Python/femur_cut.py
```diff
@@ -39,7 +39,19 @@ def Run_Pipeline(args):
     if args.interactive:
         input("Press Enter to continue")
 
-    if not args.start_with_prepped_data:
+    if args.skip_grooming:
```
sheryjoe, line 42: same comment about a function for file list.

Examples/Python/left_atrium.py
```diff
@@ -50,7 +50,7 @@ def Run_Pipeline(args):
         fileList_seg= [fileList_seg[i] for i in sample_idx]
         fileList_img = [fileList_img[i] for i in sample_idx]
         
-    if args.start_with_prepped_data:
+    if args.skip_grooming:
```
sheryjoe, line 53: Same here.

sheryjoe (changes requested):

Examples/Python/RunUseCase.py
```diff
@@ -61,6 +61,13 @@
 # make sure the shapeworks executables can be found
 robustifyShapeworksPaths()
 
+image_use_cases = ['femur', 'femur_cut', 'left_atrium']
+if args.groom_images and args.use_case.lower() not in image_use_cases:
```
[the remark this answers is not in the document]
jadie1: Agreed

jadie1:
I made a common function for getting prepped data and addressed all of your comments @sheryjoe
I updated the documentation on the makedocs branch and tested all the use cases.

sheryjoe (changes requested):
get_data (which downloads data) and get_groomed_data (which retrieves the file list in a directory) in CommonUtils do not have function names that reflect their functionality and it seems confusing.

jadie1:
I changed "get_data" to "download_and_unzip_dataset" and I changed "get_groomed_data" to "get_distance_transform_list". Let me know if you can think of a better name for these functions.

sheryjoe:
@jadie1 I don't think that get_distance_transform_list has anything specific to distance transforms. I would make this as a general get_file_list function with arguments for the extension to search for and the path to search in.

sheryjoe (approved):

akenmorris (approved):
Looks good.

Issue #759, opened 2020-10-08
consistent naming for sampler_type


sheryjoe:
Options: `Gaussian`, `mixture`, or `KDE`. Default: `KDE`.

I would use all small cases.

Update data-augmentation.md accordingly.

Issue #758, opened 2020-10-08
Add variance percentage to runDataAugmentation


sheryjoe:
It would be useful to allow the user to control the currently hard card 95% variability explained when num_dim is not given.  

Update data-augmentation.md accordingly.

PR #785, opened 2020-10-16
Data augmentation updates

jadie1:
This PR addresses issue #758 and #759:
- Changes `sampler_type` options to be consistent (lower case): gaussian, mixture, or kde
- When `num_dims` is set to 0 or not specified for embedding, instead of defaulting to selecting the dims so that at least 95% of the variability is preserved, the user selects this portion. They are asked to input this proportion and an example is provided. 
- The documentation on the mkdocs branch has been updated to reflect these changes. 

To test, run data augmentation with different sampler types and num_dims=0.


Python/DataAugmentationUtilsPackage/DataAugmentationUtils/DataAugmentation.py
```diff
@@ -11,20 +11,20 @@
 
 ################################# Augmentaiton Pipelines ###############################################
 
-def point_based_aug(out_dir, orig_img_list, orig_point_list, num_samples, num_PCA=0, sampler_type="KDE", mixture_num=0):
+def point_based_aug(out_dir, orig_img_list, orig_point_list, num_samples, num_PCA=0, sampler_type="kde", mixture_num=0):
```
sheryjoe, line 14: Also, I would use num_pca_modes rather than num_PCA (not consistent and not obvious).

Python/DataAugmentationUtilsPackage/DataAugmentationUtils/DataAugmentation.py
```diff
@@ -11,20 +11,20 @@
 
 ################################# Augmentaiton Pipelines ###############################################
```
sheryjoe, line 12: fix comment typo 
Augmentaiton -> Augmentation

Python/DataAugmentationUtilsPackage/DataAugmentationUtils/Embedder.py
```diff
@@ -39,9 +39,11 @@ def run_PCA(self, num_PCA):
 		eigen_vectors = np.flip(eigen_vectors, 1)
 		# get num PCA components
 		cumDst = np.cumsum(eigen_values) / np.sum(eigen_values)
+
 		if num_PCA == 0:
+			percent_variability = input("\nEnter the proportion of variability to preserve. \nFor example, to preserve at least 95% of variability enter: 0.95\n")
```
sheryjoe, line 44: I would add percent_variability as an input argument with a default 0.95 and use it if num_pca_modes = 0 (default)

Python/DataAugmentationUtilsPackage/DataAugmentationUtils/__init__.py
```diff
@@ -11,7 +11,7 @@
 - aug_type = type of augmentaion to run, 1 is for aug based only on particles, 2 bases aug on images and particles
 - sampler_type = type of distribution to represent embedded data with for sampling
 '''
-def runDataAugmentation(out_dir, img_list, point_list, num_samples, num_dim=0, sampler_type="KDE", mixture_num=0):
+def runDataAugmentation(out_dir, img_list, point_list, num_samples, num_dim=0, sampler_type="kde", mixture_num=0):
```
sheryjoe, line 14: - Isn't num_dim the same as num_pca_modes? I would be consistent in argument naming. 
- I would also expose percent_variabililty as input argument here.
- mixture_num -> num_mixture_components

sheryjoe (changes requested):

sheryjoe:
Please modify docs files accordingly. (data-augmentation.md, deep-ssm.md and deep-ssm-femur.md)

Python/DataAugmentationUtilsPackage/DataAugmentationUtils/__init__.py
jadie1: I am not sure how to allow either num_dim or percent_variabilty as an argument... The use would have to specify one and set the other to a value like -1 or something to indicate not to use it. Do you have an idea of how to do this?

sheryjoe: percent_variability can be set as 0.95 by default. num_pca_modes to 0 so the default behavior would be finding the number of modes that capture 95% of the data variability.

Python/DataAugmentationUtilsPackage/DataAugmentationUtils/DataAugmentation.py
jadie1: I made this num_dim everywhere so that it is consistent if we ever add another form of embedding.

jadie1:
I added the `percent_variability` option as advised, addressed the comments, and updated the documentation on the mkdocs accordingly.

sheryjoe (approved):

jadie1:
@akenmorris Can we merge this?

akenmorris (approved):
Looks good.

Issue #794, opened 2020-10-22
Studio: Allow relative weighting up to 1000


akenmorris:
Currently limited to 0-1.

Issue #793, opened 2020-10-22
Studio: Ability to set to two viewers


akenmorris:
Currently the progression only goes in squares, 1,4,9,16,etc. Would like to be able to show 2.

Issue #770, opened 2020-10-10
Studio/Reconstruction - Relax same size constraint on distance transforms


akenmorris:
Right now you can run the whole pipeline in Studio without force image to be the same size, only the surface reconstruction is requiring this.  @sheryjoe says that this can be relaxed.

sheryjoe:
I think we can relax the same-size assumption for the reconstruction method. The core ingredient is resampling each of the distance transforms to a common coordinate space based on a TPS warp computed using correspondences. This only needs a fixed (common) size for the mean distance transform (not the individual samples/dts). So we could choose a common size (may be the largest possible size) and a voxel spacing and origin from the input samples.

PR #795, opened 2020-10-22
Studio updates

akenmorris:
This PR resolves:

#791 - Studio: reset camera on centering change and after grooming with centering
#793 - Studio: Ability to set to two viewers 
#794 - Studio: Allow relative weighting up to 1000
#770 - Studio/Reconstruction - Relax same size constraint on distance transforms

### DIFF
--- a/Libs/Analyze/Reconstruction.cpp
+++ b/Libs/Analyze/Reconstruction.cpp
@@ -151,6 +151,14 @@ void Reconstruction<TTransformType,TInterpolatorType, TCoordRep, PixelType, Imag
 }
 
 template < template < typename TCoordRep, unsigned > class TTransformType,
+  template < typename ImageType, typename TCoordRep > class TInterpolatorType,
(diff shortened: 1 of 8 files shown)


sheryjoe:
There is an offset/shift between groomed and original data. Any idea why is this happening? 
image
image

akenmorris:
Yeah, this isn't new, the local particles don't get transform with the grooming transform yet.  If you don't run centering, they will be more or less on the surface (other grooming steps will also affect this).  I've added it as an issue #797.

akenmorris:
More specifically, the reason the the particles and original surface are even close is that they are both being centered, but in different ways.  One is using the center of mass of the image and the other is using center of mass of the surface mesh.

akenmorris:
@sheryjoe , this should be fixed now in the latest commit.

akenmorris:
Shall we merge this and create a development build? @sheryjoe

sheryjoe (approved):

PR #796, opened 2020-10-22
Optimizer switch to TBB

akenmorris:
This PR changes out OpenMP for TBB.  The main driver for this is to enable multithreading on Mac, which doesn't support OpenMP (at least not without great difficulty).  TBB has a nearly equivalent parallel-for construct.  I would prefer to use C++17's parallel-for, but we are compiling as C++14 for now (due to limited C++17 support on some platforms).  TBB may also allow us more control over threading than OpenMP should we choose to further parallelize the code.

* Note: This does not add an additional dependency as we already had TBB as a sub-dependency of OpenVDB.

* Diff of point files shows this to be identical before/after.  I did not run *all* the use cases, but I don't there is no reason to think that there would be a difference.

* Speed on windows/linux seems comparable (or faster).

* Speed on Mac is greatly improved.  For example, the left atrium use case runs 4x faster on my mac laptop.

* I left the USE_OPENMP option since there are still OpenMP pragmas in the trimesh code.

### DIFF
--- a/Libs/Optimize/ParticleSystem/itkParticleGradientDescentPositionOptimizer.txx
+++ b/Libs/Optimize/ParticleSystem/itkParticleGradientDescentPositionOptimizer.txx
@@ -22,6 +22,10 @@ const int global_iteration = 1;
 #include "MemoryUsage.h"
 #include <chrono>
 
+#include <tbb/parallel_for.h>
+#include <tbb/task_scheduler_init.h>
+
+
 namespace itk
 {
   template <class TGradientNumericType, unsigned int VDimension>
@@ -64,6 +68,9 @@ namespace itk
   void ParticleGradientDescentPositionOptimizer<TGradientNumericType, VDimension>
     ::StartAdaptiveGaussSeidelOptimization()
   {
+    /// uncomment this to run single threaded
+    //tbb::task_scheduler_init init(1);
+
     if (this->m_AbortProcessing) {
       return;
     }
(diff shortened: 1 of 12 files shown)


Libs/Optimize/ParticleSystem/itkParticleGradientDescentPositionOptimizer.txx
```diff
@@ -109,24 +109,25 @@ namespace itk
...
           if (m_ParticleSystem->GetDomainFlag(dom) == true)
           {
-            continue;
+            return;
           }
 
           const ParticleDomain *domain = m_ParticleSystem->GetDomain(dom);
 
           typename GradientFunctionType::Pointer localGradientFunction = m_GradientFunction;
 #ifdef SW_USE_OPENMP
```
medakk, line 127: can this #ifdef be removed?

akenmorris: Yes indeed, and the others.

akenmorris:
ellipsoid_mesh is running quite a bit faster with TBB and I'm not sure why.  The results _look_ fine, but the numbers are a bit different.  However, I am discovering that with OpenMP enabled, the mesh code is not stable in the master branch.  Investigating...

akenmorris:
Ok, confirmed after dropping the random placement on the mesh, now results from single and multi-threaded are the same and reproducible.

This should be ready to merge.

medakk:
Tested on ellipsoid, the resulting shape models have identical compactness, generalization and specificity.

medakk (approved):

Issue #790, opened 2020-10-19
Clustering on meshes


jadie1:
- Need a python function that can cluster on meshes. This will use surface-to-surface distance and spectral clustering.
- Add option to use subsample on all use cases
- Add to use case documentation that new use cases should have subsample option
- Make sure --skip_grooming can be used with subsample option

If clustering on meshes is faster and more accurate, we may replace the way we cluster on segmentations with this technique on isosurfaces.

Issue #789, opened 2020-10-19
Ensure every use case has a tiny test


jadie1:
- Make sure all use cases have a tiny test
- Add to documentation that new use cases need a tiny test
- Make sure --skip_grooming can be used with --tiny_test

PR #808, opened 2020-11-08
Use case tiny test and subsampling

jadie1:
Updated use cases to:
- Include a tiny test which always runs on the first 3 shapes (issue #789)
- Include the option to run on a subset of the shapes based on clustering on segs or meshes (issue #790)
- Updated documentation on mkdocs branch to specify these options are necessary in new-use-case.md 

The only use case which does not have these options is the ellipsoid fixed domains because these options don't make sense in that context. 

To test tiny_test run:
`python RunUseCase.py --use_case <use_case> --tiny_test`

To test subsampling run:
`python RunUseCase.py --use_case <use_case> --use_subsample --num_subsample <integer>`
(Defaults to 3 if `num_subsample` is not specified)

### DIFF
--- a/Examples/Python/CommonUtils.py
+++ b/Examples/Python/CommonUtils.py
@@ -19,6 +19,8 @@ import shutil
 import xml.etree.ElementTree as ET
 from termcolor import colored, cprint
 from zipfile import ZipFile
+import subprocess
+import GroomUtils
 
 def download_and_unzip_dataset(datasetName, outputDirectory):
     # Check if the unzipped data is present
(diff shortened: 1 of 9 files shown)


akenmorris:
What do you think about combining the two options `--use_subsample` and `--num_subample <integer>`

Unless the default of 3 is somehow very useful, it seems the user would typically always specify both.

jadie1:
We just added this in PR #784 
@sheryjoe do you have thoughts on this?

sheryjoe:
--num_subample should be --num_subsample

jadie1:
It is --num_subsample in the code, sorry that was just a typo in the pull request. I'll fix it

sheryjoe:
> We just added this after some discussion in PR #784
> @sheryjoe do you have thoughts on this?

There is merit in both ways. Combining them into one argument (e.g. --use_subsample <optional num of samples, int>) is more compact, and having them as separate is more descriptive. I am fine with either.

PR #809, opened 2020-11-11
Libigl build included

riddhishb:
This is a very small PR which includes the download for libigl via `build_dependencies.sh` script.
Additionally, I included the include path to the header files to the libigl in the `CMakeLists.txt`.

To include any libigl function we just perform 
``` #include <igl/function.h>```

Here is a simple test which can be performed to check, execute this snippet of code after adding the appropriate changes to the CMake for adding a new script

```
#include <igl/cotmatrix.h>
#include <Eigen/Dense>
#include <Eigen/Sparse>
#include <iostream>
int main()
{
  Eigen::MatrixXd V(4,2);
  V<<0,0,
     1,0,
     1,1,
     0,1;
  Eigen::MatrixXi F(2,3);
  F<<0,1,2,
     0,2,3;
  Eigen::SparseMatrix<double> L;
  igl::cotmatrix(V,F,L);
  std::cout<<"Hello, mesh: "<<std::endl<<L*V<<std::endl;
  return 0;
}
```

I have tried this out and it works.
This PR will set the ball rolling for #684 and stuff that @medakk is working on with libigl (Do mention the issues).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -134,6 +134,9 @@ find_package (Eigen3 3.3 REQUIRED NO_MODULE)
 ### XLNT
 find_package (XLNT REQUIRED)
 
+### Libigl set include path
+include_directories("${CMAKE_INSTALL_PREFIX}/../build/libigl/include")
+
 if(USE_OPENMP)
   find_package(OpenMP REQUIRED)
   add_definitions(-DSW_USE_OPENMP) # used in ExternalLibs/trimesh2/include/TriMesh.h
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,7 +135,8 @@ find_package (Eigen3 3.3 REQUIRED NO_MODULE)
 find_package (XLNT REQUIRED)
 
 ### Libigl set include path
-include_directories("${CMAKE_INSTALL_PREFIX}/../build/libigl/include")
+include_directories("${CMAKE_PREFIX_PATH}/../build/libigl/include/")
+#MESSAGE("THIS ISNISANOSBCDOSCF ${CMAKE_PREFIX_PATH}")
 
 if(USE_OPENMP)
   find_package(OpenMP REQUIRED)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,7 +135,9 @@ find_package (Eigen3 3.3 REQUIRED NO_MODULE)
 find_package (XLNT REQUIRED)
 
 ### Libigl set include path
-include_directories("${CMAKE_PREFIX_PATH}/../build/libigl/include/")
+find_package(LIBIGL REQUIRED QUIET)
+
+#include_directories("${CMAKE_PREFIX_PATH}/install/libigl/include/")
 
 if(USE_OPENMP)
   find_package(OpenMP REQUIRED)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -137,8 +137,6 @@ find_package (XLNT REQUIRED)
 ### Libigl set include path
 find_package(LIBIGL REQUIRED QUIET)
 
-#include_directories("${CMAKE_PREFIX_PATH}/install/libigl/include/")
-
 if(USE_OPENMP)
   find_package(OpenMP REQUIRED)
   add_definitions(-DSW_USE_OPENMP) # used in ExternalLibs/trimesh2/include/TriMesh.h
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,7 +136,6 @@ find_package (XLNT REQUIRED)
 
 ### Libigl set include path
 include_directories("${CMAKE_PREFIX_PATH}/../build/libigl/include/")
-#MESSAGE("THIS ISNISANOSBCDOSCF ${CMAKE_PREFIX_PATH}")
 
 if(USE_OPENMP)
   find_package(OpenMP REQUIRED)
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by akenmorris** [comment]
I know that it's header only, but does this find_package work?

https://github.com/libigl/libigl-example-project/blob/master/CMakeLists.txt

https://github.com/libigl/libigl/issues/703

Or I guess we don't need it?

File: CMakeLists.txt [hunk 1]
```diff
@@ -134,6 +134,10 @@ find_package (Eigen3 3.3 REQUIRED NO_MODULE)
 ### XLNT
 find_package (XLNT REQUIRED)
 
+### Libigl set include path
+include_directories("${CMAKE_PREFIX_PATH}/../build/libigl/include/")
+#MESSAGE("THIS ISNISANOSBCDOSCF ${CMAKE_PREFIX_PATH}")
```
akenmorris, line 139: Let's remove this line.

riddhishb: Wont that cause the headers like ```<igl/cot.h>``` fail and we will have to use the whole path ?

Yeah I dont think we need the find package for this lib.

akenmorris: I just meant the commented out message line.

CMakeLists.txt
```diff
@@ -134,6 +134,10 @@ find_package (Eigen3 3.3 REQUIRED NO_MODULE)
 ### XLNT
 find_package (XLNT REQUIRED)
 
+### Libigl set include path
+include_directories("${CMAKE_PREFIX_PATH}/../build/libigl/include/")
```
akenmorris, line 138: This won't work if CMAKE_PREFIX_PATH isn't set to the dependencies install directory.  We aren't able to use this for both windows and linux on github actions for example.  We are using CMAKE_PREFIX_PATH=$CONDA_PREFIX on linux.  Something about the alternate gcc version.

On windows we have other issues.

CMakeLists.txt
riddhishb: oh ok

CMakeLists.txt
riddhishb: Oh I see, so how do we go about setting this include path, or do we have to set the path manually when we install with with libigl in CMakeLists.txt

riddhishb: What about having a LibIGL_INCLUDE_PATH which is set by default to "${CMAKE_PREFIX_PATH}/../build/libigl/include/" but can be changes via a flag ?

medakk: Maybe copy the libigl headers to ${INSTALL_DIR}/include as part of build_dependencies.sh, where all the other dependency headers are stored? Then the include path would always be ${CMAKE_PREFIX_PATH}/include

that seems to be what the [libigl cmake build does by default](https://github.com/libigl/libigl/blob/master/cmake/libigl.cmake#L506), what was the issue with just building libigl via cmake again?

akenmorris: There’s nothing to be built. We’re just doing header only mode.

medakk: I mean the "cmake build process", like what we do with Eigen. Eigen is header-only, but we build it anyway and that just copies the headers to the ${INSTALL_PREFIX}/include directory.

akenmorris: Yeah, that would be perfect. Does libgIGL have the same?

riddhishb: I tried making the same process for libigl but there are some issues 
for one it it only includes the .h files which cause the codes to complain about the associated .cpp files which are not copied to the install directory.
And of course the build process has it's trouble with managing their own eigen.

I arrieved at this simple solution only after checking all these things :(

CMakeLists.txt
```diff
@@ -135,7 +135,9 @@ find_package (Eigen3 3.3 REQUIRED NO_MODULE)
 find_package (XLNT REQUIRED)
 
 ### Libigl set include path
-include_directories("${CMAKE_PREFIX_PATH}/../build/libigl/include/")
+find_package(LIBIGL REQUIRED QUIET)
```
riddhishb, line 138: This does not configure, I get the following error

```
CMake Error at CMake/FindLIBIGL.cmake:34 (include):
  include could not find load file:

    libigl
Call Stack (most recent call first):
  CMakeLists.txt:138 (find_package)
```
I think the FindLIBIGL.cmake file is not written in a way that it can find the cloned libigl from a custom location.
What do you think @akenmorris @cchriste 

Thanks

riddhishb: Ok it works if I specify the include directory in the cmake, as follows: 
`/home/sci/riddhishb/Documents/git-repos/ShapeWorksUser/ShapeWorks/dependencies/install/libigl/cmake/`

cchriste: Do you mean something like `cmake ... -DLIBIGL_DIR=/path/to/install/libigl/cmake ...`?

akenmorris:
This only works on mac?

build_dependencies.sh
```diff
@@ -314,8 +326,8 @@ verify_qt()
 build_all()
 {
   ## create build and install directories
-  if [[ -z $BUILD_DIR ]];   then BUILD_DIR=${SRC}/dependencies/build;     fi
-  if [[ -z $INSTALL_DIR ]]; then INSTALL_DIR=${SRC}/dependencies/install; fi
+  if [[ -z $BUILD_DIR ]];   then BUILD_DIR=${SRC}/dependencies3/build;     fi
```
akenmorris, line 329: Should this be dependencies3 now?

riddhishb: Nope I will change that, that was on my local machine.

riddhishb:
I believe this can be merged now

Issue #768, opened 2020-10-10
Studio - Should generate error for inconsistent old XML files


akenmorris:
For example, it doesn't complain that the number of local and world particle files are different.  This then causes problems with later analysis.

akenmorris:
Fixed in #841, now show:

<img width="442" alt="Screen Shot 2020-12-11 at 12 30 05 PM" src="https://user-images.githubusercontent.com/1693349/101947354-92034180-3bad-11eb-8194-ee9d430a2226.png">

akenmorris:
Done

Issue #800, opened 2020-10-28
Studio: Add options for orientation marker widget


akenmorris:
Add options in preferences for orientation marker:

A) None
B) Triad
C) Medical cube

Also, which corner.

akenmorris:
Completed in #841 

<img width="539" alt="image" src="https://user-images.githubusercontent.com/1693349/102128446-1f8f9d00-3e0b-11eb-8320-2ebf1c7da00b.png">

Triad:

<img width="1199" alt="image" src="https://user-images.githubusercontent.com/1693349/102128525-3d5d0200-3e0b-11eb-9970-9610d0ee8173.png">

Upper left option:

<img width="1199" alt="image" src="https://user-images.githubusercontent.com/1693349/102128563-4d74e180-3e0b-11eb-98e2-fb38cdb1b933.png">

akenmorris:
Done

PR #841, opened 2020-12-01
Studio mesh support

akenmorris:
This PR adds basic mesh support to Studio.  The only mesh grooming step available is centering.

This PR also adds the `shapeworks groom` command that operates on a project spreadsheet. `shapeworks groom` does the equivalent of opening a project in Studio, running groom, and saving the project.

This PR also adds the spreadsheet project support to `shapeworks optimize`

Also fixed:

#768 - Studio - Should generate error for inconsistent old XML files
#800 - Studio: Add options for orientation marker widget 
#848 - Studio optimize default for initial relative weighting should be smaller than relative weighting
#856 - Studio should not let us set the PCA mode to greater than number of samples -2
#860 - Studio: Speed up particle shape statistics for large datasets
#859 - Studio needs an "abort groom" similar to "abort optimize"


medakk:
1. I loaded a single mesh(.vtk), skipped grooming, and ran the optimize with defaults. The particles did not show up until the entire optimization was done
2. I opened a single mesh(.ply) and pressed "skip" in the groom step. It crashed with:
```
Qt has caught an exception thrown from an event handler. Throwing
exceptions from an event handler is not supported in Qt.
You must not let any exception whatsoever propagate through Qt code.
If that is not possible, in Qt 5 you must at least reimplement
QCoreApplication::notify() and catch all exceptions there.

Exception caught!
std::exception
```

akenmorris:
@medakk , good catches.  They should all be fixed now.

jadie1:
I ran studio with some of the meshes (ply) from ellipsoid-v2 and the grooming and optimization worked. The only thing that didn't work was running reconstruction, is that supported yet?

Let me know if you have more suggestions of things to test.

akenmorris:
We don't yet have a surface reconstruction method when using meshes besides the VTK one.

medakk:
image

The remove button seems to always remove the last sample, and not the selected one

akenmorris:
> The remove button seems to always remove the last sample, and not the selected one

Yep.  There were actually a lot of problems with removing a subject.  It should be working now.

medakk:
@akenmorris could we have the normals strength be any real number instead of integers?
image

akenmorris:
> could we have the normals strength be any real number instead of integers?

Done!

Studio/src/Application/Optimize/OptimizeTool.ui
```diff
@@ -572,10 +541,10 @@ background: rgb(255, 255, 127)
          </size>
         </property>
         <property name="decimals">
-         <number>1</number>
+         <number>2</number>
         </property>
         <property name="minimum">
-         <double>0.100000000000000</double>
+         <double>0.010000000000000</double>
```
medakk, line 547: not too familiar with regularization but shouldn't we able to set this to 0, in case we want to turn off regularization?

Studio/src/Application/Optimize/OptimizeTool.ui
```diff
@@ -685,6 +654,34 @@ background: rgb(255, 255, 127)
...
+         <size>
+          <width>100</width>
+          <height>16777215</height>
+         </size>
+        </property>
+        <property name="maximum">
+         <double>999999999.000000000000000</double>
+        </property>
+        <property name="value">
+         <double>10.000000000000000</double>
```
medakk, line 681: this is the minimum for normals strength, right? I think we should be able to set this lower than 10.0. 

(I've used normal strength == 0.0 for some testing, but I don't know if this is useful for users)

medakk (approved):

Studio/src/Application/Optimize/OptimizeTool.ui
akenmorris: The minimum is zero.  10.0 is just the default value.

Studio/src/Application/Optimize/OptimizeTool.ui
akenmorris: Yep, seems like it works to set it to 0.

To disable, I believe you just set starting and ending to the same value.

PR #843, opened 2020-12-07
PR package naming

akenmorris:
This PR modifies the version naming for artifacts built from PR.  PRs will now have the form:

VERSION="ShapeWorks-PR-${PR_NUMBER}-${PLATFORM}"

They were previous using the tag mode, but this was typically just "dev-linux" (on any platform) and a not very useful git hash from the merge commit.

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - pr_package
     tags:
     - '*'
   pull_request:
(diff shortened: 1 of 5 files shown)


sheryjoe:
@akenmorris this is very useful, especially when sharing artifact links with the users. This means that a single name/link would be available for the PR-Platform artifact even if the PR kept updating.

sheryjoe (approved):

akenmorris:
Well, the artifact zip will still have a different hash, but the package inside will have a consistent name with the PR number.

Also, GitHub recently changed to require you to be logged in and I think on the team in order to access the artifacts.  So you can no longer pass links to users.

sheryjoe:
Ugh :( ... 🤦‍♀️ but this naming, at least among us, is useful. thanks!

PR #847, opened 2020-12-08
Fix broken conda installs on GitHub Actions (mac)

akenmorris:
The only solutions I can find for this problem are to either re-install (not applicable) or do this ln operation in the env lib dir:

`ln -s libffi.7.dylib libffi.6.dylib`

I have no idea why, but sometimes we end up with 7 and other times with 6 and pip always needs 6.

Relevant links:

https://github.com/conda/conda/issues/9038
https://github.com/conda/conda/issues/9899

9038 was closed for some reason.  9899 is still open.

### DIFF
--- a/conda_installs.sh
+++ b/conda_installs.sh
@@ -108,11 +108,12 @@ function install_conda() {
 
   if [[ "$GITHUB_ACTION" != "" ]]; then
       echo "Running under GitHub Action"
-      cd $HOME/miniconda3/envs/shapeworks/lib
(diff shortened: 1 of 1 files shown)


akenmorris:
This seems to solve the problem.  It's a hack, but it's the only suggestion I can find anywhere.

Any objections? @cchriste?  This changes nothing for non-GitHub Action builds.

medakk (approved):

PR #871, opened 2020-12-17
Fix python deployment

akenmorris:
This PR fixes the loader path for the shapeworks python module on Mac OS.

It also fixes conda_installs.sh for the Mac binary deployment.

The change to the `setting-up-shapeworks-environment.ipynb` should fix this script to run on the binary deployments of Mac, Linux and Windows without the user needing to change the path variables.  If this works, I suppose we need to change the rest of the notebooks, although I would suggest that this logic be moved to the setupenv.py script, but it will require changing the behavior, so I have not done so yet.

### DIFF
--- a/Examples/Python/notebooks/tutorials/getting-started-with-exploring-segmentations.ipynb
+++ b/Examples/Python/notebooks/tutorials/getting-started-with-exploring-segmentations.ipynb
@@ -69,9 +69,8 @@
     "\n",
     "- This notebook is located in `Examples/Python/notebooks/tutorials`\n",
     "- You have built shapeworks from source in `build` directory within the shapeworks code directory\n",
-    "- You have built shapeworks dependencies (using `build_dependencies.sh`) in the same parent directory of shapeworks code\n",
     "\n",
-    "**Note:** If you run from a ShapeWorks installation, you don't need to set the dependencies path and the `shapeworks_bin_dir` would be set as `../../../../bin`."
+    "**Note:** If you run from a ShapeWorks installation, you don't need to set the `shapeworks_bin_dir`"
    ]
   },
   {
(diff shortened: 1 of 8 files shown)


sheryjoe:
@akenmorris Examples/Python/setupenv.py should also be modified/updated.

akenmorris:
Current problem on mac:

```
ImportError: dlopen(/Applications/ShapeWorks/bin/shapeworks.cpython-37m-darwin.so, 2): Library not loaded: @executable_path/../Frameworks/libitkgdcmopenjp2-5.0.1.dylib
  Referenced from: /Applications/ShapeWorks/bin/ShapeWorksStudio.app/Contents/Frameworks/libitkgdcmMSFF-5.0.1.dylib
  Reason: image not found
```

iyerkrithika21 (approved):
I was able to run the Windows artifact and start the notebooks.
Faced no issues with starting jupyter notebooks.
Everything works fine.

akenmorris:
I've checked the mac and linux deployments as well.

iyerkrithika21:
@sheryjoe I found some errors with the `outDir` variable used in the notebooks to save the groomed files.
For example :
Notebook: getting-started-with-grooming-segmentations 
Cell : Saving resampled segmentations
Here the outDir defined is `../../Output/Notebooks/getting-started-groom-segmentations/ellipsoid-v2/resampled/`
But when the ellipsoids are being saved the file path and name is `./../Output/Notebooks/getting-started-groom-segmentations/ellipsoid-v2/resampled/segmentations\ellipsoid_00.nrrd`.
As the segmentation folder is not created, saving the nrrd files throws an error: ` No such file or directory`

sheryjoe:
@iyerkrithika21 good catch! This is due to the name split when we load the data. The cell that fills the shapeNames list. This split assume Mac and Linux. Please update to handle windows back slash too. I would use os.path.split. I thin it should be cross platform. Please update all notebooks accordingly.

iyerkrithika21:
> 
> 
> @iyerkrithika21 good catch! This is due to the name split when we load the data. The cell that fills the shapeNames list. This split assume Mac and Linux. Please update to handle windows back slash too. I would use os.path.split. I thin it should be cross platform. Please update all notebooks accordingly.

Python 3 has a pathlib, and using the Path() will convert forward slashes into the correct kind of slash for the current operating system.

iyerkrithika21:
@sheryjoe I am not able to run this notebook completely: `getting-started-with-grooming-segmentations`
The kernel always crashes after visualizing the output of  `Aligning all segmentations to the reference shape`

Issue #869, opened 2020-12-16
Use cases in the master branch need to be fixed


iyerkrithika21:
Tasks:
- [x] ellipsoid_cut - replace `applyRigidAlignment` with `applyResampling'
- [x] femur_cut - replace `applyRigidAlignment` with `applyResampling'
- [x] femur_cut - In the cropping step, region returned by `ImageUtils.boundingBox` is large and causes error
- [x] femur-In the cropping step, region returned by `ImageUtils.boundingBox` is large and causes error
- [x] GroomUtils - change the format of the clipping plane points used in the `ClipBinaryVolumes` function

sheryjoe:
@iyerkrithika21  can you add more details about what fails exactly in each use case?

iyerkrithika21:
Femur and femur cut errors : 
```
Invalid region specified.
Traceback (most recent call last):
  File "RunUseCase.py", line 79, in <module>
    module.Run_Pipeline(args)
  File "/home/sci/iyerkrithika/ShapeWorks/Examples/Python/femur.py", line 315, in Run_Pipeline
    croppedFiles_segmentations = applyCropping(groomDir + "cropped/segmentations", clippedFiles_segmentations, groomDir + "clipped_segmentations/*.nrrd")
  File "/home/sci/iyerkrithika/ShapeWorks/Examples/Python/GroomUtils.py", line 218, in applyCropping
    img.crop(region).write(outname)
RuntimeError: /home/sci/iyerkrithika/ShapeWorks/dependencies/install/include/ITK-5.0/itkImportImageContainer.hxx:199:
Failed to allocate memory for image.
```

jadie1:
The femur error when isotropic spacing is used is caused by center of mass alignment which is causing the segmentations to just be 0. This is fixed if the order of subtraction is switched, i.e. T = img.center() - img.centerOfMass() instead of T = img.centerOfMass() - img.center().

jadie1:
I have added fixes for:
- issue when non-isotropic spacing is used in the femur use case
- COM order issue which was effecting the bounding box step in femur and femur_cut
All of the fixes are on the use_case_fix branch. 
@iyerkrithika21 can you double check that all of the use cases are functioning now and then we can make a pull request?

Issue #826, opened 2020-11-25
Implement rigid alignment in use cases


jadie1:
The rigid alignment step in GroomUtils.py was doing a lot of unnecessary and unused preprocessing then simply changing the size and spacing to match the reference image and applying the identity matrix as the transform. So I removed it and made a step that just does the resampling.

We need a step which finds the transform between the reference and each segmentation and applies it.

PR #881, opened 2021-01-08
Use case fix

iyerkrithika21:
This is a small PR that fixes the use cases in the master branch.
Check #869 for more details.
To test this PR, run the use cases: femur,femur_cut with and without isotropic spacing.


jadie1:
@iyerkrithika21 and I have both tested all of the use cases.

akenmorris (approved):

akenmorris:
Looks good to me.  Can we merge, @sheryjoe ?

Examples/Python/GroomUtils.py
```diff
@@ -457,13 +457,14 @@ def MeshesToVolumesUsingImages(outDir, meshList, imgList, printCmd=True):
             print("CMD: " + " ".join(execCommand))
         subprocess.check_call(execCommand)
 
+        spacing_string = str(img.spacing()[0]).replace(".0","")
```
sheryjoe, line 460: We are taking here the x-dim spacing. What if the input image has an anisotropic spacing?

Examples/Python/ellipsoid_cut.py
```diff
@@ -84,11 +84,11 @@ def Run_Pipeline(args):
         """Apply center of mass alignment"""
         comFiles = applyCOMAlignment(groomDir + "com_aligned/segmentations", paddedFiles, None)
 
-        """Apply rigid alignment"""
-        rigidFiles = applyRigidAlignment(groomDir + "aligned/segmentations", comFiles, None, comFiles[0])
+        """ Apply resmapling"""
```
sheryjoe, line 87: resmapling -> resampling

Examples/Python/femur_cut.py
```diff
@@ -185,16 +185,16 @@ def Run_Pipeline(args):
             This function can handle both cases (processing only segmentation data or raw and segmentation data at the same time).
             This function uses the same transfrmation matrix for alignment of raw and segmentation files.
             """
-            [rigidFiles_segmentations, rigidFiles_images] = applyRigidAlignment(parentDir + "aligned", centerFiles_segmentations, centerFiles_images, medianFile, processRaw = True)
+            [resampledFiles_segmentations, resampledFiles_images] = applyResampling(parentDir + "resized", medianFile,centerFiles_segmentations, centerFiles_images)
```
sheryjoe, line 188: Why do we call the output of this step "resized"? This naming does not reflect the alignment step.

sheryjoe (changes requested):

Examples/Python/GroomUtils.py
jadie1: The `GenerateBinaryAndDTImagesFromMeshes` command currently automatically names the output image based on the x-spacing. It correctly rasterizes with any spacing you give it, but it only includes the x in the output name which we are renaming in the script anyway. I have already spoken with Archana about how when this command gets consolidated, we will allow the user to specify the output name to simplify things.

Examples/Python/femur_cut.py
jadie1: The use cases are not doing rigid alignment right now, see issue #826. For now we are calling it "resized" because all it is doing in resampling is resizing the image if needed. I will tackle this rigid alignment issue next.

sheryjoe (approved):

jadie1 (approved):

PR #890, opened 2021-01-13
Fixed domain mesh

### LINKED ISSUE #887: Fixed domain optimization doesn't work on meshes
When I run fixed domain optimization on meshes I get a segmentation fault in the intialization step.

To test:

- Unzip this: [fd_mesh_test.zip](https://github.com/SCIInstitute/ShapeWorks/files/5804750/fd_mesh_test.zip)
- Change the paths in `fd_mesh_test/correspondece_128.xml` to be correct for you
- Run `shapeworks optimize --name=fd_mesh_test/correspondence_128.xml`
- Note that if you remove the fixed domain tag in the xml it runs fine

akenmorris:
This should fix #887 

It looks like it used to work, but some new constraints work broke it.  I've added an automated test that will make sure this doesn't get broken again in the future.


jadie1 (approved):
Looks like it is working as expected, thanks Alan! Good thinking adding the test.

Issue #880, opened 2021-01-07
Studio: bug when adding and removing shapes


akenmorris:
Steps to reproduce:

1) Import a bunch of shapes
2) Remove all the shapes
3) Add back a few shapes

Result: all the shapes show back up.

Issue #870, opened 2020-12-16
Studio: Drag and Drop image volumes and meshes to import


akenmorris:
It would be nice to be able to drag and drop image files (nrrds) and/or meshes to Studio to import.

sheryjoe:
yes! similar to paraview.

sheryjoe:
@akenmorris resolved?

akenmorris:
Yes.

Issue #801, opened 2020-10-28
Studio: Add tooltips


akenmorris:
Add tooltips to groom/optimize parameters.

Issue #792, opened 2020-10-21
Studio:  Add Help->Keyboard Shortcuts


akenmorris:

akenmorris:
This was completed.

PR #891, opened 2021-01-13
Studio Updates

akenmorris:
#892 - Studio should append ".xlsx"
#801 - Studio: Add tooltips
#792 - Studio: Add Help->Keyboard Shortcuts
* All supported mesh type as well as nrrd/mhd are supported.

#880 - Studio: bug when adding and removing shapes
#674 - Visualize DeepSSM error on the predicted mesh in Studio

@jadie1, you can run:

ShapeWorksStudio mesh.vtk

And it will create an empty project with this one file.  Rather than auto detect a scalar array called "Distance", I have modified the SurfaceToSurfaceDistance executable to set Distance as the "active scalar" (a vtk attribute).  Studio will then check if an active scalar is present and auto select this on opening.  So you may need to re-generate any VTK files that you have or they won't auto select to Distance.  You can, of course, just select "Distance".

Screen Shot 2021-01-12 at 7 07 35 PM

@sheryjoe, this capability, incidentally also extends the "feature map" functionality to meshes by default.  Meshes can store values at vertices such as "cortical thickness", or "fibrosis" and Studio will be able to use them like the feature volumes.


Studio/src/Application/Visualization/ShapeWorksStudioApp.cpp
```diff
@@ -342,6 +342,11 @@ bool ShapeWorksStudioApp::on_action_save_project_as_triggered()
   if (filename.isEmpty()) {
     return false;
   }
+
+  if (!StringUtils::hasSuffix(filename.toStdString(), "xlsx")) {
```
medakk, line 346: should be ".xlsx" (missing dot)

akenmorris: Yes indeed.  That goes for my other usage of hasSuffix.

jadie1:
Do I need to rerun conda_installs or build_dependencies to test this? When I run make on this branch I am getting this error:
image

akenmorris:
Just re-run cmake.

Studio is using globs for the files, so your build probably didn't include them.  I still need to get rid of the CMake globs.

jadie1:
Got it thanks!

jadie1 (approved):
I just tested the DeepSSM mesh visualization and that looks great! The heatmap display looked very similar to how paraview displays it.

akenmorris:
@sheryjoe , I've added a considerable about of Studio documentation to the mkdocs markdown files in this branch.  Have a look and see what you think.

docs/studio/getting-started-with-studio.md
```diff
@@ -0,0 +1,200 @@
...
+
+Studio's file menu is shown below.
+
+![ShapeWorks Studio File Menu](../img/studio/studio_file_menu.png)
+
+* New Project - Create a new project
+* Open Project - Open a project from disk
+* Save Project - Save project to disk
+* Save Project As - Save a copy of the project to a new file
+* Import Images... - Import new shapes
```
sheryjoe, line 153: Does `Import Image` import segmentations or meshes? I think we should be more explicit as we will also import imaging data with shapes. If it can import both, let's use "Image Shapes ..."

docs/studio/getting-started-with-studio.md
```diff
@@ -0,0 +1,200 @@
...
+![ShapeWorks Studio Particle Selection2](../img/studio/studio_particle_selection2.png)
+
+
+## File menus
+
+Studio's file menu is shown below.
+
+![ShapeWorks Studio File Menu](../img/studio/studio_file_menu.png)
+
+* New Project - Create a new project
```
sheryjoe, line 149: How about using this convention to make it more visible:  `New Project` - Create a new project? We use this convention when we document parameters.

sheryjoe (changes requested):
Nice documentation. Added some comments and will commit minor edits from my side. Thanks @akenmorris!

docs/getting-started/interfaces.md
```diff
@@ -28,14 +28,16 @@ We are developing Python bindings to enable users to script their customized sha
     Exemplar [use cases](../use-cases/use-cases.md) will be updated to use these Python APIs to showcase their use and flexibility.
 
 
-## ShapeWorksStudio
+## ShapeWorks Studio
```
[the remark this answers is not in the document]
akenmorris: It's my understanding that we're going with "ShapeWorks Studio".  See the splash screen for example.  I don't mind either way.  If you want to drop the space that's ok with me.  Just let me know.

docs/studio/getting-started-with-studio.md
```diff
@@ -0,0 +1,200 @@
...
+
+![ShapeWorks Studio Data Module](../img/studio/studio_data.png)
+
+## Groom Module
+
+The Groom Module provides options to preprocess/groom the input data.  Options differ for binary segmentations and meshes.
+
+You also have the option of skipping grooming if your data is already prepped.
+
+![ShapeWorks Studio Groom Module](../img/studio/studio_groom.png)
```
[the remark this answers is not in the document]
akenmorris: What do you mean?  studio_groom.png is the same as studio_data.png?  They seem different and correct to me.

docs/studio/getting-started-with-studio.md
akenmorris: Yeah, I agree.

docs/studio/getting-started-with-studio.md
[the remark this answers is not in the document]
sheryjoe: Please check out the images here. You can see in the groom image, it shows isosurfaces of segmentations. Same for the optimize one.
https://github.com/SCIInstitute/ShapeWorks/blob/studio/docs/studio/getting-started-with-studio.md

sheryjoe:
@akenmorris Is this ready to merge?

PR #894, opened 2021-01-13
Data subset

iyerkrithika21:
Thi PR ensures that running use cases with tiny test argument only downloads a subset of the data from the portal.
But for sub_sample, the whole dataset will be downloaded to pick representative samples.

To test run the following use cases with and without tiny_test:

- [ ]     Ellipsoid:
- [ ]     Ellipsoid cut: 
- [ ]     Ellipsoid mesh:
- [ ]     Femur:
- [ ]     Femur cut: 
- [ ]     Femur mesh: 
- [ ]     Left atrium: 
- [ ]     Lumps:


jadie1 (approved):
This looks good to me, I ran every use case and confirmed it only downloads the tiny test data when the tiny test data tag is used otherwise it downloads the entire zip.

I added this same functionality for the deep_ssm use case. @iyerkrithika21 double check that one when you have a chance.

We should update the data downloading explanation in use_cases.md on the mkdocs branch for this.  

Otherwise I think it is good to merge!

iyerkrithika21:
> 
> 
> This looks good to me, I ran every use case and confirmed it only downloads the tiny test data when the tiny test data tag is used otherwise it downloads the entire zip.
> 
> I added this same functionality for the deep_ssm use case. @iyerkrithika21 double check that one when you have a chance.
> 
> We should update the data downloading explanation in use_cases.md on the mkdocs branch for this.
> 
> Otherwise I think it is good to merge!

@jadie1 I ran the deep_ssm use case and it ran fne, only downloads 7 examples for tiny test and whole dataset otherwise

sheryjoe:
@akenmorris could you please have a look at this one?

akenmorris (approved):
Looks great.

PR #896, opened 2021-01-13
Gradient of normals for image domain

### LINKED ISSUE #803: Implement gradient of normals for image domain
Use [ITK's Hessian filter](https://itk.org/Doxygen/html/classitk_1_1HessianRecursiveGaussianImageFilter.html) to get hessians instead of implementing it via multiple DerivativeImageFilters in ParticleImageDomainWithHessians.h

(also look into adding a custom type to openvdb to store hessians directly, then we'd have to sample only once to get hessians instead of 6 times https://github.com/AcademySoftwareFoundation/openvdb/issues/696 )

medakk:
This PR:

* Gets rid of Hessians everywhere and replaces it with gradient of normal
* Gets rid of existing curvature computation code(that required hessian) and replaces it with openvdb::meanCurvature. The previous version used a gaussian filter, I have not observed any difference after removing this. @sheryjoe [see diff](https://github.com/SCIInstitute/ShapeWorks/compare/new_image_gradnorm?expand=1#diff-1d2fbe606b973388b9b37b9343e977256adf36ab11d8b6f8c91aea9c4bce5558)
* There is an increase in memory usage for image domain(femur went from 1018MB to 1171MB, a 15% increase)

**On a sphere with use_xyz=0, use_normals=1:**

Master branch:

20210113_sphere_master

This branch:

20210113_sphere_imagegradnorm

**On the ellipsoid use case with use_xyz=1, use_normals=1:**

Master branch:

20210113_ellipsoid_master

This branch (hyperparameters had to be adjusted):

20210113_ellipsoid_imagegradnorm

**Running the femur use case** on this branch, I see no regressions:

20210113_femur_newimagegradnorm

Closes #803


sheryjoe:
@akenmorris did you have a chance to review this PR?

akenmorris (approved):
Looks good to me.  Do we need to universally adjust the attribute scales after this change?  I notice you changed from 10->15 in some cases and 10->1 in others.

medakk:
>  Do we need to universally adjust the attribute scales after this change?

I believe so.

>  I notice you changed from 10->15 in some cases and 10->1 in others.

I've had to fiddle with the relative weighting too in order to get correct correspondences after this change(+regularization for the sphere unit test). I'm unsure if there is a concrete relationship between the changes required in hyperparameters for using hessians vs gradient of normals

akenmorris:
> > Do we need to universally adjust the attribute scales after this change?
> I believe so.

I don't see any changes to femur and left atrium that are using normals.  Do these need to be checked and adjusted?

medakk:
[femur](https://github.com/SCIInstitute/ShapeWorks/blob/67a9cac5c21e70f01bf9c8d42c2b0085063d0e82/Examples/Python/femur.py#L342) and [left atrium](https://github.com/SCIInstitute/ShapeWorks/blob/67a9cac5c21e70f01bf9c8d42c2b0085063d0e82/Examples/Python/left_atrium.py#L204) are not using normals

akenmorris:
> femur and left atrium are not using normals

Indeed they are not!

I think LA used to be (I recall suggested that it is not necessary).

Issue #845, opened 2020-12-07
Retire ShapeWorks Post after adding standalone tool for mesh warping


riddhishb:
Post is deprecated after we remove the mesh warping functionality.

akenmorris:
I assume we should wait to retire it until there is a new command for `shapeworks`?  @sheryjoe @riddhishb

sheryjoe:
Agreed.

sheryjoe:
@riddhishb is this still an issue?

PR #898, opened 2021-01-14
Mesh Warping Function

riddhishb:
This PR addresses the following aspects

1. Standalone tool, ``` WarpMeshUsingLandmarks.cxx```, it takes three arguments, template mesh file, control points on template mesh, XML file specifying the paths to moving points, and output paths.
2. The same tool integrated into MeshUtils.cpp, with functions ```generateWarpMatrix```, ```warpMesh``` (warps one mesh), ```warpMeshes``` (warp multiple meshes).
3. Basis unit test to check the accuracy of the warp matrix generation.


riddhishb:
A couple of things.
1. Do we remove the standalone function as the code is repeated.
2. Once people have tested this, I will also remove the shapeworks post stuff to address  #845.

cchriste:
Just to clarify, this executable's functionality could be distilled to:
1. precompute X using all the meshes and their associated point files (a loop that produces one thing)
2. use (const) X for each mesh, producing a set of new meshes (a loop that produces N things)

@riddhishb If this is incorrect (highly possible), will you please clarify? Thanks!

akenmorris (changes requested):
I think we should not add MeshUtils::pointReadFormat.

I believe you can use the constructor of ParticleSystem::ParticleSystem to read a particle file into an Eigen Matrix.  Will this do what you need?

Trying to limit the number of particle readers in SW we have to maintain

riddhishb:
> I think we should not add MeshUtils::pointReadFormat.
> 
> I believe you can use the constructor of ParticleSystem::ParticleSystem to read a particle file into an Eigen Matrix. Will this do what you need?
> 
> Trying to limit the number of particle readers in SW we have to maintain

riddhishb:
@akenmorris I have added the changes and also an additional test which is basically the direct use of the tool. The tests is [here](https://github.com/SCIInstitute/ShapeWorks/blob/f27206d2a7214e788d4809c450c8102bb9a621c8/Testing/MeshTests/MeshTests.cpp#L49)

akenmorris (approved):

akenmorris:
@archanasri , @cchriste 

Before I merge this PR, can you have a look at the changes to Mesh and MeshUtils?  I assume there will likely be conflicts with the `mesh_party` branch.

Libs/Mesh/CMakeLists.txt
```diff
@@ -161,3 +161,7 @@ INSTALL(TARGETS SurfaceToSurfaceDistance RUNTIME DESTINATION bin)
 #ADD_EXECUTABLE(fea2vtk fea2vtk.cxx)
 #TARGET_LINK_LIBRARIES(fea2vtk ${ITK_LIBRARIES} Mesh tinyxml)
 #INSTALL(TARGETS fea2vtk RUNTIME DESTINATION bin)
+
+ADD_EXECUTABLE(WarpMeshUsingLandmarks WarpMeshUsingLandmarks.cxx )
+TARGET_LINK_LIBRARIES(WarpMeshUsingLandmarks ${ITK_LIBRARIES} Mesh tinyxml Eigen3::Eigen igl::core)
+INSTALL(TARGETS WarpMeshUsingLandmarks RUNTIME DESTINATION bin)
```
archanasri, line 167: We can remove the standalone executable if the consolidated version exists.

Libs/Mesh/MeshUtils.cpp
```diff
@@ -29,4 +42,125 @@ const vtkSmartPointer<vtkMatrix4x4> MeshUtils::createIcpTransform(const vtkSmart
   return m;
 }
 
+Eigen::MatrixXd MeshUtils::distilVertexInfo(vtkSmartPointer<vtkPolyData> mesh){
```
archanasri, line 45: You can just pass `Mesh mesh` instead of `vtkSmartPointer<vtkPolyData>`. `vtkSmartPointer<vtkPolyData>` is defiend as MeshType in Mesh.h as private.

akenmorris:
Ok to merge? @cchriste @sheryjoe

Libs/Mesh/MeshUtils.cpp
sheryjoe: Agreed

Testing/MeshTests/MeshTests.cpp
```diff
@@ -14,3 +18,53 @@ TEST(MeshTests, coverage_test) {
...
+	igl::point_mesh_squared_distance(staticPts.transpose(),Vref2,J,sqrD,b,_2);
+  bool chkdst = (!(sqrD.sum() > 1.0e-10)) && (!(sqrD.maxCoeff() > 1.0e-10));
+  ASSERT_TRUE(chkdst);
+}
+
+TEST(MeshTests, warpTest2)
+{
+  Mesh ellipsoid( std::string(TEST_DATA_DIR) + "/ellipsoid_0.ply");
+  Mesh ellipsoid_warped( std::string(TEST_DATA_DIR) + "/ellipsoid_warped.ply");
+  Mesh::MeshType inMesh = ellipsoid.get_poly_data();
```
cchriste, line 53: One example of many in this pull request. Please use Mesh outside of a function, and if it's needed the functions themselves can call Mesh::getVTKMesh to get it.

Libs/Mesh/CMakeLists.txt
akenmorris: A consolidated version does not exist.  @riddhishb @sheryjoe @cchriste , what is the plan here?  Are we adding an executable and/or command for this?

Libs/Mesh/Mesh.h
```diff
@@ -14,8 +15,10 @@ class Mesh
   using MeshType = vtkSmartPointer<vtkPolyData>;
 
   Mesh(const std::string& pathname) : mesh(read(pathname)) {}
-  Mesh(vtkSmartPointer<vtkPolyData>&& rhs) : mesh(std::move(rhs)) {}
-
+  // Mesh(vtkSmartPointer<vtkPolyData>&& rhs) : mesh(std::move(rhs)) {}
```
cchriste, line 18: These are significant modifications that need more consideration. For example, copying a smart pointer means the previous owner can keep modifying the object, resulting in difficult bugs, particularly when the class is used by other apps, python apis, etc. Thus we intentionally provided only a move constructor.

Libs/Mesh/Mesh.h
```diff
@@ -60,7 +72,8 @@ class Mesh
   MeshType get_poly_data();
 
 private:
-  Mesh() {}
+  // Mesh() {}
+  // Mesh() : mesh(nullptr) {} // only for use by SharedCommandData since a Mesh should always be valid, never "empty"
```
cchriste, line 76: We need these functions. Why have they been commented out?

cchriste (changes requested):
A few things aren't acceptable in this PR.

1. We want ShapeWorks to be generally usable, rather than requiring everyone to know everything about every underlying library we use. That's why we want to pass around our own Image and Mesh classes rather than the underlying classes used inside their implementation. 

2. No more executables, please. We talked about this during a meeting, but apparently this wasn't made clear. We can figure out how to integrate the functionality (cut and pasted in this pull request by the way) so it can be used in general with the Python API, command line interface, and C++. As an example, we are consolidating the ComputeDistances executable, which produces two new meshes, into a function in Mesh like this: `Mesh& computeDistance(Mesh& other, …);` which modifies both meshes by adding a field called distance representing the distance to the other mesh (the name of this field will likely be a parameter so one could call the function many times for a single mesh and its peers).

3. Missing additions to the Python API, standalone shapeworks commands, and tests for both.

4. Unit tests are too complex. Each test should simply perform the operation of a single function. Please see the tests in the mesh_party branch for more examples, or look at the existing tests in ImageTests.cpp.

cchriste:
Some of the functionality in this branch has already been added to mesh_party. We should merge from here to mesh_party to master. That way all this more-than-related functionality gets tested together before entering master. Also, I understand some of the Studio functionality requires this. I think we're close as mesh_party is increasingly robust.

cchriste:
@riddhishb @akenmorris Will the remaining changes in this pull request still be needed? From what I can see all the functionality of the standalone executable now exists and is accessible via command line, python api, and directly from c++, and, while still overcomplicated, unit tests. 

If it's not needed let's close this.

akenmorris:
I don't think this functionality is accessible anywhere outside of Studio yet.

I'm not aware of a `shapeworks` command for mesh warping, nor a pythong interface.

cchriste:
Okay, I had inferred from the distilled version of this pull request that it included all the required functionality. Can we create some explicit GitHub issues for those items and close this pull request for now?

akenmorris:
There is a bug in either this code or in LibIGL with the following inputs (attached):

[warp_crash.zip](https://github.com/SCIInstitute/ShapeWorks/files/5952326/warp_crash.zip)

`WarpMeshUsingLandmarks thin_ellipsoid_14_groomed.ply thin_ellipsoid_14.local.particles out.xml 128`

The standalone does not typically crash, but Studio using it does.  If you run the standalone under a debugger, it will halt at an assert:

```
EIGEN_DEVICE_FUNC
    EIGEN_STRONG_INLINE Scalar&
    operator()(Index index)
    {
      eigen_assert(index >= 0 && index < size());
      return coeffRef(index);
    }
```

The IGL code is attempting to access a value past the size of the matrix:

```
__pthread_kill 0x00007fff6880333a
pthread_kill 0x00007fff688bfe60
abort 0x00007fff6878a808
__assert_rtn 0x00007fff68789ac6
Eigen::DenseCoeffsBase<Eigen::Matrix<int, -1, 1, 0, -1, 1>, 1>::operator()(long) DenseCoeffsBase.h:425
bool igl::min_quad_with_fixed_precompute<double, Eigen::Matrix<int, -1, 1, 0, -1, 1> >(Eigen::SparseMatrix<double, 0, int> const&, Eigen::MatrixBase<Eigen::Matrix<int, -1, 1, 0, -1, 1> > const&, Eigen::SparseMatrix<double, 0, int> const&, bool, igl::min_quad_with_fixed_data<double>&) min_quad_with_fixed.cpp:83
bool igl::min_quad_with_fixed<double, Eigen::Matrix<int, -1, 1, 0, -1, 1>, Eigen::Matrix<double, -1, 1, 0, -1, 1>, Eigen::Matrix<double, -1, -1, 0, -1, -1>, Eigen::Matrix<double, -1, 1, 0, -1, 1>, Eigen::Matrix<double, -1, -1, 0, -1, -1> >(Eigen::SparseMatrix<double, 0, int> const&, Eigen::MatrixBase<Eigen::Matrix<double, -1, 1, 0, -1, 1> > const&, Eigen::MatrixBase<Eigen::Matrix<int, -1, 1, 0, -1, 1> > const&, Eigen::MatrixBase<Eigen::Matrix<double, -1, -1, 0, -1, -1> > const&, Eigen::SparseMatrix<double, 0, int> const&, Eigen::MatrixBase<Eigen::Matrix<double, -1, 1, 0, -1, 1> > const&, bool, Eigen::PlainObjectBase<Eigen::Matrix<double, -1, -1, 0, -1, -1> >&) min_quad_with_fixed.cpp:578
bool igl::biharmonic_coordinates<Eigen::Matrix<double, -1, -1, 0, -1, -1>, Eigen::Matrix<int, -1, -1, 0, -1, -1>, int, Eigen::Matrix<double, -1, -1, 0, -1, -1> >(Eigen::MatrixBase<Eigen::Matrix<double, -1, -1, 0, -1, -1> > const&, Eigen::MatrixBase<Eigen::Matrix<int, -1, -1, 0, -1, -1> > const&, std::__1::vector<std::__1::vector<int, std::__1::allocator<int> >, std::__1::allocator<std::__1::vector<int, std::__1::allocator<int> > > > const&, int, Eigen::PlainObjectBase<Eigen::Matrix<double, -1, -1, 0, -1, -1> >&) biharmonic_coordinates.cpp:200
W_precomputation(Eigen::Matrix<double, -1, -1, 0, -1, -1>, Eigen::Matrix<double, -1, -1, 0, -1, -1>, Eigen::Matrix<int, -1, -1, 0, -1, -1>, Eigen::Matrix<int, -1, -1, 0, -1, -1>) WarpMeshUsingLandmarks.cxx:56
main WarpMeshUsingLandmarks.cxx:184
start 0x00007fff686bbcc9
```

riddhishb:
@akenmorris I do know know this error! I will take a look at this.

akenmorris:
@riddhishb , for further info, this was not a good run of the optimizer (points may be bad).  If the data is somehow invalid, that's fine, we just need a check for it and we can run the legacy reconstructor.  I just can't have it crash.

riddhishb:
@akenmorris ok I see, do we need to put an exception somewhere in order to fix this?

riddhishb:
I think I have figured out the issue, the algorithm is interpolating the value of particle to the closest mesh vertex first and then using TPS to interpolate to other vertices, the for loop contining "unknown vertices" do not have a added contion as to what happens when two different control points gets projected to the same nearest vertex. 
This as far as I can think can happen in two senarios, 

1. one where two control points are very close together (dunno how to decide threshold maybe some percentage of mean inter point distance).
2. When the mesh is sparse and number of points is decently close to number of template mesh vertices (this happens in the data you sent Alan), an easy fix to this is just subdivide the template mesh and increase the number of vertices and that fixed the problem.

There is a way to produce a flag from inside the IGL which checks fro this phenomena, but I do not know how to change teh libigl everytime after build depndency clones a libigl.

akenmorris:
@riddhishb , should we do this projection step on our end and if two particles land on the same vertex, we just arbitrarily remove one?  I'm not sure how subdividing the mesh can guarantee that it won't still happen again.

PR #905, opened 2021-01-17
Mesh party

cchriste:
There are almost 20 new shapeworks commands and their corresponding Mesh, Image, etc functions.
Many of these resolve issues for our next release, and more testing is advised. Tasks include:

- [x] ensuring each new function has a corresponding test in MeshTests/ImageTests/PythonTests/shapeworksTests
- [x] `ImageUtils::toVtk` returns a bare pointer, and probably this should be a smart pointer (`std::shared_ptr<...>`, VtkSmartPtr, etc)
- [x] `SurfaceToSurfaceDistance.cxx` has modifications that need further consideration
- [x] `ImageTests.cpp` has `toMeshTest1` and `toMeshTest2` that need to be finished
- [x] questionable `compare_double` function found its way out of the mesh points compare code into Shapeworks.h

In summary, there are a couple of items that need completion or polishing, and lots of testing would be prudent.
Please add tests as you conduct them. All are welcome to help.


sheryjoe:
@medakk @jadie1 where are we on testing this PR?

jadie1:
@sheryjoe @archanasri @cchriste Could we get a list of the functions that need to be tested and maybe @medakk and I could split them? This PR is large and I don't know how to approach testing

jadie1:
The femur use case breaks on the mesh-to-image step. When I call the command it is breaking on it gives me a segmentation fault, if call the same command but don't write the image then it doesn't give me an error.

image

archanasri:
> The femur use case breaks on the mesh-to-image step. When I call the command it is breaking on it gives me a segmentation fault, if call the same command but don't write the image then it doesn't give me an error.
> 
> image

@jadie1 Are you getting the same error when you run tiny_test?

jadie1:
@archanasri Yes I am

cchriste:
> > The femur use case breaks on the mesh-to-image step. When I call the command it is breaking on it gives me a segmentation fault, if call the same command but don't write the image then it doesn't give me an error.

Hi @jadie1, this issue is now resolved. I used the same command before and after and it crashed like you described and succeeded after fixing the Image constructor from vtkImageData.

cchriste:
This branch is up to date with alpha which is up to date with master. The pull request from alpha to master will be the final one, but changing the root branch here doesn't do anything since they're already up to date. Closing this pull request in favor of using alpha. (Our working branch is meshtest, but users should test from alpha)

PR #906, opened 2021-01-18
Rigid Alignment

jadie1:
This PR adds rigid alignment to the grooming pipeline in segmentation based use cases - issue #826 

This PR also changes the ellipsoid case to use ellipsoid-v2 data which is not already groomed to provide a better demonstration of grooming. It changes ellipsoid_cut to use ellipsoid-aligned-v0 which is already aligned. The grooming steps except creating distance transforms are removed so that the use case highlights the cutting plane and does not repeat the ellipsoid use case. 

To test run the following use cases:
- ellispoid
- ellipsoid_cut
- femur (with and without grooming images)
- femur_cut (with and without grooming images)
- left_atrium (with and without grooming images)

### DIFF
--- a/Examples/Python/GroomUtils.py
+++ b/Examples/Python/GroomUtils.py
@@ -103,7 +103,7 @@ def applyCOMAlignment(outDir, inDataListSeg, inDataListImg, processRaw=False):
         T = img.center() - img.centerOfMass()
 
         # binarize result since linear interpolation makes image blurry again
-        img.translate(T).binarize().write(outname)
+        img.translate(T).binarize().recenter().write(outname)
 
         if processRaw:
             innameImg = inDataListImg[i]
@@ -155,12 +155,16 @@ def FindReferenceImage(inDataList):
     print(" ")
     return inDataList[idx]
 
-def applyResampling(outDir, refFile, inDataListSeg, inDataListImg=[]):
+def applyRigidAlignment(outDir, refFile, inDataListSeg, inDataListImg=[]):
     """
     This function takes in a filelists(binary and raw) and makes the 
     size and spacing the same as the reference
     """
-    print("\n############# Resample #############")
+    isoValue       = 1e-20
+    icp_iterations = 200
+    antialias_iterations = 1 
+
+    print("\n############# Rigidly Align #############")
 
     # create output dirs
     segoutDir = os.path.join(outDir, 'segmentations') if inDataListImg else outDir
@@ -177,21 +181,23 @@ def applyResampling(outDir, refFile, inDataListSeg, inDataListImg=[]):
 
     # get reference image
     refImg = Image(refFile)
+    refImg.antialias(antialias_iterations)
 
     for i in range(len(inDataListSeg)):
-        segoutname = rename(inDataListSeg[i], segoutDir, 'resized')
+        segoutname = rename(inDataListSeg[i], segoutDir, 'aligned')
         outSegDataList.append(segoutname)
         if inDataListImg:
-            rawoutname = rename(inDataListImg[i], rawoutDir, 'resized')
+            rawoutname = rename(inDataListImg[i], rawoutDir, 'aligned')
             outRawDataList.append(rawoutname)
 
         # resize images to reference images
         img = Image(inDataListSeg[i])
-        transform = createTransform(Matrix()) # create identity matrix
-        img.resample(transform, refImg.origin(), refImg.dims(), refImg.spacing(), refImg.coordsys(), InterpolationType.NearestNeighbor).write(segoutname)
+        rigidTransform = ImageUtils.createRigidRegistrationTransform(Image(inDataListSeg[i]).antialias(antialias_iterations), refImg, isoValue, icp_iterations)
+        img.resample(rigidTransform, refImg.origin(), refImg.dims(), refImg.spacing(), refImg.coordsys(), InterpolationType.Linear).write(segoutname)
+
         if inDataListImg:
             img = Image(inDataListImg[i])
-            img.resample(transform, refImg.origin(), refImg.dims(), refImg.spacing(), refImg.coordsys(), InterpolationType.NearestNeighbor).write(rawoutname)
+            img.resample(rigidTransform, refImg.origin(), refImg.dims(), refImg.spacing(), refImg.coordsys(), InterpolationType.Linear).write(rawoutname)
 
     return [outSegDataList, outRawDataList] if inDataListImg else outSegDataList
 
--- a/Examples/Python/GroomUtils.py
+++ b/Examples/Python/GroomUtils.py
@@ -193,11 +193,11 @@ def applyRigidAlignment(outDir, refFile, inDataListSeg, inDataListImg=[]):
         # resize images to reference images
         img = Image(inDataListSeg[i])
         rigidTransform = ImageUtils.createRigidRegistrationTransform(Image(inDataListSeg[i]).antialias(antialias_iterations), refImg, isoValue, icp_iterations)
-        img.resample(rigidTransform, refImg.origin(), refImg.dims(), refImg.spacing(), refImg.coordsys(), InterpolationType.Linear).write(segoutname)
+        img.applyTransform(rigidTransform, refImg.origin(), refImg.dims(), refImg.spacing(), refImg.coordsys(), InterpolationType.Linear).write(segoutname)
 
         if inDataListImg:
             img = Image(inDataListImg[i])
-            img.resample(rigidTransform, refImg.origin(), refImg.dims(), refImg.spacing(), refImg.coordsys(), InterpolationType.Linear).write(rawoutname)
+            img.applyTransform(rigidTransform, refImg.origin(), refImg.dims(), refImg.spacing(), refImg.coordsys(), InterpolationType.Linear).write(rawoutname)
 
     return [outSegDataList, outRawDataList] if inDataListImg else outSegDataList
 
--- a/Examples/Python/GroomUtils.py
+++ b/Examples/Python/GroomUtils.py
@@ -85,6 +85,7 @@ def applyCOMAlignment(outDir, inDataListSeg, inDataListImg, processRaw=False):
     files in the appropriate directory.
     """
     print("\n############# COM Alignment ###############")
+    antialias_iterations = 30
     segDir = os.path.join(outDir, 'segmentations') if processRaw else outDir
     if not os.path.exists(segDir):
         os.makedirs(segDir)
@@ -103,14 +104,14 @@ def applyCOMAlignment(outDir, inDataListSeg, inDataListImg, processRaw=False):
         T = img.center() - img.centerOfMass()
 
         # binarize result since linear interpolation makes image blurry again
-        img.translate(T).binarize().recenter().write(outname)
+        img.antialias(antialias_iterations).translate(T).binarize().recenter().write(outname)
 
         if processRaw:
             innameImg = inDataListImg[i]
             outnameImg = rename(innameImg, imageDir, 'com')
             outDataListImg.append(outnameImg)
             rawImg = Image(innameImg)
-            rawImg.translate(T).write(outnameImg)
+            rawImg.translate(T).recenter().write(outnameImg)
 
     return [outDataListSeg, outDataListImg] if processRaw else outDataListSeg
 
--- a/Examples/Python/GroomUtils.py
+++ b/Examples/Python/GroomUtils.py
@@ -156,14 +156,13 @@ def FindReferenceImage(inDataList):
     print(" ")
     return inDataList[idx]
 
-def applyRigidAlignment(outDir, refFile, inDataListSeg, inDataListImg=[]):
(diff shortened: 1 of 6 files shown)


Examples/Python/GroomUtils.py
```diff
@@ -177,21 +181,23 @@ def applyResampling(outDir, refFile, inDataListSeg, inDataListImg=[]):
...
-            rawoutname = rename(inDataListImg[i], rawoutDir, 'resized')
+            rawoutname = rename(inDataListImg[i], rawoutDir, 'aligned')
             outRawDataList.append(rawoutname)
 
         # resize images to reference images
         img = Image(inDataListSeg[i])
-        transform = createTransform(Matrix()) # create identity matrix
-        img.resample(transform, refImg.origin(), refImg.dims(), refImg.spacing(), refImg.coordsys(), InterpolationType.NearestNeighbor).write(segoutname)
+        rigidTransform = ImageUtils.createRigidRegistrationTransform(Image(inDataListSeg[i]).antialias(antialias_iterations), refImg, isoValue, icp_iterations)
+        img.resample(rigidTransform, refImg.origin(), refImg.dims(), refImg.spacing(), refImg.coordsys(), InterpolationType.Linear).write(segoutname)
```
archanasri, line 196: You can call img.applyTransform() instead.

Examples/Python/GroomUtils.py
```diff
@@ -177,21 +181,23 @@ def applyResampling(outDir, refFile, inDataListSeg, inDataListImg=[]):
...
+
         if inDataListImg:
             img = Image(inDataListImg[i])
-            img.resample(transform, refImg.origin(), refImg.dims(), refImg.spacing(), refImg.coordsys(), InterpolationType.NearestNeighbor).write(rawoutname)
+            img.resample(rigidTransform, refImg.origin(), refImg.dims(), refImg.spacing(), refImg.coordsys(), InterpolationType.Linear).write(rawoutname)
```
archanasri, line 200: You can call img.applyTransform() instead.

Examples/Python/GroomUtils.py
jadie1: We'd still need to call resample though. Would is be better to call it in two steps (applyTransform then resample)?

archanasri: applyTransform() actually calls resample(). The idea is to create a "transform" and then "apply" it.

jadie1: Got it, I just tested it and it works as expected so I pushed that change.

archanasri (approved):

akenmorris:
@sheryjoe , can we merge this?

Examples/Python/GroomUtils.py
```diff
@@ -103,7 +103,7 @@ def applyCOMAlignment(outDir, inDataListSeg, inDataListImg, processRaw=False):
         T = img.center() - img.centerOfMass()
 
         # binarize result since linear interpolation makes image blurry again
-        img.translate(T).binarize().write(outname)
+        img.translate(T).binarize().recenter().write(outname)
```
sheryjoe, line 106: @jadie1 is this an image (CT/mri) , segmentation or antialiased segemenation? We should n't transform/resample a segmentation directly? we antialias it first.

Examples/Python/GroomUtils.py
```diff
@@ -155,12 +155,16 @@ def FindReferenceImage(inDataList):
...
-def applyResampling(outDir, refFile, inDataListSeg, inDataListImg=[]):
+def applyRigidAlignment(outDir, refFile, inDataListSeg, inDataListImg=[]):
     """
     This function takes in a filelists(binary and raw) and makes the 
     size and spacing the same as the reference
     """
-    print("\n############# Resample #############")
+    isoValue       = 1e-20
+    icp_iterations = 200
+    antialias_iterations = 1
```
sheryjoe, line 165: I would increase this. Maybe 20 or 30 iterations?

Examples/Python/GroomUtils.py
```diff
@@ -155,12 +155,16 @@ def FindReferenceImage(inDataList):
...
 
-def applyResampling(outDir, refFile, inDataListSeg, inDataListImg=[]):
+def applyRigidAlignment(outDir, refFile, inDataListSeg, inDataListImg=[]):
     """
     This function takes in a filelists(binary and raw) and makes the 
     size and spacing the same as the reference
     """
-    print("\n############# Resample #############")
+    isoValue       = 1e-20
+    icp_iterations = 200
```
sheryjoe, line 164: icp iterations should be added as input argument with a default? 100 vs 500 makes a difference in the alignment result based on how misaligned the shapes are.

Examples/Python/GroomUtils.py
```diff
@@ -177,21 +181,23 @@ def applyResampling(outDir, refFile, inDataListSeg, inDataListImg=[]):
...
-            rawoutname = rename(inDataListImg[i], rawoutDir, 'resized')
+            rawoutname = rename(inDataListImg[i], rawoutDir, 'aligned')
             outRawDataList.append(rawoutname)
 
         # resize images to reference images
         img = Image(inDataListSeg[i])
-        transform = createTransform(Matrix()) # create identity matrix
-        img.resample(transform, refImg.origin(), refImg.dims(), refImg.spacing(), refImg.coordsys(), InterpolationType.NearestNeighbor).write(segoutname)
+        rigidTransform = ImageUtils.createRigidRegistrationTransform(Image(inDataListSeg[i]).antialias(antialias_iterations), refImg, isoValue, icp_iterations)
+        img.applyTransform(rigidTransform, refImg.origin(), refImg.dims(), refImg.spacing(), refImg.coordsys(), InterpolationType.Linear).write(segoutname)
```
sheryjoe, line 196: @jadie1 i think we should antialias before apply transform. @archanasri this function does image resampling given the transform, right?

Examples/Python/GroomUtils.py
```diff
@@ -177,21 +181,23 @@ def applyResampling(outDir, refFile, inDataListSeg, inDataListImg=[]):
...
+
         if inDataListImg:
             img = Image(inDataListImg[i])
-            img.resample(transform, refImg.origin(), refImg.dims(), refImg.spacing(), refImg.coordsys(), InterpolationType.NearestNeighbor).write(rawoutname)
+            img.applyTransform(rigidTransform, refImg.origin(), refImg.dims(), refImg.spacing(), refImg.coordsys(), InterpolationType.Linear).write(rawoutname)
```
sheryjoe, line 200: same here.

Examples/Python/ellipsoid.py
```diff
@@ -28,7 +28,7 @@ def Run_Pipeline(args):
     if int(args.interactive) != 0:
         input("Press Enter to continue")
     # Get data
-    datasetName = "ellipsoid-v0"
+    datasetName = "ellipsoid-v2"
```
sheryjoe, line 31: should we move to ellipsoid and ellispoid-aligned rather than v2 and v0?

Examples/Python/ellipsoid.py
```diff
@@ -84,13 +84,14 @@ def Run_Pipeline(args):
         paddedFiles = applyPadding(groomDir + "padded/segmentations", centeredFiles, 10)
 
         """Apply center of mass alignment"""
-        comFiles = applyCOMAlignment(groomDir + "com_aligned/segmentations", centeredFiles, None)
+        comFiles = applyCOMAlignment(groomDir + "com_aligned/segmentations", paddedFiles, None)
```
sheryjoe, line 87: I think we don't need padding as part of the workflow before alignment. we can pad when we do cropping. Please see the grooming notebook. @jadie1

Examples/Python/ellipsoid_cut.py
```diff
@@ -18,15 +18,15 @@
...
+    The data (ellipsoid-aligned.zip) is dowloaded to /Data folder and extracted to 
     /Output/ellpsoid_cut
     """
 
     print("\nStep 1. Extract Data\n")
     if int(args.interactive) != 0:
         input("Press Enter to continue")
     # Get data
-    datasetName = "ellipsoid-v0"
+    datasetName = "ellipsoid_aligned-v0"
```
sheryjoe, line 29: do we use underscore vs hyphen in datasets name? let's be consistent.

Examples/Python/ellipsoid_cut.py
```diff
@@ -43,63 +43,11 @@ def Run_Pipeline(args):
...
-        print("\nStep 3. Groom - Convert to distance transforms\n")
-        if int(args.interactive) != 0:
-            input("Press Enter to continue")
-
-        dtFiles = applyDistanceTransforms(groomDir, croppedFiles)
+    print("\nStep 2. Get distance transforms\n")
+    groomDir = outputDirectory + 'groomed/'
+    if not os.path.exists(groomDir):
+        os.makedirs(groomDir)
+    dtFiles = applyDistanceTransforms(groomDir, fileList)
```
sheryjoe, line 50: does this function apply the full grooming pipeline?

sheryjoe (changes requested):

Examples/Python/GroomUtils.py
archanasri: Yes, that's right.

Examples/Python/ellipsoid_cut.py
jadie1: We already have left_atrium-v0 and ellipsoid_fd-v0 so I was trying to be consistent with those. The use case .py files use underscores (left_atrium.py, ellipsoid_fd.py, deep_ssm.py, etc.) so I think the datasets were named to be consistent with that. But we don't have a use case called "ellipsoid_aligned" use case so I can change it to a hyphen if you think that's better in this case.

Examples/Python/ellipsoid.py
jadie1: We could make ellipsoid-v2 just ellipsoid since it is a different dataset altogether than ellipsoid-v0. I think we should have a bigger discussion about the versioning because it could get confusing.

Examples/Python/GroomUtils.py
jadie1: Okay. I didn't remove antialiasing, it was never in the groomutils COM alignment. I will add it!

jadie1: Added b0b2e9d

Examples/Python/GroomUtils.py
jadie1: Added 2d12d22

Examples/Python/GroomUtils.py
jadie1: Added 2d12d22

Examples/Python/GroomUtils.py
jadie1: Added 5c27af1

Examples/Python/GroomUtils.py
jadie1: Same here meaning they should be antialiased? These are the raw images not segmentations, do we need to antialias them?

Examples/Python/ellipsoid_cut.py
jadie1: No the ellipsoid use case illustrates the grooming pipeline. We decided to just have this use case illustrate the cutting plane so we started with aligned data and just apply the distance transform.

Examples/Python/ellipsoid.py
jadie1: Fixed b3d9b55

Examples/Python/GroomUtils.py
sheryjoe: no, if they are images, no antialiasing needed.

Examples/Python/ellipsoid.py
sheryjoe: Till Kitware have a better solution for versioning, we will stick to -vN suffix for datasets. What I meant is rather than having ellipsoid-v0 that implicitly refer to groomed ellispsoids and ellpsoid-v2 that implicitly refer to ungroomed ones, we should consider having ellipsoid-v1 and ellispoid-aligned-v1 (that contain regenerated ellispoids without misalignment using the shape cohort generator). This way it is more explicit and consistent.

Examples/Python/ellipsoid_cut.py
sheryjoe: This is reasonable. let's stick with underscores then. We can use hyphen only for versioning, this will be consistent with RunUseCase input arguments too.

Examples/Python/ellipsoid.py
jadie1: I see, so should I just rename ellipsoid-v2 to ellipsoid-v1 and ellipsoid_aligned-v0 to ellipsoid_aligned-v1?

sheryjoe: Yes. I assume ellispoid_aligned-v1 has data generated from the cohort generator.

jadie1: Yes that's right

jadie1: Addressed b5152c3

jadie1:
Ready to merge!

Issue #914, opened 2021-01-22
Mesh and particles disappear in studio after optimization


medakk:
* open Studio, add a single mesh
* "Skip" grooming
* run optimize
* everything is fine during optimization, but at the end everything disappears:
![image](https://user-images.githubusercontent.com/10536745/105560535-44f8ac00-5cd1-11eb-8a04-ad125fbb67e4.png)
* switch to "groom" tab, everything is back to normal again
![image](https://user-images.githubusercontent.com/10536745/105560573-60fc4d80-5cd1-11eb-9d63-1e25bdf4cdf9.png)
* can switch back to optimize now if needed

In the console I see "Error: asked for groomed mesh when none is present!"

akenmorris:
Done in #927

Issue #913, opened 2021-01-22
Studio: automatic glyph size


akenmorris:
Would be really nice to have a checkbox for automatic glyph size.  Switching between projects of different sizes, I find I have to adjust it often.

sheryjoe:
@akenmorris is this resolved? close?

akenmorris:
Yes.

Issue #911, opened 2021-01-22
Studio: Allow choice of surface reconstruction method based on data


akenmorris:
Depending on the data, multiple of the surface reconstruction methods will work.

If the project has DTs, then there should be 3 options:

1) legacy
2) DT based
3) Mesh Warp based (from meshes of groomed files)

akenmorris:
Done in #927

Issue #902, opened 2021-01-15
Studio: optimize tool should load images in the worker thread


akenmorris:
Right now, the images are loaded prior to spawning the thread.  For large projects, this results in the UI being unresponsive as the optimization is being initialized.  This should be moved to run inside the worker thread and allow for canceling as well.

akenmorris:
Done in #927

PR #915, opened 2021-01-25
Studio mesh reconstruction

akenmorris:

#914 - Mesh and particles disappear in studio after optimization
#913 - Studio: automatic glyph size 
#911 - Studio: Allow choice of surface reconstruction method based on data
#902 - Studio: optimize tool should load images in the worker thread
#822 - Studio should auto-run the surface reconstruction like View2 does
#844 - Mesh reconstruction via template transform to studio for mesh support 
#917 - Studio: Particle update interval needs to be tuned based on refresh duration
#918 - Studio: Display filenames for old XML inputs

studio_analyze_surface_reconstruction


medakk:
This is great stuff!! :D 


I noticed this weird behaviour: after optimizing, if I go to analyze, it seems to be using mesh-warping based reconstruction but the radio button says its using legacy:
image

akenmorris:
@medakk , yes indeed.  Will push fixes shortly.

Libs/Mesh/Mesh.h
```diff
@@ -14,8 +15,10 @@ class Mesh
   using MeshType = vtkSmartPointer<vtkPolyData>;
 
   Mesh(const std::string& pathname) : mesh(read(pathname)) {}
-  Mesh(vtkSmartPointer<vtkPolyData>&& rhs) : mesh(std::move(rhs)) {}
-
+  // Mesh(vtkSmartPointer<vtkPolyData>&& rhs) : mesh(std::move(rhs)) {}
```
cchriste, line 18: Please see my comments about this in the mesh warp pull request.

cchriste (changes requested):
Looks mostly okay. Just the pieces from the mesh warp pull request (modifications to Mesh constructors and assignment operators, etc., just see the comments over there). Otherwise, great.

I suggest merging it with the new alpha branch, thus enabling testing the consolidated mesh work from three branches before springing it on an unsuspecting master branch.

Issue #876, opened 2021-01-06
PyTorch installation issue in conda_installs


iyerkrithika21:
Using conda to install PyTorch installs an older version(torch 1.0.0), which is not compatible with Cuda versions of 10 and 11.
This results in an error when running the Deep SSM use case.

PR #916, opened 2021-01-25
Pytorch

jadie1:
This PR attempts to address issue #876 
Ultimately we decided it is best if we continue to just support the most recent release of PyTorch but add some documentation describing what to do if is not compatible with someone's CUDA version or if `nvidia-smi` does not correctly detect the CUDA version.

I added CUDA 11.0 to the list because it is now supported as can be seen here: [pytorch.org](https://pytorch.org/)

The documentation has been added on the mkdocs branch here: [pytorch-gpu.md](https://github.com/SCIInstitute/ShapeWorks/blob/mkdocs/docs/deep-learning/pytorch-gpu.md)


akenmorris:
Looks fine to me.  There seem to be a bunch of conflicts if you check the linux build log.

jadie1:
I think it's because I actually called the install_pytorch function. It's been removed in the master branch

akenmorris:
Makes sense.  Now how to resolve the conflict?

jadie1:
@akenmorris  Looks like it built with pip!

iyerkrithika21 (approved):

PR #923, opened 2021-01-27
Fix leak in Mesh::coverage by using smart pointers

akenmorris:

### DIFF
--- a/Libs/Mesh/Mesh.cpp
+++ b/Libs/Mesh/Mesh.cpp
@@ -140,8 +140,8 @@ Mesh& Mesh::write(const std::string &pathname)
 Mesh& Mesh::coverage(const Mesh &otherMesh, bool allowBackIntersections, double angleThreshold, double backSearchRadius)
 {
   FEVTKimport import;
-  FEMesh* surf1 = import.Load(this->mesh);
-  FEMesh* surf2 = import.Load(otherMesh.mesh);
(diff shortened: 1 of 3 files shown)


cchriste (changes requested):
I thought of doing this, but it's likely to crash later on since this function (called by areaCoveraget.Apply where surf1 and surf2 are used) takes ownership of the raw pointer by address:
```c++
void FEAreaCoverage::Surface::Create(FEMesh& mesh)
{
  m_mesh = &mesh;
...
class FEAreaCoverage
{
  class Surface
  {
public:
    FEMesh*       m_mesh;
...
```

akenmorris:
Yeah, these PreviewMeshQC classes need some work, don't they?

cchriste:
Yeah, let's just close this pull request and create an issue so we don't accidentally break something. A memory leak is bad, but freeing an object that may still be in use is worse.

akenmorris:
I've pushed changes that replace all the pointers and references with shared pointers.

There is no critical need to merge this immediately.  The only code using this is to run it from the command line which a user will only do once per process timeline.

cchriste:
Cool. After a quick perusal, I feel this update is solid. Just to be extra safe let's wait till after the release to merge it.

cchriste:
We've enough work left on this release (e.g., add the remaining Mesh API to Python) that this change will still have plenty of time to be tested. Plus it looks safer than what we have. Merging. (the failed test was fixed awhile ago in the main branch)

PR #924, opened 2021-01-28
Mesh warping only

akenmorris:
This PR updates the mesh warping PR to merge against mesh_party/alpha branches.  I've removed the stand alone executable.


archanasri:
Can MeshUtils::distilVertexInfo() and MeshUtils::MeshUtils::distilFaceInfo() be called from MeshUtils::generateWarpMatrix() instead of calling them separately and passing them to generateWrapMatrix()?

akenmorris:
> Can MeshUtils::distilVertexInfo() and MeshUtils::MeshUtils::distilFaceInfo() be called from MeshUtils::generateWarpMatrix() instead of calling them separately and passing them to generateWrapMatrix()?

No, the distilled face info must be passed to the warpMesh call, for example.

The way this is used in Studio for example, is to generate the warp once and then use warpMesh later, many times.  We keep the face info for later use.

Libs/Mesh/MeshUtils.cpp
```diff
@@ -43,6 +57,125 @@ const vtkSmartPointer<vtkMatrix4x4> MeshUtils::createICPTransform(const vtkSmart
...
+  return W;
+}
+
+Mesh MeshUtils::warpMesh(Eigen::MatrixXd movPts, Eigen::MatrixXd W, Eigen::MatrixXi Fref){
+  
+  int numVertices = W.rows();
+  int numFaces = Fref.rows();
+  // std::cout<< (Vcontrol_moving.rowwise() + Eigen::RowVector3d(0,0,0)) <<std::endl;
+	Eigen::MatrixXd Voutput = W * (movPts.rowwise() + Eigen::RowVector3d(0,0,0));
+	vtkSmartPointer<vtkPolyData> outmesh = vtkSmartPointer<vtkPolyData>::New();
```
archanasri, line 139: vtkSmartPointer<vtkPolyData> is defined as MeshType in Mesh.h

akenmorris: Yes, I see that.  Personally, when I am writing VTK code, I would rather do the following:

```
vtkSmartPointer<vtkPolyData> outmesh
vtkSmartPointer<vtkPoints> outpoints
vtkSmartPointer<vtkCellArray> polys
..
```

Vs.:

```
MeshType outmesh
vtkSmartPointer<vtkPoints> outpoints
vtkSmartPointer<vtkCellArray> polys
...
```

For myself, when I see the latter, I think, MeshType?  What's that?  And then I have to look it up.

cchriste: I have the same feeling about this.

More broadly, for ShapeWorks we should pass around Mesh and Image rather than their internal data structures outside of the functions that need them. (We're already mostly doing this, but it's worth reiterating)

Libs/Mesh/MeshUtils.cpp
```diff
@@ -43,6 +57,125 @@ const vtkSmartPointer<vtkMatrix4x4> MeshUtils::createICPTransform(const vtkSmart
...
+		vtkIdType a, b, c;
+		polys->InsertNextCell(3);
+		polys->InsertCellPoint(Fref(i, 0));
+		polys->InsertCellPoint(Fref(i, 1));
+		polys->InsertCellPoint(Fref(i, 2));
+	}
+	outmesh->SetPoints(outpoints);
+	outmesh->SetPolys(polys);
+	
+  return Mesh(outmesh);
```
cchriste, line 158: 👍

Testing/MeshTests/MeshTests.cpp
```diff
@@ -295,3 +299,50 @@ TEST(MeshTests, icpTest)
 
   ASSERT_TRUE(source == ground_truth);
 }
+
+
+TEST(MeshTests, warpTest1)
```
cchriste, line 304: These tests are still pretty big. One thing that could be done to improve them is to add more assertions throughout. Otherwise if the test fails there could be a dozen different places where something went wrong.

cchriste (approved):
This is _much_ cleaner. 👍

I don't know how much work would be involved to save and compare data from intermediate steps in the two tests (even epsEq average values could be useful), so let's not unduly hold up this pull request if it seems too much.

akenmorris:
FYI, this PR is failing on Windows because mesh_party/alpha is not currently building on Windows.

Libs/Mesh/MeshUtils.cpp
archanasri: Makes sense.

archanasri (approved):

Issue #918, opened 2021-01-26
Studio: Display filenames for old XML inputs


akenmorris:
Currently filenames are not shown when an XML input is given.

sheryjoe:
@akenmorris is this resolved by #927? can we close it?

akenmorris:
Yes.

Issue #917, opened 2021-01-26
Studio: Particle update interval needs to be tuned based on refresh duration


akenmorris:
Should make the interface more responsive during optimization.

Issue #844, opened 2020-12-07
Mesh reconstruction via template transform to studio for mesh support 


riddhishb:
Based on Issues
#684 and #166 
Need to include the API for the mesh reconstruction for the shapeworks studio.

akenmorris:
Done in #927

Issue #822, opened 2020-11-17
enable python and shapeworks tests to be run individually


cchriste:
From @akenmorris:
when I run:

`PythonTests --gtest_filter=pythonTests.resize`

It doesn't work (can't find resize.py).
This makes it hard to work on one test (at the end) if you have to wait and watch all the other tests complete.

The reason for this is both Python and the shapeworks executable use external system calls for testing, and the environment is only setup once for all the tests in these groups:
```c++
//---------------------------------------------------------------------------
// only need to run shapeworksEnvSetup once or it continuously appends to paths
TEST(shapeworksTests, setup)
{
  shapeworksEnvSetup();
  ASSERT_TRUE(true);
}
```

Possible solutions:
- `shapeworksEnvSetup()` not simply append to the PATH/PYTHONPATH/LD_LIBRARY_PATH of the user's environment
- simply tolerate that these envvars are appended before each test is executed (it may be innocuous)
- check if `shapeworksTests.setup` (or `pythonTests.setup`) have completed successfully before running other tests

Related to #821.

cchriste:
Related to #594

cchriste:
Yay this is fixed now!!!

PR #927, opened 2021-02-01
Studio mesh reconstruction

akenmorris:

#914 - Mesh and particles disappear in studio after optimization
#913 - Studio: automatic glyph size
#911 - Studio: Allow choice of surface reconstruction method based on data
#902 - Studio: optimize tool should load images in the worker thread
#822 - Studio should auto-run the surface reconstruction like View2 does
#844 - Mesh reconstruction via template transform to studio for mesh support
#917 - Studio: Particle update interval needs to be tuned based on refresh duration
#918 - Studio: Display filenames for old XML inputs
#929 - Set scalar bar text color to opposite of background.
#917 - Studio: Particle update interval needs to be tuned based on refresh duration
#913 - Studio: automatic glyph size


Libs/Groom/Groom.cpp
```diff
@@ -353,9 +354,8 @@ void Groom::set_skip_grooming(bool skip)
 //---------------------------------------------------------------------------
 std::shared_ptr<Mesh> Groom::load_mesh(std::string filename)
 {
-  tbb::mutex::scoped_lock lock(this->mutex_);
   try {
-    auto mesh = std::make_shared<Mesh>(filename);
+    auto mesh = std::make_shared<Mesh>(MeshUtils::threadSafeReadMesh(filename));
```
cchriste, line 358: 👍

cchriste (approved):
These changes look very good. Let's get this branch merged so we have the full release ready to go.

Issue #936, opened 2021-02-02
TriMeshWrapper::ComputeBarycentricCoordinates


akenmorris:
TriMeshWrapper::ComputeBarycentricCoordinates does not handle degenerate triangles.  One of the femur groom meshes and the clipped boundary ends up having a face with two vertices in the same location.  This is likely due to the low precision of the vertex positions in the PLY file.

When given a triangle with shared vertices, this function returns a bunch of NaNs which end up breaking many other things in the optimizer.  This results in either crashing (see potentially #925) or hanging/infinite loop.

This function either has to be fixed to handle degenerate triangles, or we need to ensure that these triangles to not get passed to it.

Also related: #910

medakk:
@akenmorris Does vtkCleanPolyData let us close this?

akenmorris:
Yep

Issue #926, opened 2021-02-01
Replace trimesh2 with VTK or other mesh library


akenmorris:
The trimesh2 library we are using is unsupported and uses very questionable code constructs (e.g. **goto**s) and may be a source of bugs (see #925).  It also has support for only one file format, PLY and has no support for scalar values.

I propose replacing it with VTK or some other mesh library.

medakk:
@akenmorris a thought I had:

after geodesic walk, we know exactly which triangle the particle is located on. it makes sense to immediately update the `particle2tri_` cache at the end of geodesic walk instead of (potentially) doing a kdtree lookup the next iteration.

akenmorris:
Correct, I’m doing this in the replace_trimesh branch.  There were a few other caching opportunities as well.

sheryjoe:
Based on the related issue #925, I think this should be considered before the 6 release. @akenmorris @medakk ?

Issue #910, opened 2021-01-22
ShapeWorks Optimize needs some input verification


akenmorris:
Many users and developers struggle with crashes in ShapeWorks Optimize step that are the result of grooming errors and other input errors.  It would be nice if the optimizer did some input verification before running.

Some things that could be checked:

1) DT inputs are actually DTs
2) Surface on DT doesn't run off the edge
3) ...

medakk:
Here's an example the surface clipping the domain:

![image](https://user-images.githubusercontent.com/10536745/105527464-fa5f3b80-5ca0-11eb-8bdf-34dccea52110.png)

(in this instance, the optimizer would crash complaining about the narrow band)

akenmorris:
Thanks, @medakk , good example.  We will need to add the nrrd files to our tests.

iyerkrithika21:
> 
> 
> Thanks, @medakk , good example. We will need to add the nrrd files to our tests (once we have the checks)

I can share the files which have such clipping.

medakk:
+optimize should check whether the provided XML file actually exists. Currently it exits with a vague error: 

```
Could not parse XML
Inconsistency in parameters... m_domains_per_shape != m_number_of_particles.size()
```

medakk:
+if the an input filename doesn't exist, optimize should crash immediately. currently it floods the terminal with other diagnostic info and makes it hard to realize the actual problem

+if the output dir is /existing_folder/folder_that_doesnt_exist/another_folder_that_doesnt_exist/ optimize just segfaults. check for this

iyerkrithika21:
Not sure if you have covered this scenario in the previous comments.
When a list of meshes/segs/images is passed to the optimize function, it should check if the list is empty or not. Right now it crashes with the following error message :
```
Unzipping Data/ellipsoid_md_dist.zip into Output/ellipsoid_md_dist/
CALLING OPTIMIZATION CODE
Output/ellipsoid_md_dist/shape_models/correspondence_512_512.xml
'optimize' FAILED: basic_string::_M_construct null not valid
General exception caught.
	Returncode: 1
	Output: None
```

akenmorris:
> +if the an input filename doesn't exist, optimize should crash immediately.

I assume you mean exit, not crash?  :)

medakk:
@iyerkrithika21 Thanks!

@akenmorris Yikes, yes. :'D

medakk:
[...]

PR #940, opened 2021-02-03
Do not consider faces that result in NaN barycentric coordinates

akenmorris:
This is a temporary fix until we either add a QC check (e.g. something
like vtkCleanPolyData) or make other changes.

The bug that this fixes is where the mesh contains degenerate faces
(such as two vertices of a triangle that land on the same location).
The current computation returns barycentric coordinates containing
NaNs which end up propagating to many other variables resulting in
either infinite loops or crashes.

See #936 - TriMeshWrapper::ComputeBarycentricCoordinates 
See #926 - Replace trimesh2 with VTK or other mesh library
See #910 - ShapeWorks Optimize needs some input verification


akenmorris:
Closing for now while we investigate further.  This fixes one issue, but not everything.

PR #946, opened 2021-02-05
Add missing printCmd arguement

jadie1:
Fixing small groom utils bug


cchriste:
This is confusing. Why are we only using part of the Python interface?
Oh. Because only some of the new Mesh functions have been added to the Python API.

Okay, next confusion: if it's important enough to pass a parameter to this function to print stuff, why are we printing some commands but not details of every operation? Just look at what's going on with reflect at the end of this function and you can be confused, too.

Issue #961, opened 2021-02-11
Studio: Sample display bug


akenmorris:
If you load an XML project for analysis, switch to PCA, click animate, then click samples, only one shape displays.

Issue #960, opened 2021-02-11
Studio: Only display basename of file name


akenmorris:
<img width="511" alt="image" src="https://user-images.githubusercontent.com/1693349/107665967-34da4980-6c4b-11eb-9dc8-3a89062800b3.png">

It would be better to drop the path.  It's not likely that the user will be using inputs of the same filename that are in different paths.

PR #962, opened 2021-02-11
Studio fixes

akenmorris:
Resolves the following issues:

#961 - Studio: Sample display bug
#960 - Studio: Only display basename of file name 
#959 - Studio: When optimization is aborted, the file menu does not re-enable
#963 - Studio: Bug when switching out of analysis while PCA animation is running


jadie1 (approved):
I tested with segmentations and meshes and everything looked good.

Issue #615, opened 2020-07-01
for OSX, add ability to execute ShapeWorksStudio from terminal when shapeworks' bin is in PATH


cchriste:
OSX creates app bundles for executables, so the path to ShapeWorksStudio is actually $SHAPEWORKS_BIN/ShapeWorksStudio.app/Contents/MacOS. The user can explicitly add both to their own path, but it would be nice to only need SHAPEWORKS_BIN rather than always setting path to include both locations (especially for people with multiple build trees). 

Perhaps an executable shell script in $SHAPEWORKS_BIN?

sheryjoe:
This is related to #478

akenmorris:
How about a symlink?

sheryjoe:
@akenmorris @cchriste what is the status of this issue?

akenmorris:
@sheryjoe https://github.com/SCIInstitute/ShapeWorks/pull/967

cchriste:
No script in result of GitHub action for this pull request, and cannot test locally due to broken make install (#953).

akenmorris:
Completed.

PR #967, opened 2021-02-12
Address #615 by creating a script to run ShapeWorksStudio from bin

akenmorris:
This addresses #615.

More complicated than I would have liked, but a symlink doesn't work and you can't directly write an executable file using CMake, so it has to write elsewhere and copy.

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -140,8 +140,3 @@ jobs:
       working-directory: ${{runner.workspace}}/build
       shell: bash -l {0}
       run: rm -rf c:/Miniconda3
-             
-    - name: df2
-      working-directory: ${{runner.workspace}}/build
-      shell: bash -l {0}
-      run: df -h
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -46,6 +46,7 @@ jobs:
       with:
         version: "5.9.9"
         setup-python: 'false'
+        dir: "C:\Qt"
         cached: ${{ steps.cache-qt.outputs.cache-hit }}
 
     - name: Setup Conda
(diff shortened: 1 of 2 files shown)


cchriste:
I don't know how to verify this works because `make install` doesn't work for me. 
Here are the errors when I run make install:
```
ERROR: no file at "/usr/local/opt/libiodbc/lib/libiodbc.2.dylib"
ERROR: no file at "/Applications/Postgres.app/Contents/Versions/9.6/lib/libpq.5.dylib"
```

cchriste:
The latest pre-release build from four days ago doesn't seem to have the script, so I can't try that either.

cchriste:
> The latest pre-release build from four days ago doesn't seem to have the script, so I can't try that either.

I also tried the [action from this pull request](https://github.com/SCIInstitute/ShapeWorks/actions/runs/562275972), which has no such script.

akenmorris:
> I also tried the action from this pull request, which has no such script.

Yep, it just wasn't making it into the installer.  Should be fixed now (I see it in the deployment PR file)

cchriste:
Finally able to download the GitHub artifact and ShapeWorksStudio.exe works!

cchriste (approved):

cchriste:
And the Nullsoft installer: sweet! 
So both the start menu item and the command line version run. Not sure how the start menu version activates the conda environment, and didn't go deep trying to do anything, but it started up fine.

akenmorris:
> So both the start menu item and the command line version run. Not sure how the start menu version activates the conda environment, and didn't go deep trying to do anything, but it started up fine.

What start menu version?  For Python?

cchriste:
For Studio. That’s the only one I saw after installation.

akenmorris:
Oh, I see.  None of our executables (including studio) require conda to be installed in order to run.  Users only need to install and activate conda to run the python stuff.  This is the same for your local builds on your machine, you don't need to activate your conda environment to run the `shapeworks` executable or `ShapeWorksStudio`.

Issue #921, opened 2021-01-27
SW discussion forum


sheryjoe:
- [x] Update the [contact](http://sciinstitute.github.io/ShapeWorks/about/contact.html) to reflect the discourse discussion forum instead of gitter.
- [x] Update team page

sheryjoe:
@jadie1 Please update the documentation to refer to the discussion forum instead of gitter.

jadie1:
> @jadie1 Please update the documentation to refer to the discussion forum instead of gitter.

Done on the mkdocs branch

sheryjoe:
Addressed in PR #968

PR #968, opened 2021-02-15
Mkdocs: Updated team and contact

sheryjoe:
- Addressing #921
- updated data augmentation doc
- added doc from new notebooks
- added instructions for pytorch gpu


sheryjoe:
@akenmorris could you please look into this?

image

akenmorris:
I re-ran this workflow and it worked but it did give a warning about low disk space.  I googled around and I think the runner gives you about 14GB to work with.  Surprising that we're running out.  Not sure what to do if that's the case.

PR #1029, opened 2021-02-18
Added condition for unziiping data

iyerkrithika21:


jadie1 (approved):
I tested different orders of running the full use case and tiny test, looks good!

Issue #1026, opened 2021-02-18
RunUseCase.py doesn't work for users with current instructions


akenmorris:
I just tried testing the deployments for all three platforms, following the instructions given.  The python use cases won't run. Each one fails with:

```
ModuleNotFoundError: No module named 'shapeworks'
```

Should we add this to the top of RunUseCase.py?

```
import setupenv
setupenv.setup_shapeworks_env()
```

iyerkrithika21:
Adding `setupenv.setup_shapeworks_env()` could work because most of the times, this error goes away when `ShapeWorks/build/bin/` is added to the `PYTHONPATH`

akenmorris:
I believe I've fixed it for all platforms.  Ok to merge?

jadie1:
Reflection was still not working when femurs are groomed with meshes. @archanasri we need the shapeworks.cpp change

archanasri:
@jadie1 what change are we talking about?

jadie1:
Sorry ignore that, wrong issue, got mixed up.

sheryjoe:
Where are we on this issue?

archanasri:
The fix has been merged to the release branch.

PR #1033, opened 2021-02-19
Addressing #1026 to add shapeworks_path

archanasri:


akenmorris:
I tried this on Windows and got:

```
  File "RunUseCase.py", line 65, in <module>
    module = __import__(args.use_case.lower())
  File "C:\Users\amorris\Desktop\Examples\Python\ellipsoid.py", line 13, in <module>
    from GroomUtils import *
  File "C:\Users\amorris\Desktop\Examples\Python\GroomUtils.py", line 14, in <module>
    from shapeworks import *
ModuleNotFoundError: No module named 'shapeworks'
```

akenmorris:
If I change:

`setupenv.setup_shapeworks_env(shapeworks_bin_dir=args.shapeworks_path)`

to

`setupenv.setup_shapeworks_env()`

Then it works.  I believe the problem is that args.shapeworks_path is "", which is not the same as None which setup_shapeworks_env looks for to set paths automatically.

akenmorris:
Nevermind it's not "", it's ";" (semicolon) (on windows).

Can someone explain this line?

`parser.add_argument("--shapeworks_path", help="Path to ShapeWorks executables (default: "+default_binpath+")", nargs='?', type=str, default=os.pathsep)`

Why does it say the default is `default_binpath`, but the it's set to `os.pathsep` instead?

@cchriste  you are listed as the author of this change.

PR #1034, opened 2021-02-19
Relax optimize input constraints in Studio

### LINKED ISSUE #1031: Studio: Parameter ranges in Studio seem to be too constrained
Based on parameters previously used for our hip models, the parameter ranges in Studio are too constraining. Additionally, I wonder if the inclusion of decimals on most parameters is non-intuitive for new users.

- Initial relative weighting - can this have three decimals? We have used 0.001 previously.
- Relative weighting - does this need a decimal point? I don't know that I've ever used <1
- Starting regularization - max is 99,999 while we have used 1e9 previously; does this need a decimal point?
- Normals strength - do you expect users to use decimal values or can this be integers?

akenmorris:
Resolve #1031 

Rather than use QDoubleSpinBoxes for these values that must have a number of decimals set, I've switched to free response line edits with validators.  When a param is invalid, the background changes to red and the "run optimize" button disables.  This allows the user to easily type "1e9" for a value as they support scientific notation for the double parameters.

### DIFF
--- a/Studio/src/Application/Optimize/OptimizeTool.cpp
+++ b/Studio/src/Application/Optimize/OptimizeTool.cpp
@@ -4,6 +4,7 @@
 #include <QThread>
 #include <QMessageBox>
 #include <QTemporaryDir>
+#include <QIntValidator>
 
 #include <Optimize/OptimizeTool.h>
 #include <Libs/Optimize/OptimizeParameters.h>
--- a/Studio/src/Application/Optimize/OptimizeTool.cpp
+++ b/Studio/src/Application/Optimize/OptimizeTool.cpp
@@ -20,7 +20,7 @@
 
 #include <ui_OptimizeTool.h>
 
-namespace shapeworks {
+using namespace shapeworks;
 
 //---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


Studio/src/Application/Optimize/OptimizeTool.cpp
```diff
@@ -4,23 +4,20 @@
...
-#include <Data/StudioLog.h>
 
 #include <Optimize/QOptimize.h>
 
-#include <vtkPLYWriter.h>
-
 #include <ui_OptimizeTool.h>
 
-namespace shapeworks {
+using namespace shapeworks;
```
medakk, line 20: just curious, is this convention preferred to the previous one?

medakk (approved):

Studio/src/Application/Optimize/OptimizeTool.cpp
akenmorris: I don't really mind either way.  This does result in one less enclosing { brace.

It wasn't really an intentional change, I was having trouble with some type or something with a Qt slot.

PR #1035, opened 2021-02-19
Fix #964 by targeting fixed libigl fork.

### LINKED ISSUE #964: Crash in Mesh Warp
See discussion here: #898 

```
__pthread_kill 0x00007fff6880333a
pthread_kill 0x00007fff688bfe60
abort 0x00007fff6878a808
__assert_rtn 0x00007fff68789ac6
Eigen::DenseCoeffsBase<Eigen::Matrix<int, -1, 1, 0, -1, 1>, 1>::operator()(long) DenseCoeffsBase.h:425
bool igl::min_quad_with_fixed_precompute<double, Eigen::Matrix<int, -1, 1, 0, -1, 1> >(Eigen::SparseMatrix<double, 0, int> const&, Eigen::MatrixBase<Eigen::Matrix<int, -1, 1, 0, -1, 1> > const&, Eigen::SparseMatrix<double, 0, int> const&, bool, igl::min_quad_with_fixed_data<double>&) min_quad_with_fixed.cpp:83
bool igl::min_quad_with_fixed<double, Eigen::Matrix<int, -1, 1, 0, -1, 1>, Eigen::Matrix<double, -1, 1, 0, -1, 1>, Eigen::Matrix<double, -1, -1, 0, -1, -1>, Eigen::Matrix<double, -1, 1, 0, -1, 1>, Eigen::Matrix<double, -1, -1, 0, -1, -1> >(Eigen::SparseMatrix<double, 0, int> const&, Eigen::MatrixBase<Eigen::Matrix<double, -1, 1, 0, -1, 1> > const&, Eigen::MatrixBase<Eigen::Matrix<int, -1, 1, 0, -1, 1> > const&, Eigen::MatrixBase<Eigen::Matrix<double, -1, -1, 0, -1, -1> > const&, Eigen::SparseMatrix<double, 0, int> const&, Eigen::MatrixBase<Eigen::Matrix<double, -1, 1, 0, -1, 1> > const&, bool, Eigen::PlainObjectBase<Eigen::Matrix<double, -1, -1, 0, -1, -1> >&) min_quad_with_fixed.cpp:578
bool igl::biharmonic_coordinates<Eigen::Matrix<double, -1, -1, 0, -1, -1>, Eigen::Matrix<int, -1, -1, 0, -1, -1>, int, Eigen::Matrix<double, -1, -1, 0, -1, -1> >(Eigen::MatrixBase<Eigen::Matrix<double, -1, -1, 0, -1, -1> > const&, Eigen::MatrixBase<Eigen::Matrix<int, -1, -1, 0, -1, -1> > const&, std::__1::vector<std::__1::vector<int, std::__1::allocator<int> >, std::__1::allocator<std::__1::vector<int, std::__1::allocator<int> > > > const&, int, Eigen::PlainObjectBase<Eigen::Matrix<double, -1, -1, 0, -1, -1> >&) biharmonic_coordinates.cpp:200
W_precomputation(Eigen::Matrix<double, -1, -1, 0, -1, -1>, Eigen::Matrix<double, -1, -1, 0, -1, -1>, Eigen::Matrix<int, -1, -1, 0, -1, -1>, Eigen::Matrix<int, -1, -1, 0, -1, -1>) WarpMeshUsingLandmarks.cxx:56
main WarpMeshUsingLandmarks.cxx:184
start 0x00007fff686bbcc9
```

@riddhishb wrote:

I think I have figured out the issue, the algorithm is interpolating the value of particle to the closest mesh vertex first and then using TPS to interpolate to other vertices, the for loop contining "unknown vertices" do not have a added contion as to what happens when two different control points gets projected to the same nearest vertex.
This as far as I can think can happen in two senarios,

1. one where two control points are very close together (dunno how to decide threshold maybe some percentage of mean inter point distance).

2. When the mesh is sparse and number of points is decently close to number of template mesh vertices (this happens in the data you sent Alan), an easy fix to this is just subdivide the template mesh and increase the number of vertices and that fixed the problem.

There is a way to produce a flag from inside the IGL which checks fro this phenomena, but I do not know how to change teh libigl everytime after build depndency clones a libigl.

For reference, below is a studio project exhibiting this crash.  Simply open and click "Analysis"

[warp_mesh_crash.zip](https://github.com/SCIInstitute/ShapeWorks/files/6012843/warp_mesh_crash.zip)

akenmorris:
This PR should fix #964 by switching to a fixed version of LibIGL

A simple dataset exhibiting the bug is found on the #964 description.  Using this branch (must call build_dependencies.sh again to check out the new libigl), this bug should be fixed.

### DIFF
--- a/build_dependencies.sh
+++ b/build_dependencies.sh
@@ -21,7 +21,7 @@ ITK_VER_STR="5.0"
 QT_MIN_VER="5.9.8"  # NOTE: 5.x is required, but this restriction is a clever way to ensure the anaconda version of Qt (5.9.6 or 5.9.7) isn't used since it won't work on most systems.
 XLNT_VER="v1.4.0"
 OpenVDB_VER="v7.0.0"
-libigl_VER="v2.2.0"
+libigl_VER="v2.2.0-fix"
 
 usage()
 {
@@ -277,7 +277,7 @@ build_igl()
   echo " "
   echo "## Building Libigl..."
   cd ${INSTALL_DIR}
-  git clone https://github.com/libigl/libigl.git
+  git clone https://github.com/akenmorris/libigl.git
   cd libigl
   git checkout -f tags/${libigl_VER}
 


akenmorris:
@riddhishb , no rush, but we should make a bug report / PR to libIGL so that it makes it into their next yearly release and we can switch back to their repo.  I'm happy to submit this, but can you write up a description of the problem and what you did to fix it?

medakk (approved):

PR #1047, opened 2021-02-23
fix memory leak in Sampler::AddMesh()

### LINKED ISSUE #1044: Mesh Domain leaks memory in optimizer
It seems the MeshDomain class doesn't fully implement the ITK smark pointer interface from its parent classes.

medakk:
Make sure MeshDomain uses the itk Smart Pointer so there are no memory leaks

Closes #1044


akenmorris (approved):
Tried it and works great.

PR #1058, opened 2021-02-25
Dataset name and unzipping instruction

iyerkrithika21:
Edited the notebooks, changed ellipsoid-v2 to ellipsoid-v1, and added  instruction to the user that they need to decompress the zip file


akenmorris (approved):
Great!

PR #1060, opened 2021-02-27
Fix use cases for Release 6.0

jadie1:
Fixes so that femur, femur_cut, and left_atrium work for release 6.0


iyerkrithika21:
image

The femur_cut ran alright without the cropping step

iyerkrithika21 (approved):

Issue #1053, opened 2021-02-23
Generalize Image class to handle vector-valued voxels


sheryjoe:
Related to:
#1012 
#1015

Example use cases:
- volumetric correspondences (e.g. diffusion images, multiple clinically-relevant metrics - cortical thickness, bone density ... etc) - related to Andy's project

cchriste:
The very limited skeleton of VectorImage that was just added has an "evaluate" function. Perhaps we should add a similar function to our current Image class to get the weighted scalar value of a given location in space.

cchriste:
Closing as duplicate of #1080.

cchriste:
Is this a duplicate of #1080? If not, can we be more specific about what is being requested? Thanks!

akenmorris:
Duplicate of #1080

Issue #1015, opened 2021-02-17
Consolidate ProbeNormals


archanasri:
Libs/Mesh/ProbeNormals.cxx
After consolidation, add:

1. Python binding
2. MeshTest
3. shapeworksTest
4. PythonTest

cchriste:
Two parts: 
1. compute a vector field (vector image, which we don't have) from a scalar field (the Image class we already have).
2. interpolates that field for a set of points, just need a function to interpolate for one point (like #1005)

archanasri:
This seems to be similar to #1012.

akenmorris:
@sheryjoe, where is this used?

Issue #1005, opened 2021-02-17
Consolidate MeanNormals


archanasri:
Libs/Mesh/MeanNormals.cxx
After consolidation, add:

1. Python binding
2. MeshTest
3. shapeworksTest
4. PythonTest

cchriste:
This interpolates a vector field image for a set of points. It can be a single function to interpolate the vector field for a point called by a loop rather than a function that expects a list of points. 

Similar to #1015, which computes a vector field from a scalar field (function one) and then interpolates that field for a set of points (the part that's very similar to this function).

akenmorris:
@sheryjoe, where is this used?

sheryjoe:
@archanasri this is used when we identify bad and good correspondences based on surface normals.

PR #1061, opened 2021-02-27
VectorImage skeleton

cchriste:
Addresses #1005 and #1015 by constructing vector image from a [distance transform] image and providing `evaluate` function to compute best vector at any given point in space.

First pass at #1053.


archanasri (approved):

akenmorris:
This will probably be useful for #1005 and #1015, but it's not really approaching #1053.

This isn't really a vector image, this is a wrapper around a scalar image that provides a gradient interpolator to evaluate vectors on the fly.  It can't be used for more general vector images.

A vector image, I would expect to use something like `itk::Image<itk::CovariantVector<float, 3>, 3>`.  Such a `VectorImage` class would be able to directly save and load to NRRD files.

I think what's here in this PR is useful, but I would recommend using a different name.  It's more of a GradientImageWrapper or something like that.

cchriste:
> A vector image, I would expect to use something like `itk::Image<itk::CovariantVector<float, 3>, 3>`. Such a `VectorImage` class would be able to directly save and load to NRRD files.
This is exactly what this type is: `using ImageType = GradientImageFilter::OutputImageType;`

> I think what's here in this PR is useful, but I would recommend using a different name. It's more of a GradientImageWrapper or something like that.
That's why I called it a skeleton. It would be nice not to have a separate class for a vector image since a tremendous amount of functionality is identical. And the `evaluate` function would still be in there, and should actually be there in our current Image class already (it's not, but completely reasonable to add).

akenmorris:
> This is exactly what this type is: using ImageType = GradientImageFilter::OutputImageType;

Yes!  My mistake!

akenmorris:
> That's why I called it a skeleton

Sure, I think it's a fine start.  More stuff will be added later.

akenmorris (approved):

PR #1069, opened 2021-03-01
Fixed path regex search in datasubset download

iyerkrithika21:
Now the path regex accounts for "/" and \\\


iyerkrithika21:
@akenmorris this fixes the windows issue I was facing

PR #1070, opened 2021-03-01
Geodesic distances for mesh domain

medakk:
This PR allows ShapeWorks to use geodesic distances instead of euclidean distance for the sampling term in mesh domains. Most of the changes are in VtkMeshWrapper.h, VtkMeshWrapper.cpp and MeshGeoEntry.h

* Adds a new use case thin_cavity_bean. Results on this: (takes around 45 minutes to run)
20210412_thincavitybean
* Adds two new dependencies:
  * [robin-hood-hashing](https://github.com/martinus/robin-hood-hashing) - a faster hash map(10% faster than std::unordered_map). This is a single header dependency, placed in ExternalLibs
  * [geometry-central](https://github.com/nmwsharp/geometry-central) - a geometry processing library used for computing geodesics(3-4x faster than libigl). This is pulled in build_dependencies, and built via cmake `add_subdirectory`
* Some changes to optimizer:
  * Adds a new API `itkParticleDomain::IsWithinDistance()` which returns whether two points are within a given threshold. 
  * Modifies ParticleSurfaceNeighborhood to return a vector of distances, to avoid some redundant computation in EvaluateSigma
  * Modifies MeshDomain to return a point close to the upper bound, to be consistent with image domain. This has been confirmed to cause no problems with existing mesh use cases
  * Two new parameters: `<geodesics_enabled>` and `<geodesics_cache_size_multiplier>`, also exposed in ShapeWorksStudio
* A unit test to verify that the geodesic distance function is returning correct values 

`geodesics_cache_size_multiplier` is used to control how much memory to allocate for the geodesics cache in terms of multiple of the number of triangles in the mesh. If set to 0, the optimizer chooses "120", a heuristic based on some testing


medakk:
@akenmorris in this branch, could you please add a checkbox to Studio:

Geodesic Distance
Tooltip: Use geodesic distances for sampling term: may be more effective for capturing thin features. Requires ~10x more time, and larger memory footprint. Only supported for mesh inputs

The checkbox value would map to `Optimize::SetGeodesicsEnabled()`

medakk:
@akenmorris Also, could you please add a slider to Options > Mesh Preferences > "Geodesics Cache Size Multiplier". This is a number between 0 and 1000. It defaults to 0, which means the optimizer picks a default value. This would map to `Optimize::SetGeodesicsCacheSizeMultiplier()`

akenmorris:
Should this be a preference or an optimization parameter.  I guess it's more machine specific than dataset specific, so that makes sense.

medakk:
Yeah, it should be machine specific. It's a little weird because its a parameter when run from XML, but I don't know a way around that

medakk:
@akenmorris  
this can be merged now. I've fixed the conflicts and errors from contour domain stuff

Libs/Optimize/ParticleSystem/itkParticleImageDomain.h
```diff
@@ -148,6 +148,7 @@ class ParticleImageDomain : public ParticleRegionDomain
   }
 
   inline PointType GetValidLocationNear(PointType p) const override {
+    // todo why is this function ignoring the argument? Also see Optimize::AddSinglePoint
```
akenmorris, line 151: It seems to be only used to place the first point with:
```

  PointType firstPointPosition;
  firstPointPosition[0] = 0;
  firstPointPosition[1] = 0;
  firstPointPosition[2] = 0;
  firstPointPosition = m_sampler->GetParticleSystem()->GetDomain(0)->GetValidLocationNear(firstPointPosition);
```

We should probably rename these functions based on their current usage.

akenmorris (approved):

Issue #1086, opened 2021-03-03
Deep_ssm and data_augmentation test issues in Windows


iyerkrithika21:
The error is:
```
RuntimeError:
        An attempt has been made to start a new process before the
        current process has finished its bootstrapping phase.        This probably means that you are not using fork to start your
        child processes and you have forgotten to use the proper idiom
        in the main module:            if __name__ == '__main__':
                freeze_support()
                ...        The "freeze_support()" line can be omitted if the program
        is not going to be frozen to produce an executable.
```

Keyboard Interruption to stop the code also didn't work for me.

cchriste:
For the main deep_ssm use case, it continuously displays this at the end(?) of each of ~6900 "generating"s. It's still running after 24 hours, so is it hung in a loop?

cchriste:
The data_augmentation test has the same issue. Also, it uses deep_ssm directory, but might be independent (see #1081).

iyerkrithika21:
> 
> 
> The data_augmentation test has the same issue. Also, it uses deep_ssm directory, but might be independent (see #1081).

Data augmentation and deep_ssm are related so they use the same folder.

jadie1:
I am guessing this is because there is some parallel processing in the data augmentation process that windows doesn't like. 
Could you try changing the `processes` argument from 3 to 1 on line 75 of deep_ssm.py an running it? This will remove the parallelization. You can just try the tiny test.

iyerkrithika21:
> 
> 
> I am guessing this is because there is some parallel processing in the data augmentation process that windows doesn't like.
> @cchriste @iyerkrithika21 Could you try changing the `processes` argument from 3 to 1 on line 75 of deep_ssm.py an running it? This will remove the parallelization. You can just try the tiny test.

This worked out.
```

Step 2. Augment data

Running point based data augmentation.
The PCA modes of particles being retained :  3
Variablity preserved: 1.0
Fitting KDE...
Generating 1/4
Generating 2/4
Generating 3/4
Generating 4/4
Done.
Wrote augmentation_splom.html



Step 3. Reformat Data for Pytorch




********************* WARNING ****************************
Pytorch is running on your CPU!
This will be very slow. If your machine has a GPU,
please reinstall Pytorch to your shapeworks conda
environment with the correct CUDA version.
**********************************************************
```

jadie1:
Okay, I am going to make a PR with this change along with fixes for a couple other bugs I found in DeepSSM

[...]

PR #1109, opened 2021-03-04
Deep ssm fix

jadie1:
This PR addresses issue #1086 and updates two outdated commands in the DeepSSM package.
To test run the deep_ssm tiny test on Windows (should give error for not running on GPU) and on Mac/Linux with GPU (should run entire use case). 

This PR also fixes data_augmentation.py to not use parallelization and to put output in a folder called "data_augmentation" to be consistent with the other use cases (issue#1081).


cchriste:
> To test run the deep_ssm tiny test on Windows (should give error for not running on GPU) and on Mac/Linux with GPU (should run entire use case).

Do you mean warning? Or is this test something that simply doesn't work on Windows?

> This PR also fixes data_augmentation.py to not use parallelization and to put output in a folder called "data_augmentation" to be consistent with the other use cases (issue#1081).

I think this addresses my comment on the issue. Now these two tests could be performed in parallel?

jadie1:
> > To test run the deep_ssm tiny test on Windows (should give error for not running on GPU) and on Mac/Linux with GPU (should run entire use case).
> 
> Do you mean warning? Or is this test something that simply doesn't work on Windows?
> 
It is a warning but it doesn't proceed to run the use case if GPU torch is installed, DeepSSM needs to run on a GPU. Windows users have to install GPU Pytorch to the Shapeworks environments themselves if they have a GPU because there is no automatic way of detecting the cuda version on Windows. 

> > This PR also fixes data_augmentation.py to not use parallelization and to put output in a folder called "data_augmentation" to be consistent with the other use cases (issue#1081).
> 
> I think this addresses my comment on the issue. Now these two tests could be performed in parallel?
Yes!

akenmorris:
See #1118 for fixed PR.

PR #1120, opened 2021-03-05
Updating save_images helper in notebook

archanasri:


akenmorris:
Should I stop the RC7 build, merge in and go to RC8?

archanasri:
The notebook is crashing on Windows according to @cchriste so it might be nice to have a fixed notebook.

PR #1123, opened 2021-03-07
Mesh planes issue

HeavenlyBerserker:
This issue was brought up by Joe Mozingo ([joseph_dataset.zip](https://github.com/SCIInstitute/ShapeWorks/files/6098035/joseph_dataset.zip)). It stems from the integration of mesh domain and cutting plane constraints. In particular, the optimization gets stuck due to the fact that constraints get violated when not being considered by geodesic walks.

I have hence fixed geodesic walks so that it stops (breaks out of while look in GeodesicWalkOnFace) and projects back whenever it hits a constraint. 

These are the results for Joe's dataset.

Screenshot from 2021-03-07 14-59-08

A possible issue is that the point will likely be close but not on the mesh surface, depending on the angle of attack. This might be a slightly temporary fix. The best fix would be to implement a fully constrained geodesic walk that considers intersections between constraints and a certain mesh triangle.

To test:
- Unzip dataset
- Go into the directory
- `shapeworks optimize --name correspondence_1024.xml`


Libs/Optimize/ParticleSystem/VtkMeshWrapper.cpp
```diff
@@ -518,6 +519,18 @@ VtkMeshWrapper::GeodesicWalkOnFace(Eigen::Vector3d point_a, Eigen::Vector3d proj
       remainingVector = RotateVectorToFace(GetFaceNormal(currentFace),
                                            GetFaceNormal(nextFace), remainingVector);
     }
+    // Constraints handling
+    if(constraint->IsAnyViolated(intersect)){
+        vnl_vector_fixed<double, 3> gradE;
+        for(size_t i; i < 3; i ++){
```
medakk, line 525: initialization missing `size_t i = 0;`

Libs/Optimize/ParticleSystem/VtkMeshWrapper.cpp
```diff
@@ -518,6 +519,18 @@ VtkMeshWrapper::GeodesicWalkOnFace(Eigen::Vector3d point_a, Eigen::Vector3d proj
...
+            gradE[i] = intersect[i]-currentPoint[i];
+        }
+        constraint->applyBoundaryConstraints(gradE, currentPoint);
+        for(size_t i; i < 3; i ++){
```
medakk, line 529: initialization missing `size_t i = 0;`

Libs/Optimize/ParticleSystem/itkParticleGradientDescentPositionOptimizer.txx
```diff
@@ -185,6 +188,12 @@ namespace itk
               {
                 m_TimeSteps[dom][k] *= factor;
                 if (gradmag > maxchange) maxchange = gradmag;
+
+                // debugg
+                std::stringstream stream;
+                stream << "###########################################################################\n" << dom << " " << " " << k << " " << oldpoint << " " << newpoint << std::endl;
+                if(!m_ParticleSystem->GetDomain(dom)->GetConstraints()->IsAnyViolated(oldpoint) && m_ParticleSystem->GetDomain(dom)->GetConstraints()->IsAnyViolated(newpoint) ) std::cerr << stream.str();
```
medakk, line 195: we shouldn't have this and any other cout/cerr statements merged in. they clutter the terminal instead of just showing the optimizer progress.

HeavenlyBerserker:
Fixed all of the above. Thanks @medakk

medakk:
Cool, I'll test it out later. Is this PR meant to be merged to master, or the release branch?

HeavenlyBerserker:
Hmm, it's not in the 6.0 release as far as I know. I thought 6.0 was out, so I PRed to master. @akenmorris and @sheryjoe can decide this.

sheryjoe:
This fix should be part of the 6 release. (I should also update the release notes accordingly once this is reviewed and merged).

> A possible issue is that the point will likely be close but not on the mesh surface, depending on the angle of attack. This might be a slightly temporary fix. The best fix would be to implement a fully constrained geodesic walk that considers intersections between constraints and a certain mesh triangle.

This is a concerning issue. Does this mean that with the current fix particles can leave the surface without being able to snap to the surface?

akenmorris:
What's the problem in the original image?  I assume there is a cutting plane restricting the particles at a certain position.  Nothing is jumping out at me as being incorrect.

HeavenlyBerserker:
@akenmorris, it's not visible at all because the scale of the error is tiny, but I agree with @sheryjoe that it is unacceptable. I think I found a solid fix by using the SnapToMesh(PointType pointa, int idx) function as suggested by @sheryjoe. Let me test this. I'll add this later today.

HeavenlyBerserker:
@akenmorris, the scale of the error is very tiny, so it cannot be seen with the naked eye. @sheryjoe, the error will tend to be tiny. We can discuss this in our one-on-one today, as I have a suggestion for a more permanent fix that can be implemented by the end of today.

akenmorris:
@HeavenlyBerserker , should I work on merging your PR or waiting for the other fix?

HeavenlyBerserker:
@akenmorris Please wait

HeavenlyBerserker:
@akenmorris Can you check this today? Shireen and I have decided not to implement a permanent fix as of yet.

Libs/Optimize/ParticleSystem/itkParticleGradientDescentPositionOptimizer.txx
```diff
@@ -164,6 +164,9 @@ namespace itk
               // Step B Constrain the gradient so that the resulting position will not violate any domain constraints
               m_ParticleSystem->GetDomain(dom)->GetConstraints()->applyBoundaryConstraints(gradient, m_ParticleSystem->GetPosition(k, dom));
 
+              // debuggg
+              PointType oldpoint = m_ParticleSystem->GetPositions(dom)->Get(k);
+
```
akenmorris, line 169: Can we remove this debug stuff?

Libs/Optimize/ParticleSystem/itkParticleGradientDescentPositionOptimizer.txx
```diff
@@ -185,6 +188,12 @@ namespace itk
               {
                 m_TimeSteps[dom][k] *= factor;
                 if (gradmag > maxchange) maxchange = gradmag;
+
+                // debugg
+                std::stringstream stream;
+                stream << "###########################################################################\n" << dom << " " << " " << k << " " << oldpoint << " " << newpoint << std::endl;
+                //if(!m_ParticleSystem->GetDomain(dom)->GetConstraints()->IsAnyViolated(oldpoint) && m_ParticleSystem->GetDomain(dom)->GetConstraints()->IsAnyViolated(newpoint) ) std::cerr << stream.str();
+
```
akenmorris, line 196: More string stream stuff that gets executed for every particle on each iteration.  Can we remove this?

Libs/Optimize/ParticleSystem/VtkMeshWrapper.h
```diff
@@ -24,7 +24,7 @@ class VtkMeshWrapper : public MeshWrapper {
 
   double ComputeDistance(PointType pointa, PointType pointb) const override;
 
-  PointType GeodesicWalk(PointType p, int idx, VectorType vector) const override;
+  PointType GeodesicWalk(PointType p, int idx, VectorType vector, const std::shared_ptr<itk::Constraints> &constraint) const override;
```
akenmorris, line 27: Do the constraints change during execution?  Do they need to be passed as a parameter for every walk/particle update?

Also, in my opinion a const reference to a shared_ptr is almost always wrong.  See here for explanation:

https://medium.com/@vgasparyan1995/pass-by-value-vs-pass-by-reference-to-const-c-f8944171e3ce

HeavenlyBerserker: They do not change. You're right that it should not be const reference. I still think passing the pointer is the best choice here, since the point of the constraint class is to only have one place that stores the constraints.

akenmorris: What I mean is that it appears that it's passing in the same pointer every time (per domain).  What if we change MeshWrapper to have a SetConstraints function that takes and holds this pointer and then internally uses it for the geodesic walk?

akenmorris:
I tried this branch on the included dataset, it hangs for me in AdvancedAllParticleSplitting.

HeavenlyBerserker:
> I tried this branch on the included dataset, it hangs for me in AdvancedAllParticleSplitting.

Can you provide this dataset?

akenmorris:
> > I tried this branch on the included dataset, it hangs for me in AdvancedAllParticleSplitting.
> 
> Can you provide this dataset?

Yes, you already linked it though:

https://github.com/SCIInstitute/ShapeWorks/files/6098035/joseph_dataset.zip

I am running on MacOS.

PR #1127, opened 2021-03-09
Violin plot in DeepSSM

jadie1:
Adding violin plot to deep_ssm use case instead of splom. 
Tiny test is sufficient for testing.


akenmorris:
I ran it on windows with --tiny_test, but no charts displayed.  Do they display for --tiny_test mode?

akenmorris:
Screenshot (31)

Great!

sheryjoe:
@jadie1 I wouldn't use underscores in axes labels and legend.

akenmorris (approved):

jadie1:
I will remove the underscores from the violin plot labels and I will also remove the data_augmentation use case. Everything it does is done in the deep_ssm use case and we have the data augmentation notebook which does a better job of documenting and explaining the process. Then this will be ready to merge!

jadie1:
Here is a plot with updated labels from running the tiny test:
image

Ready to merge now!

akenmorris:
This will be a little complicated to merge now that all of the commits from master were merged in.

akenmorris:
I will make a new branch and cherry pick the commits we want.

akenmorris:
See #1144

PR #1132, opened 2021-03-09
Adjust narrow band based on spacing.  No change for 1.0x1.0x1.0 spacing.

### LINKED ISSUE #1134: Larger spacing causes out of narrow band error
@iyerkrithika21  noticed this:

image

The problem is that the spacing (on at least one of the ellipsoids) is 2.0 and with a 4.0 world space narrow band only gets two voxels and runs off.

akenmorris:
This fixes the problem of spacing of 4.0 only getting one voxel of
narrow band.  We now interpret the narrow band as in index space and
use the value times the largest spacing dimension.

I've tested this with spacing 0.0001 up to 1000 and it works fine.

Fixes #1134


medakk (approved):

medakk:
[Docs for narrow_band](https://github.com/SCIInstitute/ShapeWorks/blob/master/docs/workflow/optimize.md) have to be updated to reflect this

akenmorris:
@sheryjoe , this PR makes a small fix and updates the documentation, is that ok?

sheryjoe:
> @sheryjoe , this PR makes a small fix and updates the documentation, is that ok?

yes.

Issue #1112, opened 2021-03-04
notebooks' save_images helper broken on windows 


cchriste:
This is used in getting-started-with-grooming-segmentations and probably others.
- itk doesn't like linux paths
- the shapenames are themselves directories, so just creating outdir isn't enough

It can be fixed by updating save_images helper (eventually this will be added to sw module):
```python
    ...
    for curImg, curName in zip(swImageList, swImageNames):
        filename = Path(outDir + curName + '.' + extension)
        if verbose:
            print('Writing: ' + str(filename))
        #this fixes it (but could be cleaner):
        if not os.path.exists(os.path.dirname(filename)):
        os.makedirs(os.path.dirname(filename))
        curImg.write(str(filename), compressed=compressed) 
```

cchriste:
Didn't quite fix it, but #1141 should do the trick.

akenmorris:
Merged

PR #1141, opened 2021-03-10
Path class must be explicitly converted to string when used as such (#1112)

cchriste:
Very small fix of notebook on windows. Not enough to warrant creating the next RC yet.
Confirmed this works on Windows, but haven't checked other platforms.


akenmorris (approved):
Looks good to me.

Issue #965, opened 2021-02-12
Fix Python Image copy-from-numpy constructor


cchriste:
Our current attempt to _copy_ a numpy array into an image is broken.
`.def(py::init([](py::array_t<long> np_array) {`

This relates to #903 and #716

cchriste:
This is now working from the pybind pull request

sheryjoe:
@cchriste should we close this?

cchriste:
pybind merged with added tests, closing issue

PR #1147, opened 2021-03-11
fix ShapeWorksStudio link and compute mean segmentation differently

cchriste:
(due to #965)


cchriste:
So... linux actions...

akenmorris (approved):

PR #1148, opened 2021-03-12
Generate normals

archanasri:
1. New shapeworks command: GenerateNormals
2. Consolidated commands: FixCuttingPlanes, GenerateFeatureGradientFiles, GenerateNormalFeaFiles


akenmorris (approved):
This all looks good to me.  I would probably recommend that we switch to PEP8 for our Python API though rather than come up with our own non-standard conventions.

Python/shapeworks/ShapeworksPython.cpp
```diff
@@ -593,6 +594,7 @@ PYBIND11_MODULE(shapeworks, m)
   .def("centerOfMass",          &Mesh::centerOfMass, "center of mass of mesh")
   .def("numPoints",             &Mesh::numPoints, "number of points")
   .def("numFaces",              &Mesh::numFaces, "number of faces")
+  .def("getPoint",              &Mesh::getPoint, "value at given point", "p"_a)
```
medakk, line 597: Its not really clear what this means. "return (x,y,z) coordinates of the vertex at a given index" perhaps?

medakk (approved):

Python/shapeworks/ShapeworksPython.cpp
archanasri: Updated documentation

akenmorris:
What is this generateNormals used for?  I noticed that splitting is on, so this may be problematic for some uses (but desired for others).  Also, I saw that we're also using vtkPolyDataNormals with splitting off in Mesh::fillHoles.

cchriste:
Sorry, but I'm not 

> This all looks good to me. I would probably recommend that we switch to PEP8 for our Python API though rather than come up with our own non-standard conventions.

Can you please provide a specific example?

cchriste:
> What is this generateNormals used for? I noticed that splitting is on, so this may be problematic for some uses (but desired for others). Also, I saw that we're also using vtkPolyDataNormals with splitting off in Mesh::fillHoles.

We discussed this during last week's meeting, and the two changes that were made after that were to use ->SplittingOff(); for the normal computation filter and call that function in Mesh::fillHoles() rather than using the filter directly (less cut and paste). These were both finished by Archana.

cchriste:
The biggest concern I have is this pull request merges release_6.0c into master. Is that what we want to do? 
I thought 6.0 was all about meshes, which would reasonably include this function.

cchriste (commented):
Wait until release is merged back to master, then merge this branch into master.

akenmorris:
> Can you please provide a specific example?

Sure, a specific example would be something like:

"Function names should be lowercase, with words separated by underscores as necessary to improve readability."

https://www.python.org/dev/peps/pep-0008/#naming-conventions

This means that "generateNormals" should be "generate_normals".

More details are at #1149

Issue #1125, opened 2021-03-08
Clip meshes to apply cutting plane constraints


HeavenlyBerserker:
This was your idea, @medakk.

This has been selected as a temporary, more principled fix to what I implemented in #1123.

I can implement this unless you want to.

medakk:
You can go ahead and implement it.

Which version is going into the release? Does  PR  #1123 still need to be tested?

sheryjoe:
Please keep in mind this is not a full-fledged fix. It won't work if given multiple constraints per sample that result in disjoint meshes/compartments.

HeavenlyBerserker:
@akenmorris or @medakk , I've successfully clipped the meshes in the constructor for VtkMeshWrapper, but even though the cutting planes are the same as in the constraints class, the place where it cuts on the femur differs. Any ideas why this might be the case? Is the coordinate system during the optimization different from the one in the parameter reading steps?

![Screenshot from 2021-03-11 13-51-01](https://user-images.githubusercontent.com/11202328/110854729-9d304100-8272-11eb-941b-69b327d24000.png)
Fig. 1 This image shows a successful clipping, but the clip happens higher on the femur than expected.

akenmorris:
They should be the same.

I would suggest having the VtkMeshWrapper print out the bounds that it computes and compare to your cutting plane.

PR #1150, opened 2021-03-14
Cut mesh

HeavenlyBerserker:
This PR deals with Joe Mozingo's issue in a more principled and clean way by clipping the mesh before the optimization (#1125). The results look pretty good.

As before, the test dataset used was [joseph_dataset.zip](https://github.com/SCIInstitute/ShapeWorks/files/6137211/joseph_dataset.zip), and can be run by 
`shapeworks optimize --name correspondence_1024.xml`

Screenshot from 2021-03-14 11-38-36
Fig. 1 Shows the results on Joe's dataset

Screenshot from 2021-03-14 11-42-47
Fig. 2 Shows a visualization of the clipping in Paraview. The red wireframe is the original shape and the white surface is the clipped mesh.


medakk (changes requested):

akenmorris (changes requested):
This looks like it will solve the problem.  Great Job.

But I think we should move the I/O and cutting plane stuff out of VtkMeshWrapper.  Seems like the polydata should just be already cut before being provided to this constructor.  In that respect, I don't think VtkMeshWrapper needs to be touched at all for this change.

Also, we already have an implementation of cutting meshes with clipping planes using vtkClipPolyData in Mesh.cpp (Mesh::clip).  Can we re-use this or is this too different?

HeavenlyBerserker:
@akenmorris Oh, I didn't realize Mesh::clip existed. The mechanism is the same, but I don't want to create a mesh object unnecessarily. I think the easiest way here is to just move the clipping from VtkMeshWrapper::VtkMeshWrapper to where it's constructed at OptimizeParameterFile::read_mesh_inputs and OptimizeParameters::set_up_optimize. Would this be acceptable?

akenmorris:
There's not really any cost to constructing a Mesh class, and we are actually doing it already:

https://github.com/SCIInstitute/ShapeWorks/blob/58ffac86cbea1e7f0b4ede9ff6ded167bd5dfc14/Libs/Optimize/OptimizeParameters.cpp#L281

If you just clip before it calls ::getVTKMesh, then it won't even be creating a new one.

akenmorris:
Looks great. Thanks @HeavenlyBerserker .  I will just commit a style/cleanup commit and then merge it.

akenmorris (approved):

PR #1156, opened 2021-03-16
Fix build_dependencies.sh return code

### LINKED ISSUE #1152: GitHub actions do not stop at "Build Dependencies" failure
If build dependencies fails for whatever reasons, it returns an exit code 1 and exits the script. However, our GitHub actions workflow seems to consider this a "success" and moves on to the next step. This is on atleast windows, I'm not sure of other platforms.

See https://github.com/SCIInstitute/ShapeWorks/runs/2107766692 for an example:

image
Build Dependencies succeeded even though the ITK install failed.

It eventually fails at the later cmake step with a message about FindITK failing.

akenmorris:
This should resolve #1152

See: https://www.shellscript.sh/tips/pipestatus/

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -74,7 +74,7 @@ jobs:
         key: QtCache-linux
 
     - name: Install Qt
-      uses: jurplel/install-qt-action@v2.11.1
+      uses: jurplel/install-qt-action@v2.13.0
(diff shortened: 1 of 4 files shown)


medakk (approved):
I see, we were looking at the exit status of "tee"!

akenmorris:
Works!

Issue #1104, opened 2021-03-04
ellipsoid_cut model shows poor correspondence


akenmorris:
Here's what I see at the end of ellipsoid_cut.  Notice there are swapped particles, etc.

<img width="1152" alt="ellipsoid_cut" src="https://user-images.githubusercontent.com/1693349/109993000-135c0300-7cc9-11eb-9cb8-f9903c44e2d9.png">

https://user-images.githubusercontent.com/1693349/109993135-35ee1c00-7cc9-11eb-9b34-6817d70a3c68.mp4

jadie1:
Will be fixed for release via issue #1133

jadie1:
@HeavenlyBerserker After updating the ellipsoid data to only have one mode of variation I am still seeing bad correspondence. The shape model correctly predicts only one mode of variation but some of the particles are crossing over each other. 
Can you look at the optimization parameters? The new dataset is called "ellipsoid_1mode_aligned". I have changed the ellipsoid_cut use case to use it on the "one_mode_ellipsoid" branch.

HeavenlyBerserker:
@jadie1 I'll look at it on Thursday. Thanks for the note.

sheryjoe:
@HeavenlyBerserker @jadie1 is this issue resolved?

HeavenlyBerserker:
To the best of my knowledge, yes.

jadie1:
That was with 128 particles though. I will test with the new params on the one_mode_ellipsoid branch which uses the new data. If that looks good then we can close this

jadie1:
![image](https://user-images.githubusercontent.com/54382613/111656553-407fd980-87d0-11eb-9d60-a3a4656ccadb.png)
Looks good, packing isn't super uniform but no particles are switching. Closing this now.

PR #1157, opened 2021-03-16
Tuned parameters to fix the ellipsoid_cut use case

HeavenlyBerserker:
This PR fixes the parameters of the ellipsoid_cut use case so that the correspondence is reasonable. I decreased the number of particles because the clipped ellipsoid, which had less than 1/4 of the surface area of the whole ellipsoid was using the same number of particles (128 -> 32). I also increased the number of iterations so that the particles have the chance to swap in the cramped space of particles. This fixes issue #1104. The results are shown in fig 1

Screenshot from 2021-03-16 14-24-13
Fig. 1 Shows the result of running the use case with the new parameters

To run this use case
1. cd into ShapeWorks/Examples/Python
2. `python RunUseCase.py --use_case ellipsoid_cut`

### DIFF
--- a/Examples/Python/ellipsoid_cut.py
+++ b/Examples/Python/ellipsoid_cut.py
@@ -87,13 +87,13 @@ def Run_Pipeline(args):
         cutting_plane_counts.append(2)
 
     parameterDictionary = {
-        "number_of_particles": 128,
+        "number_of_particles": 32,
         "use_normals": 1,
         "normal_weight": 15.0,
         "checkpointing_interval": 200,
         "keep_checkpoints": 0,
-        "iterations_per_split": 2000,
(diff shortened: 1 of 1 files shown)


jadie1:
The ellipsoid_cut use case should use the dataset "ellipsoid_1mode_aligned" now. We are going to remove "ellipsoid_aligned-v1". Also Shireen suggested keeping the particle count at 128 for more uniform packing. I will try with more iterations on the new data.

sheryjoe:
> Also Shireen suggested keeping the particle count at 128 for more uniform packing. 

@jadie1 Per to a discussion with @HeavenlyBerserker and @medakk it is reasonable to keep the particle could around 1/4 of the particle count used in the ellispoid use case.

PR #1167, opened 2021-03-23
fix getting started with grooming notebook on windows

cchriste:
launch studio using list of files since on windows it doesn't support wildcard arguments


cchriste:
Oops, didn't clean up the notebook (it shows no changes on my machine so I thought it was cleaned). Doing that now.

cchriste:
Apparently git uses something called nbstripout to only show changes to the calls, not the output.

cchriste:
branch name has bad special characters, renaming to fix_notebook

PR #1178, opened 2021-03-24
fix Mesh::toImage

### LINKED ISSUE #1175: mesh.toImage(spacing) gives std::bad_alloc (femur_cut use case)
When I run the femur_cut example and answer "n" to the isotropic spacing question.  It asks for spacing and it seems that unless I give it 1/1/1, I get:

```
  File "RunUseCase.py", line 94, in <module>
    module.Run_Pipeline(args)
  File "/Users/amorris/sci/data/test/Python/femur_cut.py", line 242, in Run_Pipeline
    fileList_seg = MeshesToVolumes(parentDir + "volumes", reflectedFiles_mesh, spacing)
  File "/Users/amorris/sci/data/test/Python/GroomUtils.py", line 451, in MeshesToVolumes
    image = mesh.toImage(spacing)
MemoryError: std::bad_alloc
```

I've tried a lot of different spacings and always seem to get this unless it's 1.0/1.0/1.0.  Some examples

2.0 2.0 2.0
0.5 0.5 0.5
1.1 1.1 1.1
1 2 3

cchriste:
Fix Mesh::toImage so it no longer crashes when custom spacing or dims are passed (#1175)
- cleaned up and simplified implementation, removing extraneous functions
  - also removed commands and tests accidentally added for these functions 
  - note they were already removed in the generate_normals branch; merge conflicts ahoy
- more robust implementation
  - tests that spacing is non-zero when computing rasterization size
  - ensures that only dims or spacing are specified, but not both
- as before, pads created image out by two instead of one, and set its origin to the corner of the closest pixel instead of the center
- to correctly compute new size if explicit spacing passed to Mesh::toImage, can only add that extra padding when dims aren't specified, which coincidentally makes antialiasing look much fluffier
- updated all unit tests

femur_cut tinytest still passes, as do all the unit tests

Resolves #1175


akenmorris:
Looks good. I will test it out.

akenmorris:
Works for me!

PR #1185, opened 2021-03-30
Merge release 6

akenmorris:
Merge 6.0 release into master.  Take a look at the Mesh stuff and see if I merged it right.  I'm more confident of the other conflicts.


archanasri:
Looks good to me

archanasri (approved):

cchriste (changes requested):
The contents of #1180 are missing from GroomUtils.py:
```
+++ b/Examples/Python/GroomUtils.py
@@ -121,7 +121,9 @@ def FindReferenceImage(inDataList):
     """
     x = y = z = 0
     for i in range(len(inDataList)):
-        dim = itk.GetArrayFromImage(itk.imread(inDataList[i])).shape
+        img = Image(inDataList[i])
+        tmp = img.toArray()
+        dim = tmp.shape
```

Examples/Python/setupenv.py
```diff
@@ -29,22 +29,17 @@ def setup_shapeworks_env(shapeworks_bin_dir = None,   # path to the binary direc
...
-        sys.path.append(shapeworks_bin_dir + "/ShapeWorksStudio.app/Contents/MacOS")
     
-    # add shapeworks and studio to the system path
-    os.environ["PATH"] = shapeworks_bin_dir   + os.pathsep + os.environ["PATH"]
+    # add shapeworks to the system path
+    os.environ["PATH"] = shapeworks_bin_dir + os.pathsep + os.environ["PATH"]
     if dependencies_bin_dir is not None:
         os.environ["PATH"] = dependencies_bin_dir + os.pathsep + os.environ["PATH"]
-    if platform.system() == "Darwin": # MacOS
-        os.environ["PATH"] = shapeworks_bin_dir + "/ShapeWorksStudio.app/Contents/MacOS" + os.pathsep + os.environ["PATH"]
```
cchriste, line 45: Is this no longer necessary or was it moved to another place to add studio to the path on OSX?

cchriste (commented):
there are conflicting changes in conda_installs.sh:

docs/index.md
```diff
@@ -42,6 +45,10 @@ Its optimized approach offers increased power for testing biological hypotheses
 
 The underlying scientific premise of ShapeWorks, particle-based shape modeling, is a groupwise approach to placing landmarks (i.e., correspondences) that consider variability in the entire cohort of images. 
 
+*ShapeWorks 5.5 tour - feature support in ShapeWorksStudio*
```
cchriste, line 48: Should this be 6.0?

cchriste (commented):
I've looked closely at most everything, but I'm not too familiar with Studio, the ellipsoid use cases, and optimize. I pushed the missing changes from pr #1180. Please look at my other comments just in case I noticed anything important.

akenmorris:
> there are conflicting changes in conda_installs.sh:
> 
> ```
> <<<<<<< HEAD
>     pkg-config=0.29.2
> =======
>     pkgconfig=1.5.2
> >>>>>>> release_v6.0c
> ```
> 
> I can't remember the story, but is it appropriate for us to keep pkg-config and just dump pkgconfig? Just want to make sure.

The python pkgconfig was mistakenly added.  I had meant to add pkg-config since we were seeing a cmake message about how it couldn't find it on the windows runner.  I don't think that was the problem after all.  

I'm not convinced we need either of them.  Do you want to hold off merging this until we test without?

Examples/Python/setupenv.py
akenmorris: We don't need this anymore, correct.  We now have the ShapeWorksStudio shell script sitting in the bin directory, so it works the same as on Linux.

akenmorris: I'm not sure what you mean about the "regular build"?

cchriste: I looked in my bin after building and didn't see it, but it's there. 👍

cchriste:
> I'm not convinced we need either of them. Do you want to hold off merging this until we test without?

No, let's merge.

PR #1181, opened 2021-03-26
pymodule

archanasri:
This branch needs testing:
1. Check if the notebooks are producing correct results
2. Check if any cell or combinations of cells can be just added as helper function or removed
3. Read through markdown text to see if it can be removed or paraphrased to reduce verbosity

PR #1200, opened 2021-04-10
Pybind

cchriste:
Updated python bindings that get rid of all the opaque structs (Coord, Dims, Point, Vector, Matrix, Index)  in favor of simply using numpy and/or standard python lists and std::vector. The transparency and simplicity are much more python-esque.


cchriste:
I added Jadie and Krithika to ensure the (existing) Jupyter notebooks still work, but otherwise all use cases and unit tests succeed. Getting this pull request merged with master (and then back down to the python module branches) will make the rest of the python work more straightforward.

Python/shapeworks/BC.cpp
```diff
@@ -0,0 +1,21 @@
+#include <pybind11/pybind11.h>
```
archanasri, line 1: We probably don't need to add this file

cchriste: Removing now, thanks

archanasri:
The bug mentioned by @jadie1 in #1181 exists in this PR as well. It has been fixed so the reviewers can continue testing.

jadie1:
The mesh code updates have made it so the shape cohort generation code doesn't work. This can be seen by running the getting-started-with-shape-cohort-generation.ipynb notebook.

cchriste:
> The mesh code updates have made it so the shape cohort generation code doesn't work. This can be seen by running the getting-started-with-shape-cohort-generation.ipynb notebook.
> 
> In particular the generate_segmentations() function in [CohortGenUtils.py](https://github.com/SCIInstitute/ShapeWorks/blob/pybind/Python/ShapeCohortGenPackage/ShapeCohortGen/CohortGenUtils.py) does not work because the mesh.toImage() function has changed.

Thanks, @jadie1. I think we fixed this working together and then the commit didn't make it back to the source tree. It's great to have another set of eyes on these notebooks. What we changed was to make the Mesh.toImage function more clear in its request for the specific region of space to convert to an image of a specified resolution. These were determined indirectly in the previous version and the usage wasn't as clear.

jadie1:
Update on testing:
- toImages() looks like it is working
- Pyvista has moved to version 30.1 which broke the notebooks, so I will work on updating that now (issue #1256 )

jadie1:
I upgraded Pyvista. To test you either need to rerun conda installs or run:
```
pip install --upgrade pyvista
pip install ipyvtklink
```

jadie1 (approved):
The notebooks and use cases ran as expected for me

Libs/Common/Shapeworks.cpp
```diff
@@ -13,9 +13,13 @@ Coord toCoord(const Dims &d) {
                 static_cast<itk::IndexValueType>(d[1]),
                 static_cast<itk::IndexValueType>(d[2])}); }
 Dims toDims(const Coord &c) {
-  return Dims({static_cast<itk::SizeValueType>(c[0]),
-               static_cast<itk::SizeValueType>(c[1]),
-               static_cast<itk::SizeValueType>(c[2])}); }
+  return Dims({static_cast<Dims::value_type>(c[0]),
```
cchriste, line 16: Dims is itk::Size, which is unsigned, so let's make sure these are std::max(0,arg) when converted from both Coord and Point.

Libs/Common/Shapeworks.h
```diff
@@ -133,7 +133,7 @@ P operator-(const P &p, const P &q)
 template<typename P, typename = std::enable_if_t<std::is_same<Image, P>::value ||
                                                  std::is_same<Coord, P>::value ||
                                                  std::is_same<Dims, P>::value ||
-                                                 std::is_same<Vector, P>::value ||
+                                                 std::is_same<Vector, P>::value || // use operator*(v0, v1); (or call dotProduct)
```
cchriste, line 136: This isn't super clear, though I know what is meant: `v0 * v1` won't call this function, so if you want to use it, you have to call `operator*(v0, v1)` or just use the `dotProduct` function. Clarification or suggestions are welcome.

Testing/PythonTests/toImage.py
```diff
@@ -13,11 +13,13 @@ def toImageTest1():
 val = toImageTest1()
 
 if val is False:
+  print("toImageTest1 failed")
   sys.exit(1)
 
 def toImageTest2():
   mesh = Mesh(os.environ["DATA"] + "/femur.ply")
-  img = mesh.toImage(spacing=[2.0, 2.0, 1.0])
+  bbox = mesh.boundingBox() # changing spacing means we need to pad more to compensate
```
cchriste, line 21: Unfortunately... I wonder if we could interpret this padding argument in the to be image space since the user very likely wants the image in the end to have a logical padding added.

Testing/shapeworksTests/boundingboximage.sh
```diff
@@ -1,3 +1,26 @@
...
+	max: [42, 34, 34]
+}")
+if [[ $? != 0 ]]; then echo "bbox test 1 failed"; exit -1; fi
+diff --strip-trailing-cr <(shapeworks readimage --name $DATA/femurImage.nrrd image-bounds) <(echo "Bounding box:
+{
+	min: [46.4821, -192.471, -737.593],
+	max: [134.482, -129.471, -609.593]
+}")
+if [[ $? != 0 ]]; then echo "bbox test 2 failed"; exit -1; fi
+diff --strip-trailing-cr <(shapeworks readimage --name $DATA/femurImage.nrrd image-bounds --isovalue 1) <(echo "Bounding box:
```
cchriste, line 15: multiplatform friendly diff. very nice!

Testing/shapeworksTests/crop.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-shapeworks readimage --name $DATA/1x2x2.nrrd crop --xmin 25 --xmax 50 --ymin 1 --ymax 40 --zmin 12 --zmax 23 compareimage --name $DATA/cropped1.nrrd
+shapeworks readimage --name $DATA/1x2x2.nrrd imageinfo set-region --xmin 25 --xmax 49 --ymin 2.5 --ymax 78.5 --zmin 24.5 --zmax 44.5 crop writeimage --name "/tmp/yay.nrrd" compareimage --name $DATA/cropped1.nrrd
```
cchriste, line 3: need to get rid of the temporary file write and the image info call

cchriste (commented):
Wow! I made it through the whole thing. Couple comments in here for changes, some of which have already been made in the pybind_work branch.

Here is a summary of the changes in this branch, including a couple very small fixes to be made:
```
- added new commands: ImageBounds, MeshBounds, SetRegion
- made it possible to ensure an option passed to command line is valid
- update several commands:
 - let several commands fail themselves instead of double-checking params in command, including...
  - antialias, scale, rotate, reflect
 - improved, simplified, and clarified additional commands, including...
  - mesh-to-image, image-info (show bounds), clip mesh,  
 - remove padding option from boundingbox-image since this can be done with the pad command
  - used by both Mesh and Image, but unclear how to pad
  - the resultant image can be padded using the pad command
   [x] add example of doing this somewhere (do we have examples of using the command line in the wiki docs?)
   - note: Mesh::toImage and python binding accept padding argument, but simpler from command line to omit this
- updated GroomUtils with improved python bindings
 - added comments to clarify what is going
- added makePlane functions from three pts or pt and normal
 - used for mesh and image clipping
- upgraded to latest pyvisa for notebooks (thanks, Jadie)
[x] toDims(Point) should also std::max(0, arg) (in Shapeworks.cpp) since Dims have to be positive.
[x] same with toDims(Coord)
- more robust functions in both Image and Mesh that check for validity and throw if not (e.g., Image::antialias), or just no longer permit invalid calls to be made (e.g., Mesh
 - Image::antialias, Image::createTransform, Image::applyCurvatureFilter, Image::crop
   Image::crop (_much_ safer and more accurate), Image::reflect, Image::pad,
   Mesh::toImage (so much cleaner, no longer accept conflicting parameters, and more robust, go us!),
   Mesh::toDistanceTransform, Mesh stream insertion operator, 
- split Region into both Physical and Index (0-based) Regions, using best ones where appropriate
 - changed *Region::grow to expand to make it more clear what is happening and be more consisted with others
- ability to request both logical (0-based) and physical (origin-based) bounding box for an image
- can convert between physical and logical regions in Image
  (of course you lose accuracy if you try to go the other way)
- Mesh::boundingBox always returns physical since there is no such thing as logical (index) mesh space
- added EigenUtils to convert between itk (boo) and Eigen (yay) matrices
 - used for passing matrices such as Image coordinate system back and forth between Python
- CohortGenUtils now uses shapeworks Image and Mesh classes instead of ITK
 - direct calls rather than system subprocesses
 - initialization().and().chaining() for elegance and potential confusion
- More than 20 Improved and added Image and Mesh tests
- Python bindings additions and improvements (thus the branch name)
 - Image initialization from and export to numpy arrays (dtype.float32, same as Image)
 - Simplified testing using direct output and exceptions (some still left to go, but not urgent for 6.1)
 - added and improved tests for dozens of functions
 - consolidated all *fail tests to same scripts as their complementary functions
 - print which test failed for scripts with multiple tests (most of them)
 - not binding Plane for python; instead simply using three-pt or pt plus origin
  - Mesh.clip and Image.clip
 - ability to modify *Region.min/max by index
  - paves the way for sharing data between C++ and Python without copying
  - explanations and clarity added to notes in open issues #966 and ...
 - removed all classes inflexibly bound that could simply use python/numpy arrays
  - Coord, Dims, *Point, Vector, Matrix, TransformPtr
  - using c++ lambdas instead to maintain natural python interface
 - removed excessive output from optimize test
 - fixed chaining of functions
 - use Eigen matrices (a native c++/python library) for Transforms
 - much more readable source for bindings
- updated to latest pyvista
- notebook cleanup
```

Libs/Common/Shapeworks.cpp
cchriste: fixed

Testing/PythonTests/toImage.py
cchriste: Removed padding from these two function. If user wants a larger region of the mesh they can make the region larger (added a function to PhysicalRegion to do this). If they want a number of pixels on the boundary they can call Image::pad on the result. Updated all tests and examples and Python bindings.

Testing/shapeworksTests/crop.sh
cchriste: done

Libs/Common/Shapeworks.h
cchriste: This is definitely not a reason to postpone a merge. We only added a comment to the `enable_if`.

Libs/Image/Image.cpp
```diff
@@ -726,16 +753,18 @@ Image& Image::gaussianBlur(double sigma)
   return *this;
 }
 
-Image& Image::crop(const Region &region)
+Image& Image::crop(PhysicalRegion region, const int padding)
 {
+  region.shrink(physicalBoundingBox()); // clip region to fit inside image
   if (!region.valid())
-    std::cerr << "Invalid region specified." << std::endl;
+    std::cerr << "Invalid region specified (it may lie outside physical bounds of image)." << std::endl;
```
akenmorris, line 760: Does this need to return?  Or throw an exception?  What happens in the ITK filter below if the region is invalid?

akenmorris (approved):

Libs/Image/Image.cpp
cchriste: The itk filter takes a zero-based index region, but we wanted the user to consider the physical region, which is consistent with Mesh. If we decided to also have an index region we can add this at any time.

Issue #837, opened 2020-11-30
Implement Geometric Primitives as augmented lagrangian inequality constraints


HeavenlyBerserker:
We aim to bake sphere and cutting plane constraints into the energy function as inequalities g(x) within an augmented lagragian inequality equation.

PR #1202, opened 2021-04-12
Use augmented lagrangian to apply cutting plane and sphere constraints

HeavenlyBerserker:
Handles issue #837.
Also handles #1211.

The augmented lagrangian method is a way to formulate a constrained optimization problem into an unconstrained one by adding terms to the cost function. In particular, augmented lagrangian uses a linear and an L2 method to serve as a repelling force when any particle violates a constraint. This is superior to the log barrier method which is numerically unstable. Unlike the regular lagrangian method, the augmented lagrangian adjusts lambda automatically, exposing no hyperparameters to the user. There are many improvements thanks to the augmented lagrangian:

1. Particle splitting and constraints will no longer require that no particle violates constraints. It will work even if a particle violates a constraint.
2. Should allow multiple constraints to be applied simultaneously, even if they are of different types.
3. It should be numerically stable.
4. As before, it does not expose any parameters to the user.

The only disadvantage is that the ApplyConstraints function may cause some particles to violate some constraints by a fraction of a unit (<0.35). This should never cause any major problems, however. This approach replaces the gradient projection method, which is problematic in general, with the only benefit being that it guarantees no constraint is violated. To access the old gradient projection method, look to release 6.0 and this PR, where it was removed.

Apr 17th update: Now this PR includes sphere and sphere-cutting_plane functionality. This includes an updated cutting plane unit test that evaluates the particles themselves instead of the means of particles over domains. Additionally, two unit tests have been added. 

-  The sphere_constraint unit test with 25 spheres per domain. The first sphere cuts out a section atop the sphere, while the remaining 24 create a cutting plane that splits the sphere in half.
- The sphere_cutting_plane_constraint unit test is the same as the sphere_constraint, but with an added cutting plane that again splits the half-sphere in half.

Results on new cutting_planes unit test with 3 cutting planes
Screenshot from 2021-04-12 13-55-59
Results on new sphere_constraint unit test with 25 spheres
Screenshot from 2021-04-17 15-39-26
Results on new sphere_cutting_plane_constraint unit test with 25 spheres and one cutting plane
Screenshot from 2021-04-17 12-54-58
Results on ellipsoid_cut use case
Screenshot from 2021-04-12 09-30-27
Results on femur_cut use case
Screenshot from 2021-04-06 18-00-03

For basic testing:
- [ ] Run the unit test and see if `ShapeWorksStudio analyze.xml` on Testing/data/cutting_plane_multi looks good.
- [ ] See if `ShapeWorksStudio analyze.xml` on Testing/data/sphere_constraint looks good.
- [ ] See if `ShapeWorksStudio analyze.xml` on Testing/data/sphere_cutting_plane_constraint looks good.
- [ ] Run the ellipsoid_cut use case.
- [ ] Run the femur_cut use case.

Would also be nice if you have another dataset to test on.

For more information on augmented lagrangian inequality constraints, please see [http://www.cs.cmu.edu/~pradeepr/convexopt/Lecture_Slides/Augmented-lagrangian.pdf](url).


Libs/Optimize/ParticleSystem/itkParticleGradientDescentPositionOptimizer.txx
```diff
@@ -162,10 +162,36 @@ namespace itk
               VectorType gradient = original_gradient_projectedOntoTangentSpace * m_TimeSteps[dom][k];
 
               // Step B Constrain the gradient so that the resulting position will not violate any domain constraints
-              m_ParticleSystem->GetDomain(dom)->GetConstraints()->applyBoundaryConstraints(gradient, m_ParticleSystem->GetPosition(k, dom));
+              //m_ParticleSystem->GetDomain(dom)->GetConstraints()->applyBoundaryConstraints(gradient, m_ParticleSystem->GetPosition(k, dom));
 
               gradmag = gradient.magnitude();
 
+              // Step B 1: if the magnitude is larger than the Sampler allows, scale the gradient down to an acceptable magnitude
```
akenmorris, line 169: This seems like a lot of lines to add to the main body of the gradient descent optimizer, can this new section be placed in a separate function?  Also, should there be a flag that enables or disables this, or does this all need to be run even with no constraints?

PR #1212, opened 2021-04-17
Warp landmarks[WIP]

riddhishb:
This is a standalone tool for performing mesh reconstruction using mesh warping. 
The reason for the WIP tag is due to a mesh/particle data I have that is causing a seg fault in the reconstruction code. It works well with other datasets.

@akenmorris @medakk Can you guys test this and see if it is causing the same problems on your end?

The data is here
[data.zip](https://github.com/SCIInstitute/ShapeWorks/files/6329992/data.zip)

The way to run this is 
```
./WarpLandmarks templatemeshpath templatepointpath outputmeshpath outputpointpath
```


akenmorris:
I need to combine this into a consolidated command, use Studio's MeshWarper.  The data you attached works after the fixes in PR #1217 .

akenmorris:
Also @riddhishb , while I have fixed this from crashing, the resulting mesh has lots of problems, but that's because the reference -> target particles are trying to rip it apart:

image

akenmorris:
@riddhishb See #1217, I think we can close this PR as #1217 adds a `shapeworks` command to do this.

riddhishb:
Thanks @akenmorris I think we can close this

PR #1217, opened 2021-04-22
Fix mesh warp

akenmorris:
This PR fixes more mesh warping problems.

I found that when the add_particles method added vertices that were very close to other vertices, this caused NaNs in the build meshes.  So we now do an epsilon check based on the barycentric coordinates.

This required a different strategy on the good/bad particles because two particles may not have identical positions, but could epsilon resolve to the same vertex.

I also fixed the case where the new vertex lands exactly on an edge.  In this case, we split the current triangle into two and have to split the neighbor triangle into two as well, if there is one (e.g. unless boundary).

This PR also now checks the return status of the ign biharmonic call.  So the API has changed to pass the warp (W) as a reference and return bool for success.

Update (04/27/21):  This PR now adds the warp mesh command:

`$ shapeworks warp-mesh --reference_mesh=template.ply --reference_points=template.particles --target_points=output.particles --target_mesh=output.vtk`

This also pulls the MeshWarper class out of Studio and moves the MeshUtils functionality into it.

Automate tests are updated.


Studio/src/Application/Data/MeshWarper.cpp
```diff
@@ -216,6 +303,64 @@ Eigen::MatrixXd MeshWarper::remove_bad_particles(const Eigen::MatrixXd& particle
...
+
+    list->SetId(0, new_vertex);
+    list->SetId(1, cell->GetPointId(2));
+    list->SetId(2, cell->GetPointId(0));
+    this->reference_mesh_->InsertNextCell(VTK_TRIANGLE, list);
+  }
+  else if (edge == 1) {
+    list->SetId(0, new_vertex);
+    list->SetId(1, cell->GetPointId(0));
+    list->SetId(2, cell->GetPointId(1));
```
medakk, line 340: should it be GetPointId(2)?

(also, is it possible to use modulo arithmetic in this section? maybe something like GetPointId((edge+1)%3) and GetPointId((edge+2)%3) but with the correct ordering)

akenmorris: There are 3 cases:

image

It can definitely be simplified since two triangles are repeated in each case.  It can be change to a **not** and only one triangle per section.

Are you sure that it's wrong?

medakk: Oh I see,  I  messed up the edge conventions. I thought edge==1 implies that the edge is opposite to vertex 1

medakk (approved):

Studio/src/Application/Data/MeshWarper.cpp
akenmorris: Is that the preferred convention?  If so, we should probably change to conform.

medakk: I don't know if there is a preferred convention

riddhishb (approved):
looks good to me

PR #1226, opened 2021-04-28
Change procrustes to use positions after prefix transform.

### LINKED ISSUE #1225: Procrustes ignores prefix transforms
Using procrustes with prefix transforms results in incorrect correspondence models:

This is because the procrustes registration is ignoring the prefix transforms.

https://user-images.githubusercontent.com/1693349/116468768-80ee6080-a82e-11eb-817f-fb79e2a3715d.mp4

akenmorris:
This fixes #1225.

After this change:

https://user-images.githubusercontent.com/1693349/116474590-f01b8300-a835-11eb-9129-10b701221b25.mp4

### DIFF
--- a/Libs/Optimize/ParticleSystem/itkParticleProcrustesRegistration.cxx
+++ b/Libs/Optimize/ParticleSystem/itkParticleProcrustesRegistration.cxx
@@ -41,9 +41,9 @@ ParticleProcrustesRegistration<3>::RunRegistration(int d)
         shapevector.clear();
         for(int j = 0; j < numPoints; j++)
         {
-            point(0) = m_ParticleSystem->GetPosition(j,i)[0];
-            point(1) = m_ParticleSystem->GetPosition(j,i)[1];
-            point(2) = m_ParticleSystem->GetPosition(j,i)[2];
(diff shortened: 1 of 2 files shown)


medakk:
What is "prefix transform" for? Why not always pass world coords to the optimizer?

akenmorris:
In Studio, I’ve switched from transforming the data (e.g. centering) to passing the transform as the “prefix transform”.  There are several advantages to this approach. For example, if we perform an ICP registration, we won’t suffer from resampling a distance transform at a slight rotation. This also makes it so local particle positions are in the original space (e.g. CT or MRI). 

Internally, the optimizer can perform procrustes registration at regular intervals. Prior to this PR, the procrustes step would discard the prefix transform and register using the local (original) coordinates. It seems incorrect to me to discard the transforms given. 

Furthermore the GetTransformedPosition call multiples the local coordinate by the prefix transform and the procrustes transform. This is clearly incorrect because the procrustes transform will likely be duplicating the prefix transform in most cases. This is why the model is wrong in #1225.

medakk:
I see, makes sense. Whats in the "transform" file that shapeworks finally emits?

medakk (approved):

akenmorris:
I’m pretty sure this is the procrustes transforms for each domain, concatenated.

PR #1228, opened 2021-05-07
fix osx Big Sur dependencies xlnt build and upgrade it to version 1.5.0

cchriste:
Failed due to a file called `version` in one of the xlnt third party libs. Fixed by temporarily moving that file out of the way to build the library then back afterwards (in case it's needed for anything, which seems unlikely.


akenmorris (approved):
👍

Issue #1229, opened 2021-05-07
Studio: Mesh Warping needs a progressbar


akenmorris:
With all of the fixes, the mesh warping is significantly slower, depending on the number of particles. This is due to iteratively adding vertices and splitting triangles.  The biggest cost is probably that the locator and links have to be rebuilt each time as we don't have a way of only incrementally changing them inside of VTK.

medakk:
>  The biggest cost is probably that the locator and links have to be rebuilt each time

It _may_ be faster to check distance to every triangle and pick the minimum vs rebuilding the (kd)tree for every query. O(n) vs O(n log n) where n is the number of triangles. But its nice to have VTK take care of all the details of the query

PR #1230, opened 2021-05-07
Studio mesh warp progress

akenmorris:
This PR addresses #1229 by adding a progressbar for mesh warping.


medakk (approved):
This is useful!

Maybe add an issue for mesh warp performance in the long run

PR #1252, opened 2021-05-20
Mesh Warping for Multiple Meshes

riddhishb:
Updates the` warp-mesh` command to handle warping multiple meshes at once utilizing the same reference mesh and points.

Usage:

- For single warp, it it the same usage `shapeworks warp-mesh --reference_mesh=template.ply --reference_points=template.particles --target_points=output.particles --target_mesh=output.vtk`
- For multiple point files set `--multiple` tag to be true (false by default) and pass a txt file of target point paths and target mesh paths. `shapeworks warp-mesh --reference_mesh=template.ply --reference_points=template.particles --target_points=output_particles.txt --target_mesh=output_meshes.txt --multiple=true`

### DIFF
--- a/Applications/shapeworks/MeshCommands.cpp
+++ b/Applications/shapeworks/MeshCommands.cpp
@@ -1023,8 +1023,6 @@ bool CompareMesh::execute(const optparse::Values &options, SharedCommandData &sh
   }
 }
 
-
-
 ///////////////////////////////////////////////////////////////////////////////
 // WarpMesh
 ///////////////////////////////////////////////////////////////////////////////
@@ -1033,16 +1031,11 @@ void WarpMesh::buildParser()
   const std::string prog = "warp-mesh";
   const std::string desc = "warps a mesh given reference and target particles";
   parser.prog(prog).description(desc);
-
   parser.add_option("--reference_mesh").action("store").type("string").set_default("").help("Name of reference mesh.");
   parser.add_option("--reference_points").action("store").type("string").set_default("").help("Name of reference points.");
-  parser.add_option("--target_points").action("store").type("string").set_default("").help("Name of target points. If multiple--> list of target points");
-  parser.add_option("--target_mesh").action("store").type("string").set_default("").help("Name of output target mesh. If multiple--> list of target meshes");
-  parser.add_option("--multiple").action("store").type("bool").set_default(false).help("Flag true is multiple meshes are to be warped.");
-
+  parser.add_option("--target_points").action("store").type("multistring").set_default("").help("Names of target points (must be followed by `--`), ex: \"... --target_points *.particles -- ...");
   Command::buildParser();
 }
-
 bool WarpMesh::execute(const optparse::Values &options, SharedCommandData &sharedData)
 {
   std::string inputMeshFilename = options["reference_mesh"];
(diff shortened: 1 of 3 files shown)


akenmorris:
@cchriste, do you have any feedback for this?  @riddhishb needs to be able to generate multiple meshes without having to regenerate the warp (for speed).

We still need to add the Python API for this functionality.

cchriste:
I have feedback and would like to comb over this code more carefully (upon cursory inspection it could probably be done with one loop rather than `if (multiple) ... else ...`).

There appear to be no new tests for this functionality. There should already be tests in ParticleTests for all types of cases (if not, please also add them). These new tests must be added to shapeworksTests.

In the code itself, we shouldn't be loading files to test the functionality. Think of it as layers of an onion: the outer layers load the files and the inner layers perform the computations. If the classes themselves need to be reorganized, so be it.

Adding the Python API can be considered a separate issue (please create one). 

As a suggestion for a test case: how can we warp the mean cranium mesh (used in some of Riddhish's previous work) to the given sample being evaluated by the model? Ross says it's easy and to just ask, so maybe we can address both things with one example case.

Applications/shapeworks/MeshCommands.cpp
```diff
@@ -1068,26 +1069,64 @@ bool WarpMesh::execute(const optparse::Values &options, SharedCommandData &share
...
-    staticPoints.resize(3, numParticles);
-    staticPoints.transposeInPlace();
-    movingPoints.resize(3, numParticles);
-    movingPoints.transposeInPlace();
-
-    MeshWarper warper;
-    warper.set_reference_mesh(inputMesh.getVTKMesh(), staticPoints);
-    Mesh output = warper.build_mesh(movingPoints);
-    output.write(targetMeshFilename);
+    if(multiple){
```
cchriste, line 1079: add spaces to more closely match the style in the rest of the file: `if (multiple) {` (it also seems the rest of the file inconsistently puts open braces on the same line or a separate line, so this example uses same line, but it does use spaces)

Applications/shapeworks/MeshCommands.cpp
```diff
@@ -1068,26 +1069,64 @@ bool WarpMesh::execute(const optparse::Values &options, SharedCommandData &share
...
+      while(inputFileMeshes >> line)
+          outMeshes.push_back(line);
+      inputFileMeshes.close();
+      
+      if (paths.size() - 1 != outMeshes.size()){
+        std::cerr << "warpmesh error: number of output point files should be equal to the number of output mesh files\n";
+        return false;
+      }
+
+      ParticleSystem particlesystem(paths);
```
cchriste, line 1097: can this loop simply be used for single-file cases as well?

cchriste (changes requested):
(please see my comment on the request)

Applications/shapeworks/MeshCommands.cpp
riddhishb: got it

Applications/shapeworks/MeshCommands.cpp
riddhishb: Yes, I just conditioned if people want to pass directly the file paths in place of .txt files. If we are ok with that then we can just have the single command without the --multiple tag and change the tests accordingly.

riddhishb:
> I have feedback and would like to comb over this code more carefully (upon cursory inspection it could probably be done with one loop rather than `if (multiple) ... else ...`).
> 
> There appear to be no new tests for this functionality. There should already be tests in ParticleTests for all types of cases (if not, please also add them). These new tests must be added to shapeworksTests.
> 
> In the code itself, we shouldn't be loading files to test the functionality. Think of it as layers of an onion: the outer layers load the files and the inner layers perform the computations. If the classes themselves need to be reorganized, so be it.
> 
> Adding the Python API can be considered a separate issue (please create one).
> 
> As a suggestion for a test case: how can we warp the mean cranium mesh (used in some of Riddhish's previous work) to the given sample being evaluated by the model? Ross says it's easy and to just ask, so maybe we can address both things with one example case.

riddhishb:
As for the warping test I think we can warp the cranium to other shapes as long as we have two things. 1) A good template mesh that is closed and manifold faces and all that good and pretty stuff, 2) the corresponding landmarks on the template and the moving mesh positions.

Applications/shapeworks/MeshCommands.cpp
riddhishb: Fixed

Applications/shapeworks/MeshCommands.cpp
riddhishb: @cchriste @akenmorris what do you think about this ?

riddhishb:
So changes after a fruitful discussion with Cameron. Again thanks a bunch Cameron for the starting code that saved me a lot of time.

The function now takes three aruments
```
Usage: warp-mesh [args]...

warps a mesh given reference and target particles

Options:
  -h, --help            show this help message and exit
  --reference_mesh=STRING
                        Name of reference mesh.
  --reference_points=STRING
                        Name of reference points.
  --target_points <list of strings>
                        Names of target points (must be followed by `--`), ex:
                        "... --target_points *.particles -- ...
```
I have tested this for a single .particle file as well as passing `--target_points  <path/to/directory/containing/target/points/*.particles>` and both works as expected

I think this is cleaner and looks good, I just have one question in regards to the output meshes whose argument is now removed. Currently, the output will be saved where the target points are with .particles replaced by .vtk 
We might consider having an optional argument `save_dir` that will be the directory where the output meshes will be saved. 

Thoughts @akenmorris @cchriste?

cchriste:
> We might consider having an optional argument `save_dir` that will be the directory where the output meshes will be saved.
I think that's a good idea.

riddhishb:
Ok done

Applications/shapeworks/MeshCommands.cpp
```diff
@@ -1033,62 +1031,62 @@ void WarpMesh::buildParser()
...
-
+    std::string infnm;
+    
+    if (saveDir.length() > 0) {
+      boost::filesystem::create_directories(saveDir);
+    }
+
+    for (int i = 0; i < targetPointsFilenames.size() - 1; i++) {
+      infnm = targetPointsFilenames[i];
+      infnm.replace(infnm.end()-9, infnm.end(), "vtk");
```
cchriste, line 1078: I'm nervous about assuming input files end with `particles`. Maybe check this or simply replace everything after the last `.` with `vtk` instead of `-9`.

cchriste (changes requested):
please see my comment below.

Applications/shapeworks/MeshCommands.cpp
```diff
@@ -1033,62 +1031,62 @@ void WarpMesh::buildParser()
...
     staticPoints.transposeInPlace();
-    movingPoints.resize(3, numParticles);
-    movingPoints.transposeInPlace();
-
     MeshWarper warper;
     warper.set_reference_mesh(inputMesh.getVTKMesh(), staticPoints);
-    Mesh output = warper.build_mesh(movingPoints);
-    output.write(targetMeshFilename);
-
+    std::string infnm;
```
cchriste, line 1070: I know it's picky, but a better name for this variable would make the code much easier to understand. As it is now, I still don't get what `infnm` means. I think the `nm` part means `name`, but `inf`?

Applications/shapeworks/MeshCommands.cpp
```diff
@@ -1033,62 +1031,62 @@ void WarpMesh::buildParser()
...
+    
+    if (saveDir.length() > 0) {
+      boost::filesystem::create_directories(saveDir);
+    }
+
+    for (int i = 0; i < targetPointsFilenames.size() - 1; i++) {
+      infnm = targetPointsFilenames[i];
+      infnm.replace(infnm.end()-9, infnm.end(), "vtk");
+      if (saveDir.length() > 0) {
+        int idx = static_cast<int>(infnm.rfind('/'));
```
cchriste, line 1080: My compliments for this being a very clear way to set the directory for the output files. 👍

Applications/shapeworks/MeshCommands.cpp
riddhishb: I usually use it for 'in file name' ill change it :D

Applications/shapeworks/MeshCommands.cpp
riddhishb: ok makes sense I'll make that change.

cchriste:
Looks great! Please merge when it passes checks.

riddhishb:
I dunno what the check fail is, @akenmorris any idea ? 

Thanks

akenmorris:
Not sure, re-trying first.

akenmorris:
Just a fluke, proceed.

cchriste:
Alan, are you good on this?

cchriste (approved):

PR #1254, opened 2021-05-23
Try tiny tests via GH action

akenmorris:
This PR adds all of the use cases with --tiny_test to the automated tests.  Success is determined with a zero exit code and the existence of a particular check file for each use case.

Several use cases have been modified to run non-interactively.

The only remaining test case is deep_ssm.   It is currently commented out.

These use cases rely on the data portal and use login details from a GH secret.


cchriste (approved):

Testing/UseCaseTests/UseCaseTests.cpp
```diff
@@ -0,0 +1,149 @@
...
+{
+  run_use_case("femur_cut",
+               "Output/femur_cut/shape_models/32/m03_L_femur.isores.pad.com.center.aligned.tpSmoothDT_local.particles");
+}
+
+//---------------------------------------------------------------------------
+TEST(UseCaseTests, deep_ssm)
+{
+  /// TODO: Disabled, doesn't run on windows GH
+  return;
```
cchriste, line 128: Is there a GitHub issue for this?

akenmorris: As we discussed in the developer meeting, I think we just need to drop it for now.  @jadie1 was going to look into whether it's feasible to have a CPU based tiny test for this.

PR #1260, opened 2021-05-25
Fix #846 - <use_shape_statistics_after> particle count.

### LINKED ISSUE #846: use_shape_statistics_after can potentially create more particles than number_of_particles
If use_shape_statistics_after > number_of_particles, the total number of particles is increased to use_shape_statistics_after. The fix is probably just a `use_shape_statistics_after = std::min(use_shape_statistics_after, number_of_particles)` somewhere

akenmorris:
@medakk's suggestion worked.  I tested before and after the fix and confirmed.

### DIFF
--- a/Libs/Optimize/Optimize.cpp
+++ b/Libs/Optimize/Optimize.cpp
@@ -67,8 +67,6 @@ bool Optimize::Run()
 
   }
 
-
-
   // sanity check
   if (this->m_domains_per_shape != this->m_number_of_particles.size()) {
     std::cerr <<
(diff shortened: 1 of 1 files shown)


Libs/Optimize/Optimize.cpp
```diff
@@ -76,6 +76,10 @@ bool Optimize::Run()
     return false;
   }
 
+  // ensure use_shape_statistics_after is no more than the total
+  this->m_use_shape_statistics_after = std::min(this->m_use_shape_statistics_after,
+                                                this->m_number_of_particles[0]);
```
medakk, line 81: shouldn't it be the min of m_number_of_particles, and not m_number_of_particles[0]? (i'm assuming that this vector corresponds to multi domains)

akenmorris: Yes, you're right.

Issue #1242, opened 2021-05-13
VTK/VTP output from shapeworks command and python is somehow wrong


akenmorris:
I don't see any obvious explanation for this in the code but compare the output of these commands, which should result in identical meshes:

```
shapeworks read-mesh --name in.vtk smooth --iterations 10 --relaxation 1 write-mesh --name out.vtk
shapeworks read-mesh --name in.vtk smooth --iterations 10 --relaxation 1 write-mesh --name out.vtp
shapeworks read-mesh --name in.vtk smooth --iterations 10 --relaxation 1 write-mesh --name out.obj
shapeworks read-mesh --name in.vtk smooth --iterations 10 --relaxation 1 write-mesh --name out.stl
shapeworks read-mesh --name in.vtk smooth --iterations 10 --relaxation 1 write-mesh --name out.ply
```

<img width="701" alt="image" src="https://user-images.githubusercontent.com/1693349/118151936-693bde00-b3d1-11eb-8bc8-c7b5a7bf17d1.png">

Notice that the .vtk and .vtp meshes aren't smoothed.

The same thing can be observed using the python API.

This is with the 6.0 release as well as master branch.

akenmorris:
Ah, on closer inspection and thought, I realize the difference is that vtk/vtp format are storing the normals and we are not regenerating the normals after this smooth operation.

Issue #1043, opened 2021-02-23
Small memory leak in StringUtils::getFileNameWithoutExtension


akenmorris:
This is pretty minor, but this function does leak memory.

https://github.com/SCIInstitute/ShapeWorks/blob/f9f0eb798895c77dd3c4d071028f28a043e7b2b6/Libs/Utils/StringUtils.cpp#L81

Issue #984, opened 2021-02-16
Add mesh grooming to Studio


sheryjoe:
Parent issue for adding mesh grooming tools to Studio.

Children issues thus far:
#922

sheryjoe:
- Mesh fixes from FEMesh lib, expose individual ops, e.g. smoothing, decimation, fix winding
- remeshing from mmgtools (scale functions? for adaptive meshing especially with high curvature regions)
- fill holes
- center of mass alignment
- ICP alignment

akenmorris:
@sheryjoe , what parameters to FEFixMesh should we expose in Studio?  Right now I just have a checkbox called "Quality Control" that uses whatever the defaults for Mesh::fix are:

  Mesh& fix(bool smoothBefore = true, bool smoothAfter = true, double lambda = 0.5, int iterations = 1, bool decimate = true, double percentage = 0.5);

akenmorris:
See https://github.com/SCIInstitute/ShapeWorks/blob/c42e24ee262eda1af1c196f2b37ff7b76f48c3f6/Libs/Mesh/Mesh.cpp#L508

akenmorris:
@sheryjoe , should these steps from PreviewQC be broken up into separate steps?  I'm not sure where this combined "fix" call came from.

sheryjoe:
> @sheryjoe , should these steps from PreviewQC be broken up into separate steps? 

Yes.

> I'm not sure where this combined "fix" call came from.

I think this is the consolidation of the PreviewCmd command.

akenmorris:
Ok, do we want to support smoothing both before and after decimation?

sheryjoe:
@akenmorris should we push this issue to 6.2 as it is partially resolved?

akenmorris:
Yes

akenmorris:
Completed.

PR #1261, opened 2021-05-26
Studio mesh grooming

akenmorris:
This PR adds basic mesh grooming to Studio.  (See #984).

This adds:

* Fill Holes
* Smooth (VTK Laplacian, VTK Windowed Sinc) with parameters
* ICP (with parallel find reference mesh).

This also fixes:

#1262 - Disable mesh/DT surface reconstructions if local particles are not specified in an analysis XML
#942 - Reset hover state on new optimize
#1043 - Small memory leak in StringUtils::getFileNameWithoutExtension
#1242 - VTK/VTP output from shapeworks command and python is somehow wrong

This also adds support for full 4x4 prefix matrices from project spreadsheet into Optimize.


jadie1 (approved):
I tested this with a few different kinds of meshes and the grooming seemed to work as expected!

Studio/src/Application/Visualization/ShapeWorksStudioApp.cpp
```diff
@@ -765,7 +765,7 @@ void ShapeWorksStudioApp::handle_progress(int value)
   if (value < 100) {
     this->progress_bar_->setVisible(true);
     this->progress_bar_->setValue(value);
-    this->disable_all_actions();
+    //this->disable_all_actions();
```
cchriste, line 768: ?

Studio/src/Application/Data/Session.cpp
```diff
@@ -684,6 +698,11 @@ double Session::update_auto_glyph_size()
   }
 
   this->auto_glyph_size_ = max_range / std::sqrt(static_cast<double>(num_particles)) / 2;
+
+  std::cerr << "max_range = " << max_range << "\n";
```
cchriste, line 702: Need to print these? I sometime keep them commented since they can be so useful when debugging (see Mesh::toImage and Python binding of Image.init).

cchriste (approved):
All this looks good to me.
Tests verify functionality.
I haven't tried it in Studio.

Studio/src/Application/Data/Session.cpp
akenmorris: Yeah, I need to push a cleanup commit I already made.

Studio/src/Application/Visualization/ShapeWorksStudioApp.cpp
akenmorris: The commit message for this probably didn't make it clear.  I decided we no longer need to disable the actions such as "save", etc just because something is going on the progressbar.  We have other things using the progress bar such as the mesh warping.  I'll add to the cleanup commits along with conflict resolution.

PR #1265, opened 2021-06-02
Image::getMesh returns a mesh instead of vtkPolyData

archanasri:


Libs/Image/Image.h
```diff
@@ -244,7 +244,7 @@ class Image
   TransformPtr createRigidRegistrationTransform(const Image &target, float isoValue = 0.0, unsigned iterations = 20);
 
   /// creates a vtkPolyData for the given image
-  static vtkSmartPointer<vtkPolyData> getPolyData(const Image& image, PixelType isoValue = 0.0);
+  static Mesh getMesh(const Image& image, PixelType isoValue = 0.0);
```
cchriste, line 247: Unless there is a specific reason for this function to be static, let's simply make it a member of Image.

cchriste (changes requested):
Looks good except for the (perhaps) unnecessarily static function in Image. 

I also wonder if we might add `using Transform = vtkSmartPointer<vtkMatrix4x4>;` to Shapeworks.h.

Libs/Image/Image.h
archanasri: Done

cchriste (approved):
An even more worthwhile fix than expected, simplifying the file layouts and compilation.

Issue #1223, opened 2021-04-28
FileNotFoundError: [Errno 2] No such file or directory: 'SurfaceToSurfaceDistance': 'SurfaceToSurfaceDistance'


sainatarajan:
Hello,

Can someone help how to solve this error? I'm pretty sure that the script is calling a command `SurfaceToSurfaceDistance` which is not available in my system. I have tried searching on the web and I couldn't find anything.

Use case command: `python RunUseCase.py --use_case deep_ssm` and the dataset is femur-v0.

Here is the stack trace below:

```
raceback (most recent call last):
  File "RunUseCase.py", line 94, in <module>
    module.Run_Pipeline(args)
  File "/home/sainatarajan/Downloads/ShapeWorks-v6.0.0-linux/Examples/Python/deep_ssm.py", line 124, in Run_Pipeline
    avg_distance = DeepSSMUtils.analyzeResults(out_dir, DT_dir, prediction_dir, mean_prefix)
  File "/home/sainatarajan/anaconda3/envs/dl_gpu/lib/python3.7/site-packages/DeepSSMUtils/__init__.py", line 24, in analyzeResults
    avg_distance = Analyze.getDistance(out_dir, DT_dir, prediction_dir, mean_prefix)
  File "/home/sainatarajan/anaconda3/envs/dl_gpu/lib/python3.7/site-packages/DeepSSMUtils/Analyze.py", line 36, in getDistance
    orig_dist_mesh_list, pred_dist_mesh_list = surface2surfaceDist(orig_mesh_list, pred_mesh_list, out_dir)
  File "/home/sainatarajan/anaconda3/envs/dl_gpu/lib/python3.7/site-packages/DeepSSMUtils/Analyze.py", line 119, in surface2surfaceDist
    subprocess.check_call(execCommand)
  File "/home/sainatarajan/anaconda3/envs/dl_gpu/lib/python3.7/subprocess.py", line 358, in check_call
    retcode = call(*popenargs, **kwargs)
  File "/home/sainatarajan/anaconda3/envs/dl_gpu/lib/python3.7/subprocess.py", line 339, in call
    with Popen(*popenargs, **kwargs) as p:
  File "/home/sainatarajan/anaconda3/envs/dl_gpu/lib/python3.7/subprocess.py", line 800, in __init__
    restore_signals, start_new_session)
[...]

Issue #1105, opened 2021-03-04
Refactor DeepSSM


jadie1:
- Make model and options modular for eventual studio integration
- Use conflict file for model parameters

PR #1270, opened 2021-06-04
DeepSSM Updates

jadie1:
This PR addresses:

- Issue #1105 Refactoring DeepSSM to use a config file
- Issue #1223 Fixing surface to surface distance in eval
- Updates getting meshes from distance transforms to use pybind `toImage()` instead of command line `MeshFromDistanceTransforms`
- Updates getting meshes from particles to use `mesh-warp` instead of `ReconstructSurface`
- Updates use case to print warning but still run on CPU rather than exiting
- Improves the tiny test so it runs in about 2 min

To test, run the tiny test or the full use case (both should run on GPU or CPU). 
This will probably require running:
 `pip install Python/DeepSSMUtilsPackage/` 
`pip install Python/DataAugmentationUtilsPackage/`

### DIFF
--- a/Examples/Python/RunUseCase.py
+++ b/Examples/Python/RunUseCase.py
@@ -16,85 +16,88 @@ import subprocess
 import sys
 import setupenv
 
-# check that required modules are found
-try:
-    import termcolor
-    import DatasetUtils
-except ImportError as error:
-    print("\nError: Unable to import ShapeWorks Python dependencies\n")
-    print("Please make sure you have run \"source conda_installs.sh\" (once) and \"conda activate shapeworks\" (each time)")
-    print("See https://github.com/SCIInstitute/ShapeWorks for more information")
-    sys.exit(1)
 
-from CommonUtils import *
+if __name__ == '__main__':
 
-# Default installation path for each platform. If using your own build, specify --shapeworks_path
-# Note, the path for linux differs from setupenv.py because it is set up for the notebooks
-# that are two more levels deep
-default_bin_path = "../../bin" # Linux
-if platform.system() == "Windows":
-    default_bin_path = "C:\\Program Files\\ShapeWorks\\bin" # Windows
-if platform.system() == "Darwin":
-    default_bin_path = "/Applications/ShapeWorks/bin" # Mac
-default_subsample = 3
+    # check that required modules are found
+    try:
+        import termcolor
+        import DatasetUtils
+    except ImportError as error:
+        print("\nError: Unable to import ShapeWorks Python dependencies\n")
+        print("Please make sure you have run \"source conda_installs.sh\" (once) and \"conda activate shapeworks\" (each time)")
+        print("See https://github.com/SCIInstitute/ShapeWorks for more information")
+        sys.exit(1)
 
-parser = argparse.ArgumentParser(description='Example ShapeWorks Pipeline')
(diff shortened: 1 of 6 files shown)


akenmorris:
Looks pretty good.  In order to use deep_ssm --tiny_test with the github actions, we will need to turn off all the displays.  Right now it's bringing up the violin plot and this will hand on the Windows GH.  What do you think?

akenmorris:
It's very possibly I've done something wrong (I did do the two pip installs), but it's hanging for me, using no CPU after it brought up the violin plot.  On the console, I see:
```

********************* WARNING ****************************
Pytorch is running on your CPU!
This will be very slow. If your machine has a GPU,
please reinstall Pytorch to your shapeworks conda
environment with the correct CUDA version.
**********************************************************
Loading data loaders...
Done.
Defining net...
Done.
Beginning training on device = cpu

Epoch	  LR	  Train_Err	  Train_Rel_Err	  Val_Err	  Val_Rel_Err	  Sec
[W ParallelNative.cpp:206] Warning: Cannot set number of intraop threads after parallel work has started or after set_num_threads call when using native parallel backend (function set_num_threads)
[W ParallelNative.cpp:206] Warning: Cannot set number of intraop threads after parallel work has started or after set_num_threads call when using native parallel backend (function set_num_threads)
[W ParallelNative.cpp:206] Warning: Cannot set number of intraop threads after parallel work has started or after set_num_threads call when using native parallel backend (function set_num_threads)
[W ParallelNative.cpp:206] Warning: Cannot set number of intraop threads after parallel work has started or after set_num_threads call when using native parallel backend (function set_num_threads)
[W ParallelNative.cpp:206] Warning: Cannot set number of intraop threads after parallel work has started or after set_num_threads call when using native parallel backend (function set_num_threads)
[W ParallelNative.cpp:206] Warning: Cannot set number of intraop threads after parallel work has started or after set_num_threads call when using native parallel backend (function set_num_threads)
[W ParallelNative.cpp:206] Warning: Cannot set number of intraop threads after parallel work has started or after set_num_threads call when using native parallel backend (function set_num_threads)
[W ParallelNative.cpp:206] Warning: Cannot set number of intraop threads after parallel work has started or after set_num_threads call when using native parallel backend (function set_num_threads)
```

jadie1:
> Looks pretty good. In order to use deep_ssm --tiny_test with the github actions, we will need to turn off all the displays. Right now it's bringing up the violin plot and this will hand on the Windows GH. What do you think?

Okay, I will fix it so the plot is not displayed for the tiny_test

jadie1:
Are you running on mac? I am not seeing this CPU error on linux.... I'll try to look into it.

akenmorris:
Yeah, Mac. I am going to try on another Linux machine. Waiting for conda to “solve” the environment.

jadie1:
@akenmorris Looks like this is a mac problem and the solution is to not use parallel computing: https://stackoverflow.com/questions/64772335/pytorch-w-parallelnative-cpp206

akenmorris:
Yeah, setting OMP_NUM_THREADS=1 fixes it.  Should we set that env var in the use case for Mac?

jadie1:
Yeah that sounds good. So should I just add this to the use case:
```
if platform.system() == "Darwin":
	os.system("export OMP_NUM_THREADS=1")
```

Issue #1247, opened 2021-05-18
Use case and notebooks reference selection


jadie1:
Need to create Pymodule function for reference selection using the auto reference selection that studio uses (all pair ICP (parallel) and finds the sample that is minimum average distance to every other shape). Reference selection for segmentations and meshes should both use this.

Need to update all use cases and notebooks to use the pymodule function.

jadie1:
Closed via PR #1275

PR #1275, opened 2021-06-09
Updating reference selection

jadie1:
Addresses issue #1247
Updates find_reference_image_index() to convert the segmentations to meshes and use sw.MeshUtils.findReferenceMesh() to select reference. 
Removes find_reference_mesh_index() because it wasn't being used anywhere and now sw.MeshUtils.findReferenceMesh() can be used directly.

### DIFF
--- a/Python/shapeworks/shapeworks/utils.py
+++ b/Python/shapeworks/shapeworks/utils.py
@@ -1,6 +1,7 @@
 import numpy as np
 import os
 import vtk
+import shapeworks as sw
 
 # helper function to determine the best grid size (rows and columns) given the number of samples in a dataset.
 def postive_factors(num_samples):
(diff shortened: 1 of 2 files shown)


Python/shapeworks/shapeworks/utils.py
```diff
@@ -1,6 +1,7 @@
 import numpy as np
 import os
 import vtk
+import shapeworks as sw
```
cchriste, line 4: Doesn't this mean the library is including itself?

Python/shapeworks/shapeworks/utils.py
```diff
@@ -87,54 +88,11 @@ def get_file_with_ext(file_list,extension):
...
-                dists[mesh_index, comp_index] = dist
-                dists[comp_index, mesh_index] = dist
-    dists = np.sum(dists, axis=0)
-    median_index = np.argmax(dists)
-    return median_index
+    mesh_list = []
+    for img in inDataList:
+        mesh = img.toMesh(0.5)
+        mesh_list.append(mesh)
+    return sw.MeshUtils.findReferenceMesh(mesh_list)
```
cchriste, line 95: Since we import * from shapeworks_py into the library namespace, will simply using MeshUtils.find... work?

cchriste (commented):
The bigger concern is that our module might be including itself in other files as well.

cchriste:
> The bigger concern is that our module might be including itself in other files as well.

I guess this is how it's supposed to work for traditional packages. https://docs.python.org/3/reference/import.html

Python/shapeworks/shapeworks/utils.py
cchriste: Yes it does. Import away!

cchriste (approved):

cchriste:
Btw, this runs fine, but I have to set PYTHONPATH to include both the path to <swcode>/Python/shapeworks as well as <build>/bin. Anyway, I'm gonna sleep now. Request approved.

akenmorris (approved):

Issue #1267, opened 2021-06-02
use try/catch for all Python tests


cchriste:
Examples that already do this include toArray, initImage, and crop.

cchriste:
We could also do this for shell script tests using something like this: https://stackoverflow.com/questions/22009364/is-there-a-try-catch-command-in-bash.

I'm just worried this will complicate creating tests for the command line. The advantage is it would remove FAILED from the output when tests are supposed to fail in the first place.

PR #1277, opened 2021-06-09
[WIP] simplify static python bindings, add tests for boundingBox functions, make calls to these functions more efficient

cchriste:
- [x] simplify binding static functions that don't require lambda custom calls in Python bindings; 
- [x] make boundingBox functions safer by passing const images or meshes (used std::reference_wrapper<const X>); 
- [x] improve python boundingBox tests by verifying results and add a Python boundingbox test for meshes; 

...and use try/catch for all tests, not only the ones expected to throw exceptions (separated into #1267 instead of this pr)


akenmorris:
Is this for 6.1?

cchriste:
> Is this for 6.1?

I thought it was fine to include it since it's an improved test and some binding simplifications, but it's not compiling for some reason and not critical.

cchriste:
This pull request consists of three items:

cchriste:
This is ready to go. Please give it a gander and merge when approved.

It also include some timing code. The bounding box tests are now faster (no more copying the Image/Mesh for each iteration in their for loops) as well as safer (const). The Python bindings also work with the same degree of efficiency.

While these are worthwhile improvements, it's still questionable that all images/meshes should be loaded before trying to compute their bounds. Someone couldn't even get past that part of one of the use cases when testing 6.1 (was an issue created for this?).

sheryjoe:
> it's still questionable that all images/meshes should be loaded before trying to compute their bounds. Someone couldn't even get past that part of one of the use cases when testing 6.1 (was an issue created for this?).

@cchriste we don't have to load all images. Determining the largest bounding box for a group of segmentations is an incremental computation process.

cchriste:
> > it's still questionable that all images/meshes should be loaded before trying to compute their bounds. Someone couldn't even get past that part of one of the use cases when testing 6.1 (was an issue created for this?).
> 
> @cchriste we don't have to load all images. Determining the largest bounding box for a group of segmentations is an incremental computation process.

That is completely correct, and we even have a function that accepts a list of filenames instead of an array of images/meshes, performing the computation incrementally.

archanasri (approved):

akenmorris (approved):
Looks good to me.

Issue #1296, opened 2021-06-15
installation using OSX .pkg does not remove previous install (Windows has the same problem with nullsoft installer)


cchriste:
This can result in many different issues, including:
- remaining dependent on now-retired executables 
- old python scripts still being used

It is a much bigger issue that the short list above might suggest. For example, if a user has created their own custom version of an old use case script that called a now-consolidated executable, not only will they incorrectly believe to have upgraded ShapeWorks, but they will not get bug fixes from the consolidation. Retired scripts like CommonUtils still exist and may be used. 

The brute force solution of first deleting the destination directory would be reasonable. Or we can ask the user to manually delete it.

akenmorris:
I'm not sure I understand this issue, or more specifically, what it has to do with an OSX pkg installer removing a previous install.

If a user has copied the Examples/Python folder as instructed (since /Applications/ShapeWorks is not writable) and modified it for their particular use case, and then they upgrade ShapeWorks and the old executables they were using are no longer present, what do you want the pkg installer to do about it?

There's a fundamental issue that python scripts from previous versions of ShapeWorks will generally not work with newer versions of ShapeWorks.

cchriste:
> the old executables they were using are no longer present, what do you want the pkg installer to do about it?

The old executables, unless replaced, still exist in the new installation directory. That's the problem.

> 
> There's a fundamental issue that python scripts from previous versions of ShapeWorks will generally not work with newer versions of ShapeWorks.

And that's _correct_ behavior. We want them to break. But they won't always do that when /Applications/ShapeWorks isn't first deleted by the new installation because cruft that has been removed will still be there.

akenmorris:
Ok, so we can add a "preinstall" script to be run by the pkg installer.  There's various env vars including:

DSTROOT=/Applications/ShapeWorks

Do you think it's safe to just do "rm -rf $DSTROOT"?

akenmorris:
Tried it on another machine and it doesn't seem to work afterall.

cchriste:
Same issue for Windows.

akenmorris:
Windows installer now prompts to remove old and does so.

cchriste:
Two issues,
- windows uninstall leaves "uninstall.exe" in default installation directory (minor issue, but wanted to note it)
- windows uninstall **assumes default directory installed**, then reinstalls **without deleting existing install** when using alternate path
[...]

PR #1302, opened 2021-06-15
Add OSX preinstall script to remove /Application/ShapeWorks

akenmorris:
Address #1296.

This won't work if the user hasn't installed it on their main volume.  I'm wary of using the undocumented $DSTROOT which seems to point to the right place, but without guarantees I don't want to remove it if it could be wrong.

### DIFF
--- a/Support/osxscripts/preinstall
+++ b/Support/osxscripts/preinstall
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+echo "ShapeWorks preinstall"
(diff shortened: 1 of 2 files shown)


cchriste (approved):
Given the lack of documented alternatives, I think this is an appropriately conservative approach to address this. It might not cover every possible installation case (e.g., a user selects a different installation volume), but it takes care of the vast majority.

Issue #1297, opened 2021-06-15
install_name_tool is not available on non-developer macs


akenmorris:
Unfortunately, install_name_tool isn't available on macs without Xcode.  Running it (as the conda/python install scripts do) results in a window being popped up prompting the user to install the compiler toolchain stuff.  Even if the user does that, it's too late for the script which has already moved on and then the python module stuff doesn't work.

cchriste:
Nice catch. Thanks to whomever let you use their non-dev system to try this. 
I'm on it.

Issue #884, opened 2021-01-08
Identify subset of dependencies needed for user installations


cchriste:
Users don't need to run the full conda_installs since it brings things like cmake in which are not necessary for using a prebuilt ShapeWorks installation. We should identify the subset of dependencies needed for users. This will make user installations faster and less prone to conflicts (though combining conda and pip still renders this more likely).

In addition, this will enable partitioning the [e.g., conda] installation between shapeworks-core, shapeworks-python, and shapeworks-studio, since presumably Qt/Vtk OpenGL libraries aren't needed for command line tools (though they may be needed for the Python module since it provides helper functions that use itkwidgets, pyvista, and vtk; unless we further partition the module).

akenmorris:
My suggestion is that we add a "--developer" option to the script rather than split it into two scripts (one for developers and one for us).  Then with the --developer argument, the additional packages will be installed.

akenmorris:
Hey, maybe a better idea is to just detect if you're sitting in a git repo (e.g. is .git a folder?) and if so, do the developer stuff.  Then we don't run into the problem that someone on the team forgets --developer and is confused why stuff isn't working.

cchriste:
As I await conda_installs.sh for testing the 6.1 release, it begins to install cmake and ctest, obviously unnecessary to users.
This is a huge time sink for users. 
I also wonder if we need should be including things like our test executables for users.

cchriste:
> I also wonder if we need should be including things like our test executables for users.

Actually, we're not packaging test executables, but on Windows all the test .pdb files are being included, and on all platforms libTesting is added (maybe it's still needed).

cchriste:
Overall proposal for installation:
- create and package install_shapeworks.sh/bat that only includes libraries necessary for users
- create and **do not package** developer_installs.sh that includes all libraries necessary for dev

PR #1315, opened 2021-06-16
Fix Install stuff

akenmorris:
* Don't call nbstripout --install --attributes .gitattributes if git is not installed.
* Add missing update_installed_python_module_binary_path.py to package
* Fix deletion /Application/ShapeWorks -> /Applications/ShapeWorks

#1297
#1296

### DIFF
--- a/Support/osxscripts/preinstall
+++ b/Support/osxscripts/preinstall
@@ -3,4 +3,4 @@
 echo "ShapeWorks preinstall"
 env
 echo "Removing previous ShapeWorks install"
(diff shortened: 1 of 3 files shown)


cchriste:
This is a quick fix to #884, which we really need to get resolved. It's painful for everyone to have to wait the inordinate amount of time it takes to re-install from scratch.

cchriste (approved):
Some overlap with the changes that fix the windows install. Gonna merge that as well.

PR #1323, opened 2021-06-16
Studio multiple domain example 

iyerkrithika21:


akenmorris (approved):
Looks good.  I reset the Studio project back to original state since it want to open in Analysis mode.

PR #1327, opened 2021-06-17
Fix python install on windows

cchriste:
Working with Krithika we verified this works on Windows, and that the minor modifications to conda_installs.bat also work.


iyerkrithika21 (approved):

cchriste:
The installation scripts now work. Users can install and...
- import shapeworks from the python in the shell 
- import shapeworks from jupyter notebooks
- RunUseCase with no extra arguments *(as long as they installed to the default locations and/or run from the expected locations on Linux)

PR #1330, opened 2021-06-17
add things that might help with some installations (low-priority, so just add when convenient)

cchriste:
- try to remove existing anaconda environment when running conda_installs.bat
- ensure it doesn't try to creat a .bak file when updating setup_path.py


cchriste:
Just added these small items to #1335.

Issue #1333, opened 2021-06-17
add devsetup.sh/bat for developers to easily set up their environment


cchriste:
This script is basically a more comprehensive version of the two parameters that currently only work for RunUseCase.py (and are notably confusing)[https://github.com/SCIInstitute/ShapeWorks/issues/1308]. 

Rather than narrowly try to simply make them less confusing, let's add this often-requested script, which will take the same two parameters as RunUseCase, and from them:
- add their build/bin to the PATH, and
- add the Python modules in their source to PYTHONPATH

For those who have been developing Python modules by pip installing every time they change anything, this will make their lives simpler by eliminating that necessity. And for everyone it will reduce potential errors due to inconsistencies between these two env vars.

cchriste:
Completed as of pr #1338, making development and testing significantly less likely to inadvertently import or execute incompatible modules or exectables.

Issue #1332, opened 2021-06-17
add shapeworks_startup.sh/bat to conda's activate.d directory.


cchriste:
Conda provides a way to run scripts that can help complete your environment when activated (and a complementary way to remove these items when deactivated). Let's use this to simply add shapeworks/bin to the environment so that when users activate everything Just Works™. 

Currently users have a variety of ways they keep the shapeworks executable and company in their path, from their .bashrc to manually setting it each time to passing it to RunUseCase (which only works for that particular script) to relying on their installation to be in a "default" location. A solution to ALL these issues for users is to simply provide this simply script to anaconda when the shapeworks environment is created.
https://docs.conda.io/projects/conda/en/master/user-guide/tasks/manage-environments.html
**Any .sh/.bat files placed in ${CONDA_PREFIX}/etc/conda/activate.d/ will be sourced when the environment is activated.**

This is very simple and uses almost exactly the same method as already used to modify our python module's search path during installation.

cchriste:
#1335 resolves this

PR #1335, opened 2021-06-18
Add activate/deactivate scripts to conda so executables are available when ShapeWorks is installed

cchriste:
Puts everything in the path when conda is activated, so no more need to specify paths (#1332), except for developers, for whom setupenv.sh/bat scripts will be coming next (#1333). 

Tested and works for local osx install. Need to test...
- [x] osx user install  
- [x] linux user install
- [x] windows user install

First need to ensure it works everywhere, then...
- [x] de-clutter main directory (move to Support dir? @akenmorris)
- [x] rename conda_installs.sh/bat to install_shapeworks.sh/bat (makes more sense even if we change nothing else at all)
  - works as is, but will be more understandable for users, and really does what it says: installs ShapeWorks


cchriste:
The only trouble with this is deactivating the environment doesn't remove shapeworks from the path. This is a known and supposedly resolved issue with Anaconda. It's not that bad for us since we really only activate shapeworks and once that's done use it. 

This change is worth merging even as is since it enables users to simply `conda activate shapeworks` and have access to all executables, jupyter notebooks (including calling system executables without path modifications), and use cases without ever needing to pass parameters for shapeworks binary paths and module directories.

cchriste:
This is all working wonderfully! Happy Saturday!!
- [x] windows - fresh custom dir, after which command line, python command line, runusecase, and jupyter notebooks all work, and studio from the command line!
- [x] linux - fresh custom dir, command line, python command line, runusecase
  - ShapeWorksStudio found in path, but no OpenGL to run remotely through XQuartz
  - jupyter server runs, but need to open port on remote machine (this will be cool 😎)
- [x] osx - default dir, uninstalls correctly(!), command line, python command line, runusecase, and jupyter notebooks all work, and studio from the command line!

PR #1338, opened 2021-06-21
Add `devenv` scripts for developers to easily configure their environments

cchriste:
- also includes a small fix for the windows installation to allow for directories with spaces


cchriste:
In summary, an excerpt from the updated developer docs:

The **devenv** script, short for **dev**eloper **env**ironment, only requires one parameter: the path to the build binaries. It infers the source path from its own location.  

It configures the environment of the terminal from which it is sourced such that:  

- PATH and PYTHONPATH point to $BUILD_BIN directory  
- PYTHONPATH points to each module in $SOURCE/Python  

This is very useful for development, since after sourcing this, processes in the
environment will use executables from the given build and import modules from
the developer's source--this includes the compiled portion of the Python bindings!

It dramatically reduces the probability of using the wrong paths for testing,
and eliminates the need to **pip install** any of the Python modules during
development.

As an example of using it with a make-based build directory inside source on Linux or OSX:
```
(shapeworks) ~/code/ShapeWorks$ source ./devenv.sh ./build/bin
```

On Windows, to point to the RelWithDebInfo binaries in a build directory inside source:
```
(shapeworks) ~/code/ShapeWorks> devenv .\build\bin\RelWithDebInfo
```

akenmorris (approved):

Issue #1349, opened 2021-06-24
RC12 documentation for Examples


patkins:
#1273 The install worked well (I was able to run a use case and notebook tutorial), however it seems the commands for the use cases have changed, but have not been updated in the Windows_README.txt 
```
4. Run one of the included use cases. To list them all, run: `python RunUseCase.py --help`.
`python RunUseCase.py --use_case ellipsoid_fd`. 
```
Anaconda console output:
```
(shapeworks) C:\Users\penny\Documents\shapeworks\Examples\Python>python RunUseCase.py --use_case ellipsoid
Using shapeworks module from C:\Users\penny\miniconda3\envs\shapeworks\lib\site-packages\shapeworks\__init__.py
Using shapeworks from c:\Program Files\ShapeWorks\bin\shapeworks.EXE
usage: RunUseCase.py [-h] [--use_subsample] [--num_subsample NUM_SUBSAMPLE]
                     [--interactive] [--skip_grooming] [--groom_images]
                     [--use_single_scale] [--tiny_test]
                     {ellipsoid,ellipsoid_evaluate,ellipsoid_mesh,ellipsoid_fd,ellipsoid_cut,ellipsoid_pca,ellipsoid_multiple_domain,ellipsoid_multiple_domain_mesh,lumps,left_atrium,femur,femur_mesh,femur_cut,deep_ssm,supershapes_1mode_contour,thin_cavity_bean}
RunUseCase.py: error: unrecognized arguments: --use_case

(shapeworks) C:\Users\penny\Documents\shapeworks\Examples\Python>python RunUseCase.py ellipsoid
```

There is also no feedback for the user if the folder doesn't exist (i.e. has not yet been unzipped) in the notebook` getting-started-with-exploring-segmentations.ipynb`, this is super minor, but may help novice users.

cchriste:
Good catch, Penny. Thank you!

Devs: I've created `misc_61` branch for little stuff like this so we don't have to wrangle a bunch more PRs but can instead get everything that doesn't modify functionality in with one later today.

sheryjoe:
@cchriste is this resolved?

PR #1350, opened 2021-06-24
little things that tidy up the 6.1 release 

cchriste:
Included items that have _no effect on functionality_:
- remove --use_case from RunUseCase command instructions in platform readmes


akenmorris:
I think it was fine to make --use-case optional, but it's now an error to specify it? (#1349).

Is there any way to make the old command work too?

cchriste:
> Is there any way to make the old command work too?

I don't think it should be optional. It's always required.

akenmorris:
I get that it's required.  It's just a backward compatibility thing.

For example, when I do ctrl-r from my terminal to search for "--groom_images" or whatever to bring up a RunUseCase command, all those commands will fail now.  Or if I had a script that was running RunUseCase.py, I have to modify it now for 6.1.  And I also won't be able to swap between 6.0 and 6.1.

I think it's probably pretty minor for RunUseCase.py, but I just bring it up in general that if it's ever simple to maintain previous syntax, we should opt for it.

sheryjoe:
I agree with @akenmorris

cchriste:
> I think it's probably pretty minor for RunUseCase.py, but I just bring it up in general that if it's ever simple to maintain previous syntax, we should opt for it.

I agree with you about maintaining previous syntax _to some reasonable degree_. 

But in this case I think it's neither simple (parameters without double-dash are required; calling this script without that parameter is impossible), nor a big deal like you said. Plus it makes things _much_ clearer to the user since `--help` no longer shows this parameter under `optional arguments:` as it did before.

akenmorris:
Is this PR ready to merge?

PR #1395, opened 2021-07-22
Stats-pvalues for the group difference

iyerkrithika21:
@akenmorris/ @jadie1  can you upload the dataset to the portal [femur_stats.zip](https://github.com/SCIInstitute/ShapeWorks/files/6864554/femur_stats.zip)
This dataset has the shape model for femur with the group ids.
This PR includes the stats function to calculate the p-values for the group difference and includes a use case. 

To test the PR run - `python RunUseCase.py femur_pvalues`

### DIFF
--- a/Examples/Python/RunUseCase.py
+++ b/Examples/Python/RunUseCase.py
@@ -50,7 +50,9 @@ if __name__ == '__main__':
     # parse arguments
     parser = argparse.ArgumentParser(description='Example ShapeWorks Pipeline')
     parser.add_argument("use_case", help="Must specify which of these use cases to run.",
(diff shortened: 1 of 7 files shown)


akenmorris:
Where do we need to upload it?  As use-case-data-v2/femur_pvalues-v0  ?

akenmorris:
```
    from hotelling.stats import hotelling_t2
ModuleNotFoundError: No module named 'hotelling'
```

Am I missing a package?

iyerkrithika21:
> 
> 
> ```
>     from hotelling.stats import hotelling_t2
> ModuleNotFoundError: No module named 'hotelling'
> ```
> 
> Am I missing a package?

Oh sorry! 
I will update the install script with this package.

iyerkrithika21:
> 
> 
> Where do we need to upload it? As `use-case-data-v2/femur_pvalues-v0` ?
Yes. 

I will change the dataset name in the code to -femur_pvalues-v0.

akenmorris (approved):
Looks good to me.  I've added an automated test and uploaded the data.

I'll need to modify this when we integrate into studio, but we can cross that bridge when we come to it.

Issue #1374, opened 2021-07-05
DeepSSM in Studio - Testing 


jadie1:
Need a tab for DeepSSM inference in studio 

Run inference button - progress bar at bottom

Display (after running): 

- All reconstructed test meshes with error heat map
- Mean error

Issue #1373, opened 2021-07-05
DeepSSM in Studio - Training


jadie1:
Need a tab in Studio for training DeepSSM

Parameters:

- Epochs
- Learning rate
- Decay learning rate - true or false 
- Fine tuning - true or false
- Epochs

Run training button - progress bar at bottom

Display (updated every 10 epochs): 

- Model training dashboard with train and validation loss
- Examples of reconstructed meshes with error heat map - best, median, and worst in validation

akenmorris:
@jadie1 , where can I get access to the training dashboard?  Is it written out after every epoch?  How do I reconstruct meshes with heat maps?

jadie1:
Currently it just prints out the info to the command line. The default is every epoch but he `trainer` `val_freq` parameter can be set to have it test on the validation set and print less often than every epoch (for example if set to 2 it prints every other) . For now in studio I think we could just set the frequency it updates to something constant like 10.  

To get the distance meshes in Python you call:
```
import shapeworks as sw
mesh1 = sw.Mesh(path_to_mesh1)
mesh2 = sw.Mesh(path_to_mesh2)
mesh1.distance(mesh2).write(path_to_out_distance_mesh1)
```

akenmorris:
Ok, this is the dashboard you want to see somewhere?

```
Epoch	  LR	  Train_Err	  Train_Rel_Err	  Val_Err	  Val_Rel_Err	  Sec
1.00000	  0.00100	  2.07978	  5.10448	  8.79684	  1040.82276	  11.55580
2.00000	  0.00098	  68.11378	  5475.04883	  12.66105	  2341.53674	  11.90325
3.00000	  0.00097	  6.27155	  46.41597	  84.35498	  122743.24353	  11.96991
```

Should we have this drawn as a chart?  Or just the table?  Or Both?

For the meshes, where are they written?  I haven't found any meshes being written as part of the training.

jadie1:
I think drawing them as a chart would be more user friendly. We could just plot the training and validation relative error over epochs. In the use case the meshes with error heat maps are not written until after training. During evaluation they are generated for the test set. We probably want to adapt this so that the best worst and average examples from the validation set are displayed.  I can write a Python function for this if that'd be helpful.

akenmorris:
Yeah, a python function that does this would be great.  We should draw the chart in python with matplotlib as well.

I will add an epoch hook to send control back to Studio and these things can be updated.

Issue #1372, opened 2021-07-05
DeepSSM in Studio - Train/Test split


jadie1:
Need to add a way of defining train/validation/test split in studio.
For now this will just be randomly split using a percentage amount for train, val, and test.

akenmorris:
This could be done with a group column, e.g.

input,group_split
one,train
two,train
three,val
four,val
five,test
six,test
seven,train
...

Issue #1371, opened 2021-07-05
DeepSSM in Studio - Data Augementation


jadie1:
Need a tab in studio for DeepSSM data augmentation 

Parameters:

- Number of samples
- Number of PCA dimensions
- Percent variability preserved
- Sampler type: Gaussian, Mixture of Gaussian, or KDE

Run Augmentation button - progress bar at bottom

Display (after running):

- Parallel violin plot
- Examples of real and augmented data

akenmorris:
Are we planning on supporting multiple domains with DeepSSM?  If so, how will that work?

sheryjoe:
@akenmorris Yes, it is still under research though. For now, let's assume single domain.

PR #1397, opened 2021-07-26
Studio DeepSSM

akenmorris:
This PR implements phase one of #977 and includes:

- [x] #1371
- [x] #1372
- [x] #1373
- [x] #1374

Phase one includes the same assumptions that the current deep_ssm use cases contains.  Namely that the images are already pre-groomed.

image

image

image

image


jadie1:
I am seeing an issue where if I change the number of dims in augmentation, it is not relayed to the training part and it throws an error. For example, if I change "Number of PCA Dimensions" to 6 in augmentation, then when I run training I get the error: 

> "Error: RuntimeError: The size of tensor a (3) must match the size of tensor b (6) at non-singleton dimension 1 At: /home/sci/jadie/software/anaconda3/envs/shapeworks/lib/python3.7/site-packages/DeepSSMUtils/trainer.py(104): supervised_train /home/sci/jadie/software/anaconda3/envs/shapeworks/lib/python3.7/site-packages/DeepSSMUtils/trainer.py(52): train /home/sci/jadie/software/anaconda3/envs/shapeworks/lib/python3.7/site-packages/DeepSSMUtils/__init__.py(25): trainDeepSSM "

Because it is expecting there to only be 3 PCA dims but there are 6.

akenmorris:
> number of dims

Ok, let me check on it.  I don't think the auto-dims part got implemented either now that I think of it.

akenmorris:
@jadie1 , the number of dimensions stuff should all be fixed now in the latest commit.

jadie1:
Looks like the number of dimensions is working as expected now!

I did notice one other thing, after running training if you resize the viewer the heat map goes away (see screenshots)

Before resizing:
image
After resizing: 
image

akenmorris:
Hmm, I haven't been able to reproduce that.  When I change the number of viewers, the error heat map stays.

Yeah, we could force to 2x2.

akenmorris:
I think the colormap issue is a problem with the scalar ranges not being reset.  In your second image, the scale goes to 20,000.

Let me investigate and fix.

jadie1 (approved):
Tested on multiple datasets and it looks like everything is working as expected now!

PR #1413, opened 2021-08-10
Move to Ubuntu 16 docker image

akenmorris:
Ubuntu 16 GitHub Action Runner will be removed in September 2021.  This PR switches to ubuntu-latest runner but using a specially made Ubuntu 16 docker image (Dockerfile included).

### DIFF
--- a/.github/workflows/Dockerfile
+++ b/.github/workflows/Dockerfile
@@ -1,0 +1,48 @@
+FROM ubuntu:xenial-20210429 AS env
+MAINTAINER akenmorris@gmail.com
+
+
+RUN apt-get update -y && apt-get upgrade -y && apt-get dist-upgrade -y && apt-get install build-essential software-properties-common -y && add-apt-repository ppa:ubuntu-toolchain-r/test -y && apt-get update -y
+RUN apt-get install gcc-9 g++-9 -y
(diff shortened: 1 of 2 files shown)


cchriste (approved):
Most of this is witchcraft to me, but the biggest question I have is: is this still using Ubuntu 16? Regardless, I'm approving this PR since I presume it works. At the very least it doesn't not work.

cchriste:
> Most of this is witchcraft to me, but the biggest question I have is: is this still using Ubuntu 16? Regardless, I'm approving this PR since I presume it works. At the very least it doesn't not work.

I could be wrong about that last part: where are the checks? I don't see any and it's not a draft PR.

akenmorris:
GitHub had a problem and would run anything at the time:

“ Error : .github#L1
An unexpected error has occurred and we've been automatically notified. Errors are sometimes temporary, so please try again. If the problem persists, please check whether the Actions service is operating normally at https://githubstatus.com. If not, please try again once the outage has been resolved. Should you need to contact Support, please visit https://support.github.com/contact and include request ID: 14e82d5e461300181844ba0f46dcbf0a”

akenmorris:
Looks like they are still having an outage:

https://www.githubstatus.com/

akenmorris:
To answer your other question, it’s still building on Ubuntu 16, just now inside a docker container running on ubuntu-latest. 

One nice thing is that you can build the docker image yourself, even on OSX and then you will have the same build environment that GH uses. Could be helpful for debugging build/deployment issues.

Issue #1192, opened 2021-04-02
Restructure the logic and computation of the evaluation commands 


iyerkrithika21:
Right now, the ellipsoid evaluate use case calculates the generalization and specificity values for just one mode. The output is only printed on the command line. It would be useful if the use case calculated the generalization and specificity for all the modes and save the values in text files like the scree.txt.

The computation needs to be revisited.

iyerkrithika21:
@sheryjoe, does this seem reasonable to you?

akenmorris:
Keep in mind this may take many hours on larger size models.  Some progress output is likely needed.

medakk:
> Keep in mind this may take many hours on larger size models.  Some progress output is likely needed.

Indeed. [tqdm](https://github.com/tqdm/tqdm) perhaps

@iyerkrithika21 I agree that this would be useful. do you know how long it takes to run it for all modes on the ellipsoid_evaluate use case? 

+a simple optimization that could be done is to just run them in parallel ie: spawn multiple processes in EvaluationUtils.py one for each mode at the same time

iyerkrithika21:
> @iyerkrithika21 I agree that this would be useful. do you know how long it takes to run it for all modes on the ellipsoid_evaluate use case?

I will try to run it and update the time here.

Another small modification, for the scree plot, can we display the plot, save it, and then close the plot.? Right now the code freezes until the user manually closes the plot window.

medakk:
> Another small modification, for the scree plot, can we display the plot, save it, and then close the plot.? Right now the code freezes until the user manually closes the plot window.

I preferred having it open to see the result, but that was for a single mode(similar to how it opens studio to show the reconstructions). Saving it instead sounds reasonable. Should entail changing a `plt.show()` somewhere to a `plt.savefig(...)`

+Eigen has a [BDCSVD](https://eigen.tuxfamily.org/dox/group__SVD__Module.html) implementation, which claims to be a faster, but less accurate way to compute SVD. Perhaps we can try replacing all instances of JacobiSVD in [ShapeEvaluation.cpp](https://github.com/SCIInstitute/ShapeWorks/blob/76a411824b9225b85a2c97efbdfd735a66f22439/Libs/Particles/ShapeEvaluation.cpp) with BDCSVD and see if it makes a difference

medakk:
There are some inefficiencies in the ShapeEvaluation because it assumes we only need a single mode.

[...]

PR #1421, opened 2021-08-18
Studio Stats

### LINKED ISSUE #1410: Group difference visualization problem
group_diff_bug

In this image, the whole thing should be blue.  I believe what's happening is that when the particle is exactly on a vertex, the distance to that point is zero and a division by zero gives an infinite value.

akenmorris:
This PR adds a number of new things:

1) Under the hood, a new asynchronous job scheduling system with new support for arbitrary python tasks.  DeepSSM code migrated over

2) Group difference p-value displays (a python task)

3) Address #1192 with new shape evaluation code that is much faster

4) New asynchronous tasks for compactness, generalization, and specificity.  Each run in parallel with separate progress bars under the Metrics panel

5) Fix #1410 - Group difference visualization problem

6) New plots based on JKQtPlotter

Screen Shot 2021-08-17 at 12 00 00 AM


iyerkrithika21:
Was able to test the new shape evaluations commands and plots. 

But facing this error when trying to click on the Pvalues button in Studio :
image

This test data is not in the ShapeWorks folder but I ran `source ./devenv.sh ./build/bin` with the correct build folder before launching studio. 

From the test data folder if I try to import `shapeworks.utils`, it works fine. 
```
python
Python 3.7.8 | packaged by conda-forge | (default, Nov 27 2020, 19:24:58) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import shapeworks as sw
>>> sw.utils.postive_factors(2)
[1, 2]

```

akenmorris:
@iyerkrithika21 , I've updated the devenv scripts, try that again and see if it works.

iyerkrithika21:
@akenmorris, the p-values button works correctly now!

iyerkrithika21 (approved):

Issue #1389, opened 2021-07-07
Incorrect limitation to length of num_planes_per_input parameter for multiple domains


patkins:
The parameter `num_planes_per_input` seems to be restricted to a length of length(`inputs`) / `domains_per_shape` (see comment in #910), while it seems that it is expected to be the length of `inputs`.

Trial 1: My first thought was that each cutting plane was then used for the whole shape (including multiple domains, here n=2), however this resulted in the cutting planes getting applied one by one to each input (leaving the second half of the inputs without cutting planes and two different planes being applied within a shape).  
Trial 2: I then tried duplicating each cutting plane in the list, such that the number of `cutting_planes` matched the number of `inputs`. But the result was the same (although since the planes were duplicated, the same plane was applied to both domains for the first half of the shapes). 
Trial 3: I then used 2 `num_planes_per_input` and the duplicated list of planes from Trial 2, however this resulted in similar results to Trial 1

Sample data can be found in my folder here: /home/sci/atkins/SWIssues/corticalThickness/multipleDomain/

patkins:
I think @HeavenlyBerserker was working on cutting planes, so maybe this is relevant.

HeavenlyBerserker:
Hi @patkins , thanks for bringing this to my attention! The structure for constraints input was written before my time and I have intended to change it since I find it confusing as well. This was also implemented before "multiple domains" was working, so I appreciate any feedback to make this functionality easier to use.

The current structure stores domains in a single vector/list/array. In your case, where you have 4 inputs, 2 domains per input:
```
item 0: input_0_domain_0 aligned/dod_n02_R_femur_aligned.vtk
item 1: input_0_domain_1 aligned/dod_n02_R_femur_trabec_aligned.vtk
item 2: input_1_domain_0 aligned/dod_n09_R_femur_aligned.vtk
item 3: input_1_domain_1 aligned/dod_n09_R_femur_trabec_aligned.vtk
item 4: input_2_domain_0 aligned/dod_n10_L_femur_aligned.vtk
item 5: input_2_domain_1 aligned/dod_n10_L_femur_trabec_aligned.vtk
item 6: input_3_domain_0 aligned/dod_n11_L_femur_aligned.vtk
item 7: input_3_domain_1 aligned/dod_n11_L_femur_trabec_aligned.vtk
```
[...]

Issue #1408, opened 2021-08-02
Refactor vtkmeshwrapper such that low level mesh operations are in the mesh class


HeavenlyBerserker:
vtkmeshwrapper does not use any mesh class functionality, leading to the possibility of redundant functions in libs/mesh/mesh and libs/optimize/ParticleSystem/vtkmeshwrapper.

cchriste:
I think removing trimeshwrapper was also part of this, sadly. 
But adding functions to Mesh that can give you a TriMesh was also discussed if needed.

akenmorris:
TriMeshWrapper can be removed at this point.

HeavenlyBerserker:
Oh, can we hold this off until after the next release? I'm working on constraints and making tons of modifications and I don't wanna undo anything you guys refactor. Shireen told me to make this issue and not assign anybody as of yet. Sorry, I should have mentioned that.

cchriste:
> Oh, can we hold this off until after the next release? I'm working on constraints and making tons of modifications and I don't wanna undo anything you guys refactor. Shireen told me to make this issue and not assign anybody as of yet. Sorry, I should have mentioned that.

What kind of changes are you making? Do any of them related to issue #935? It is meant to create an independent Field class to be used by Meshes. The main question I have is whether Fields composed of 1, 3, and 9-component tuples are reasonable or if we would prefer a more generalized container, requiring a more complicated interface.

cchriste:
@HeavenlyBerserker Can you please answer the question above about the use of fields in the MeshWrappers? Thanks!

cchriste:
This is a reasonable place to start to understand how and where things should be split. @archanasri and I have been working on the geodesic distance function, but there are other components of these classes. @HeavenlyBerserker can you please share a summary (bullet list) of the components/features of these wrapper classes? And also please confirm we can dump TriMeshWrapper.

So far I have:
- compute geodesic distance on a mesh
- ?

HeavenlyBerserker:
Firstly, we can dump Trimesh for sue. Karthik tried it out and was not working for our purposes. All the functionality that is used is in VtkMesh.

I only added some methods to mesh that are needed for free-form constraint operation and eventually visualizations. I have not touched many of the other methods. Please see #1422 for more information.

[...]

PR #1422, opened 2021-08-19
Free Form Constraints

HeavenlyBerserker:
This PR adds FFCs support. FFCs are tested with included unit tests.

The unit case is the typical sphere unit test with a constraint that cuts the sphere like a tennis ball grove.
Screenshot from 2021-08-14 14-38-46

Also added a unit test with two domains, both spheres in different locations. The first sphere has one cutting plane and 25 sphere constraints, the second sphere has one cutting plane and one free form constraint. This also addresses Penny's issue detailed in #1389 .
Screenshot from 2021-08-30 11-55-56


akenmorris:
Adding @cchriste and @archanasri for review to look at the changes to Libs/Mesh

Libs/Groom/Groom.h
```diff
@@ -6,6 +6,10 @@
 #include <tbb/atomic.h>
 #include <tbb/mutex.h>
 
+#include <vtkContourFilter.h>
+#include <vtkPolyDataWriter.h>
```
akenmorris, line 10: Are these header includes necessary?

akenmorris:
@cchriste , @archanasri , can you have a look at the changes to the Mesh class since you are more familiar with it?

Libs/Groom/Groom.h
HeavenlyBerserker: They're not. I've removed them.

archanasri:
Some functions in Mesh.h are missing docstring

Libs/Mesh/Mesh.cpp
```diff
@@ -967,6 +989,599 @@ std::ostream& operator<<(std::ostream &os, const Mesh& mesh)
...
+
+
+        vtkSmartPointer<vtkDoubleArray> values = vtkSmartPointer<vtkDoubleArray>::New();
+        vtkSmartPointer<vtkDoubleArray> absvalues = setDistanceToBoundaryValueFieldForFFCs(values, points, boundaryVerts, inout, V, F, dom);
+
+        this->mesh->GetPointData()->SetActiveScalars("value");
+
+        std::vector<Eigen::Matrix3d> face_grad_ = this->setGradientFieldForFFCs(absvalues, V, F);
+
+        //this->visualizeVectorFieldForFFCs(values, face_grad_, V, F);
```
archanasri, line 1076: Why is it commented out?

Libs/Mesh/Mesh.cpp
```diff
@@ -967,6 +989,599 @@ std::ostream& operator<<(std::ostream &os, const Mesh& mesh)
...
+
+Eigen::Vector3d Mesh::GetFFCGradient(Eigen::Vector3d query){
+
+    locator->BuildLocator();
+
+    double closestPoint[3];
+    vtkIdType cellId;
+    int subId;
+    double dist;
+    locator->FindClosestPoint(query.data(), closestPoint, cellId, subId, dist);
```
archanasri, line 1135: Have you tried using `Mesh.closestPointId` or `Mesh.closestPoint`?

Libs/Mesh/Mesh.cpp
```diff
@@ -677,7 +677,7 @@ std::vector<std::string> Mesh::getFieldNames() const
   return fields;
 }
 
-Mesh& Mesh::setField(std::string name, Array array)
+Mesh& Mesh::setField(std::string name, Array array, bool multi)
```
archanasri, line 680: What is done with the multi parameter?

Libs/Mesh/Mesh.cpp
```diff
@@ -696,6 +696,28 @@ Mesh& Mesh::setField(std::string name, Array array)
...
+{
+  if (!array)
+    throw std::invalid_argument("Invalid array.");
+
+  if (name.empty())
+    throw std::invalid_argument("Provide name for the new field");
+
+  int numVertices = mesh->GetPoints()->GetNumberOfPoints();
+  if (array->GetNumberOfTuples() != numVertices) {
+    std::cerr << "WARNING: Added a mesh field with a different number of elements than points\n";
```
archanasri, line 709: This function is supposed to add a field for faces. So the number of faces needed to be considered

Libs/Mesh/Mesh.cpp
```diff
@@ -967,6 +989,599 @@ std::ostream& operator<<(std::ostream &os, const Mesh& mesh)
...
+            Eigen::Vector3d pt = boundaries[bound][i];
+            double ptdob[3] = {pt[0],pt[1],pt[2]};
+            vtkIdType ptid = locator->FindClosestPoint(ptdob);
+            mesh->GetPoint(ptid, ptdob);
+
+            // Add first point in boundary
+            if(i == 0){
+                lastId = ptid;
+                double pathpt[3];
+                mesh->GetPoint(ptid, pathpt);
```
archanasri, line 1025: Mesh.getPoint can be used

Libs/Mesh/Mesh.cpp
```diff
@@ -967,6 +989,599 @@ std::ostream& operator<<(std::ostream &os, const Mesh& mesh)
...
+                //std::cout << pt[0] << " " << pt[1] << " " << pt[2] << " -> " << ptdob[0] << " " << ptdob[1] << " " << ptdob[2] << std::endl;
+                // Add points in path
+                dijkstra->SetStartVertex(lastId);
+                dijkstra->SetEndVertex(ptid);
+                dijkstra->Update();
+                vtkSmartPointer<vtkIdList> idL = dijkstra->GetIdList();
+                for(size_t j = 1; j < idL->GetNumberOfIds(); j++){
+                    vtkIdType id = idL->GetId(j);
+                    double pathpt[3];
+                    mesh->GetPoint(id, pathpt);
```
archanasri, line 1040: Mesh.getPoint can be used

Libs/Mesh/Mesh.cpp
```diff
@@ -967,6 +989,599 @@ std::ostream& operator<<(std::ostream &os, const Mesh& mesh)
...
+        select->SetInputData(this->mesh);
+        select->GenerateSelectionScalarsOn();
+        select->SetSelectionModeToLargestRegion();
+
+        // Clipping mesh
+        vtkSmartPointer<vtkClipPolyData> selectclip = vtkClipPolyData::New();
+        selectclip->SetInputConnection(select->GetOutputPort());
+        selectclip->SetValue(0.0);
+
+        selectclip->Update();
```
archanasri, line 1061: Mesh.clip can be used

Libs/Mesh/Mesh.cpp
```diff
@@ -967,6 +989,599 @@ std::ostream& operator<<(std::ostream &os, const Mesh& mesh)
...
+
+double Mesh::GetFFCValue(Eigen::Vector3d query){
+    locator->BuildLocator();
+
+    double closestPoint[3];
+    vtkIdType cellId;
+    int subId;
+    double dist;
+    locator->FindClosestPoint(query.data(), closestPoint, cellId, subId, dist);
+
```
archanasri, line 1111: Mesh.closestPoint can be used

archanasri (changes requested):

Libs/Mesh/Mesh.cpp
HeavenlyBerserker: Because we want functionality to visualize vector fields, but don't want them to show for a user.

Libs/Mesh/Mesh.cpp
HeavenlyBerserker: These functions create a locator every time they run. I don't think it's very efficient for me to use these since I query this a lot. I also don't think a locator should be created for these functions, since they need to be general enough.

Libs/Mesh/Mesh.cpp
HeavenlyBerserker: There used to be a warning if your tuples were more than 1. This was an optional parameter to suppress this warning. I've removed it though.

Libs/Mesh/Mesh.cpp
archanasri: Okay. You can add this as a comment

Libs/Mesh/Mesh.cpp
HeavenlyBerserker: The input to Mesh.clip is a plane. I'm not clipping by a plane.

Libs/Mesh/Mesh.cpp
HeavenlyBerserker: Same as above. I want my own member locator that I can reuse.

Libs/Optimize/OptimizeParameterFile.cpp
```diff
@@ -918,7 +922,7 @@ int OptimizeParameterFile::get_num_inputs(TiXmlHandle* docHandle)
     throw std::invalid_argument(message);
   }
 
-  return inputs.size() / domains_per_shape;
+  return inputs.size(); // domains_per_shape;
```
akenmorris, line 925: Is this intentional?  Changing the output of get_num_inputs to be the total number of domains rather than the number of samples?

Libs/Groom/Groom.h
```diff
@@ -6,6 +6,8 @@
 #include <tbb/atomic.h>
 #include <tbb/mutex.h>
 
+#include <vtkDecimatePro.h>
```
akenmorris, line 9: Groom.h needs vtkDecimatePro?

akenmorris (changes requested):
Are we only doing the free-form constraints for Images because you can just clip meshes?  Is that right?

I would suggest using the Image library to create the Mesh objects from the images.  It appears that you are duplicating the code of generating a VTK mesh from an ITK image and then store this vtkSmartPointer in Sampler::m_meshes, but the only usage of this is to create a Mesh object anyway.  You should be able to use Image(image).toMesh(0.0) and just store that.

Libs/Optimize/OptimizeParameterFile.cpp
HeavenlyBerserker: Yes, this is what fixes Penny's issue. This function is only used for the constraints input functions, so it doesn't affect anything else. Also, you should expect get_num_ inputs to return the number of inputs (as enclosed in the input tabs in the xml) and not the number of shapes.

akenmorris: Sounds good.  Let's get rid of the comment "// domains_per_shape;"

Libs/Mesh/Mesh.h
```diff
@@ -3,11 +3,35 @@
...
+#include <geometrycentral/surface/surface_mesh.h>
+#include "geometrycentral/surface/heat_method_distance.h"
+#include "geometrycentral/surface/meshio.h"
+
+#include <vtkLookupTable.h>
+#include <vtkArrowSource.h>
+#include <vtkNamedColors.h>
+#include <vtkPolyDataMapper.h>
+#include <vtkRenderWindow.h>
+#include <vtkRenderWindowInteractor.h>
```
akenmorris, line 32: Why does Mesh.h need vtkRenderWindow and vtkRenderWindowInteractor?  Mesh.h is included in a lot of places, so extra unneeded headers slow down compilation of the entire codebase.

Libs/Mesh/Mesh.cpp
```diff
@@ -967,6 +988,600 @@ std::ostream& operator<<(std::ostream &os, const Mesh& mesh)
...
+    }
+
+    this->setField("Gradient", vf);
+
+    this->mesh->GetCellData()->AddArray(vff);
+
+    return face_grad_;
+}
+
+void Mesh::visualizeVectorFieldForFFCs(vtkSmartPointer<vtkDoubleArray> values, std::vector<Eigen::Matrix3d> face_grad_, Eigen::MatrixXd V, Eigen::MatrixXi F){
```
akenmorris, line 1379: I think this belongs elsewhere and not in the Mesh class.  I don't think the Mesh class needs the ability to visualize itself.  This could go in MeshUtils or a new MeshVizUtils or something.

Libs/Mesh/Mesh.cpp
```diff
@@ -1038,6 +1096,413 @@ std::ostream& operator<<(std::ostream &os, const Mesh& mesh)
...
+  double pcoords[3];
+  double dist2;
+  Eigen::Vector3d bary;
+  this->mesh->GetCell(face)->EvaluatePosition(pt.data(), closest, sub_id, pcoords, dist2, bary.data());
+  return bary;
+}
+
+double Mesh::getFFCValue(Eigen::Vector3d query)
+{
+  locator->BuildLocator();
```
akenmorris, line 1214: Should this be calling BuildLocator for every getFFCValue?  Will BuildLocatorIfNeeded work and result in less rebuilding of the locator?

akenmorris (approved):

akenmorris:
```
Verbosity 0: This will be the only output on your screen, unless there are any errors. Increase the verbosity if needed.
Splitting mesh FFC for domain 0 shape 0 with query point      48      48 40.9668
Number of boundary vertices 144
Loading eval values for FFCs in domain 0
Adding free-form constraint to domain 0
```

These messages should only be written if the verbosity level is at least above zero, otherwise the first message is wrong.

Libs/Mesh/Mesh.h
```diff
@@ -144,6 +141,9 @@ class Mesh
   /// sets the given field for points with array (*does not copy array's values)
   Mesh& setField(std::string name, Array array);
 
+  /// sets the given field for faces with array (*does not copy array's values)
+  Mesh& setFieldForFaces(std::string name, Array array);
```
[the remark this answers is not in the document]
HeavenlyBerserker: It is the same as setField, so it has the same implementation. It displays a warning or raises an exception when there are discrepancies in the scope.

Libs/Mesh/Mesh.h
```diff
@@ -161,6 +161,9 @@ class Mesh
   /// gets the value at the given index of field
   double getFieldValue(const std::string& name, int idx) const;
 
+  /// gets the multi value at the given index of field
+  Eigen::VectorXd getMultiFieldValue(const std::string& name, int idx) const;
```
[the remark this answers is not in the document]
HeavenlyBerserker: It returns a multifield tuple at index idx in the form of an Eigen::VectorXd. Could come up in the fields discussion. I'd like to finish the discussion of whether we need fields before putting small stuff like this there.

Libs/Mesh/Mesh.cpp
```diff
@@ -1038,6 +1096,411 @@ std::ostream& operator<<(std::ostream &os, const Mesh& mesh)
...
+
+  // Write mesh for debug purposes
+//    std::string fnin = "dev/mesh_" + std::to_string(dom) + "_" + std::to_string(num) + "_in.vtk";
+//    this->write(fnin);
+
+  return true;
+}
+
+// Copied directly from Meshwrapper
```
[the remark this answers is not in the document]
HeavenlyBerserker: Moving is difficult since both mesh and vtkmeshwrappers are wrappers for a vtkpolydata mesh. If I move it, I need to refactor out the whole of vtkmeshwrappers. This is why #1408 exists in the first place.

Libs/Optimize/ParticleSystem/Sampler.h
```diff
@@ -27,6 +26,11 @@
 
 #include "TriMesh.h"
 
+#include "Mesh.h"
+
+// Uncomment to enable FFC visualizations
```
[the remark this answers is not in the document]
HeavenlyBerserker: Good point! I'll do a define.

Libs/Mesh/Mesh.cpp
```diff
@@ -1038,6 +1096,411 @@ std::ostream& operator<<(std::ostream &os, const Mesh& mesh)
   return os;
 }
 
+//Splits the mesh for FFCs by setting scalar and vector fields
```
cchriste, line 1099: nit-picking, but coding standards ask that documentation should be a triple-slash-prefixed comment in the header. That way it is reflected consistently and within Doxygen API. Please go through the set of functions and make it consistent with the rest of the class. Thanks!

HeavenlyBerserker: Done

akenmorris (approved):

akenmorris:
Ok to merge?  @cchriste , @archanasri ?

cchriste:
Yup!

archanasri (approved):

Issue #1375, opened 2021-07-05
DeepSSM in Studio - Documentation


jadie1:
Add documentation for running DeepSSM in Studio

PR #1425, opened 2021-08-19
DeepSSM in Studio Doc

jadie1:
Adding "DeepSSM in Studio." under the Studio tab in the documentation for issue #1375

@riddhishb Since you haven't run DeepSSM in studio yet could you possibly try following the documentation and running the example to see if we need to add anything?


akenmorris:
Should this be a zip, supershapes.zip or a directory like the others?  Also, it didn't get tracked by git-lfs, so this is increasing out checkout by 100MB in perpetuity unless we squash it.

Issue #1391, opened 2021-07-13
Change evaluation use case  to use the restructured evaluation commands


iyerkrithika21:
The calculation of specificity and generalization will be changed to make it more efficient and give out statistics for all modes of variation. The use case will have to be changed such that all three plots - specificity, compactness, generalization can be plotted.
Related to #1192

PR #1426, opened 2021-08-19
Restructured ellipsoid evaluate use case

iyerkrithika21:
Related to #1391


jadie1:
Just want to check - when it opens the best and worst reconstruction in studio it is not supposed to show the evaluation plots right? The use case is just using studio to show the sample?

For example:
image

iyerkrithika21:
> 
> 
> Just want to check - when it opens the best and worst reconstruction in studio it is not supposed to show the evaluation plots right? The use case is just using studio to show the sample?
> 
This is correct. Studio calculates the evaluation metrics based on the data being loaded. In this case, it is just one shape so the plots are right.

jadie1 (approved):

akenmorris (approved):

PR #1427, opened 2021-08-21
Image initialization and sharing without copying from/to numpy arrays in Python

cchriste:


cchriste:
This is first pass ready and reviews will be welcome. (@jadie in particular, please make sure the modifications to loaders.py and the other use case tests seem correct.)

archanasri (approved):

jadie1 (approved):
Seems to be functioning correctly in loaders.py and shape cohort generation

Issue #1300, opened 2021-06-15
Subset download for femur needs to include "R" samples that need reflection


akenmorris:
The reflection code sections are not included in the automated tests since none of the subset needs reflection.  To increase test coverage, we need at least one sample to be an "R" that needs reflection.

We don't need this for 6.1 as we are manually testing and fixing these code paths now, this is just something to do afterward.

iyerkrithika21:
Instead of downloading the right side, in the case of a tiny test, reflect the left femurs.

Issue #534, opened 2020-06-02
RunUseCase mesh/image mode


0leks:

sheryjoe:
The use case should be detect whether to use <data>/meshes or the isosurface of distance transforms from groomed <data>/segmentations

akenmorris:
I was suggesting that we provide both paths for each example

sheryjoe:
Agreed.

sheryjoe:
> > The use case should be detect whether to use /meshes or the isosurface of distance transforms from groomed /segmentations
> 
> What do you mean? The isosurface should be extracted by `shapeworks optimize` when running on distance transform data but is configured to mesh mode?

Extracting meshes and maybe decimating them using PreviewCmd can be a grooming step. shapeworks optimize will work on extracted meshes in this case.

sheryjoe:
let's consider this when we restructure running use cases with native python apis.

PR #1434, opened 2021-08-30
Mesh/Image use case mode

jadie1:
This PR addresses issues #534 (Image/Mesh mode) and #1300 (reflection in femur tiny test)

To test #534 run uses cases with --mesh_mode (for example if running the ellipsoid use case with --mesh_mode it will run ellipsoid_mesh) 

To test #1300 run femur tiny test and note that the reflection is no longer skipped


iyerkrithika21 (approved):

jadie1:
Closing this PR because we are updating issue #534 so that mesh_mode indicates optimization should happen on meshes generated from distance ransforms

PR #1445, opened 2021-09-01
Use case mesh mode

jadie1:
Adding mesh mode option to use cases for issue #534 
The use cases which run optimization on distance transforms (ellipsoid, ellipsoid_cut, ellipsoid_multiple_domain, femur, femur_cut, and left_atrium) can now be run with the --mesh_mode tag and meshes will be created and optimized on. 

Also adds reflection to femur tiny tests for issue #1300


Examples/Python/ellipsoid.py
```diff
@@ -240,6 +240,20 @@ def Run_Pipeline(args):
...
+        # Make output folder
+        mesh_dir = groom_dir + 'meshes/'
+        if not os.path.exists(mesh_dir):
+            os.mkdir(mesh_dir)
+        # Convert distance transforms to meshes
+        mesh_files = []
+        for dt, shape_name in zip(shape_seg_list, shape_names):
+            print('Ceating mesh from: ' + shape_name)
+            dt.toMesh(0).write(mesh_dir + shape_name + ".ply")
+            mesh_files.append(mesh_dir + shape_name + ".ply")
```
iyerkrithika21, line 254: This gives a warning:
```WARNING: Trying to save mesh with new field. Only vtk and vtp files save associated fields
```

iyerkrithika21:
Other than the minor comment about the file format, the `--mesh_mode` runs correctly for the intended use cases.

Examples/Python/ellipsoid.py
```diff
@@ -240,6 +240,20 @@ def Run_Pipeline(args):
         dt_files = sw.utils.save_images(groom_dir + 'distance_transforms/', shape_seg_list,
                                         shape_names, extension='nrrd', compressed=False, verbose=True)
 
+    # If running optimzation in mesh mode, convert distance transforms to meshes
```
cchriste, line 243: Please create a helper function for this and the identical instances of it elsewhere in this PR.

Examples/Python/ellipsoid.py
jadie1: I am not seeing that warning but I will switch it to vtk

Examples/Python/ellipsoid.py
jadie1: I thought about doing this... Ultimately this is kind of part of the grooming process which we decided to make transparent when we switched to pybind and got rid of groom utils. I could make a helper function in the python package but this adds a layer of abstraction and makes the use cases less readable... what do you think?

cchriste: I think that it's complex enough that it warrants a single function.

Examples/Python/ellipsoid.py
```diff
@@ -283,9 +297,17 @@ def Run_Pipeline(args):
     # Run multiscale optimization unless single scale is specified
     if not args.use_single_scale:
         parameter_dictionary["use_shape_statistics_after"] = 32
-    # Execute the optimization function
+
+     # If in mesh mode optimize on meshes
```
cchriste, line 301: The `if args.mesh_mode` will be completely abstracted when the consolidation-into-a-single-function request is complete. There are six of these copies, each 15 lines long, split over two disparate sections in each test in which they're used, all reducible to a single function.

cchriste (changes requested):
I strongly suggest consolidating the six times two relatively complex sections if'd on args.mesh_mode into a single function.

Examples/Python/ellipsoid.py
```diff
@@ -283,9 +297,17 @@ def Run_Pipeline(args):
     # Run multiscale optimization unless single scale is specified
     if not args.use_single_scale:
         parameter_dictionary["use_shape_statistics_after"] = 32
-    # Execute the optimization function
+
```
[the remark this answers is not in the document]
jadie1: I made this change, I am still concerned this will be confusing for users because it is less readable. But it did greatly reduce the repetition so I added it

akenmorris:
Is this ready to merge?

jadie1:
I believe so, I made all of the requested changes

akenmorris:
@cchriste , ok to merge now?

cchriste (approved):
Looks good!

cchriste:
> @cchriste , ok to merge now?

Sorry, I thought I had approved this after our meeting on Tuesday. I think the checks were still failing then, so I was waiting.

PR #1449, opened 2021-09-02
Consolidation of cvdDecimate and fixElement

archanasri:
This PR includes the following:
1. Mesh::cvdDecimate
2. Mesh::fix is now Mesh::fixElement
3. Fix for memory leak issue caused by FECVDDecimationModifer
4. ShapeworksUtils::setRngSeed and ShapeworksUtils::rngSeed which helps in setting a seed for a random number generator
4. Associated command line utility
5. Associated python bindings
6. Associated tests (MeshTests, PythonTests, shapeworksTests)


Applications/shapeworks/Commands.cpp
```diff
@@ -44,6 +45,29 @@ bool Example::execute(const optparse::Values &options, SharedCommandData &shared
 }
 #endif
 
+///////////////////////////////////////////////////////////////////////////////
+// Seed
+///////////////////////////////////////////////////////////////////////////////
+void Seed::buildParser()
+{
+  const std::string prog = "seed";
+  const std::string desc = "sets the seed for random number generation";
```
cchriste, line 54: please add "(useful for debugging)" to the help string.

Applications/shapeworks/MeshCommands.cpp
```diff
@@ -261,6 +261,35 @@ bool Decimate::execute(const optparse::Values &options, SharedCommandData &share
   return sharedData.validMesh();
 }
 
+///////////////////////////////////////////////////////////////////////////////
+// CVDDecimate
+///////////////////////////////////////////////////////////////////////////////
+void CVDDecimate::buildParser()
+{
+  const std::string prog = "cvd-decimate";
+  const std::string desc = "applies cvd decimation filter";
```
cchriste, line 270: add "(centroidal voronoi diagram)" as with the other help and docs

Libs/Common/ShapeworksUtils.h
```diff
@@ -22,6 +22,12 @@ class ShapeworksUtils
   /// return elapsed time in seconds, optionally printing to console
   static double elapsed(time_point start, time_point end, bool print_elapsed = true);
 
+  /// returns random seed
```
cchriste, line 25: We probably want to replace this with the random number generator itself, since as written, it will use the same seed each time functions like `FECVDDecimationModifier::Initialize` are called.

The reasons for this in the first place are debugging and testing, which should perhaps also be documented. 

The downside to replacing this with the rng itself is that threaded calls to these functions will still not always be reproducible. Do we need both [for testing/debugging]?

Libs/Mesh/Mesh.cpp
```diff
@@ -238,6 +237,24 @@ Mesh &Mesh::decimate(double reduction, double angle, bool preserveTopology)
   return *this;
 }
 
+Mesh &Mesh::cvdDecimate(double percentage)
+{
+  FEVTKimport import;
+  std::shared_ptr<FEMesh> meshFE(import.Load(this->mesh));
```
cchriste, line 243: Yay! One less memory leak ftw 🏅

Libs/Mesh/PreviewMeshQC/FECVDDecimationModifier.cpp
```diff
@@ -127,9 +121,12 @@ bool FECVDDecimationModifier::Initialize(FEMesh* pm)
 		if (pm->Element(i).IsSelected()) selectedFaces.push_back(i);
 	}
 
+	// create random number generator
```
cchriste, line 124: Overall, this function seems much cleaner with const variables marked as such. But the comment above regarding using the same rng seed for each call is still an issue.

cchriste (commented):
The main issue here is the ability to reproduce results for testing or debugging using a rng / seed consistently. The (possible) reason not to use the same seed every time is that every time the function is called, even when not used for testing, the rng will produce the same values in the same order. 

On the testing side, even if the rng itself is provided in ShapeworksUtils, threaded calls to these functions may still results in different orders and thus not the same order of values and irreproducible results. Let's talk offline about an appropriate solution for this issue that enables genuinely (pseudo-)random rng values as well as reproducible values for testing and debugging.

cchriste:
@akenmorris any ideas why this is failing to build on Linux?
```
2021-09-02T18:44:06.0810697Z [ 53%] Linking CXX executable ../../bin/MeshFromDT
2021-09-02T18:44:08.6092441Z /usr/bin/ld: /github/home/install/lib/libvtkFiltersModeling-8.2.so.1: undefined reference to symbol '_ZN12vtkTransform3NewEv'
2021-09-02T18:44:08.6108942Z /github/home/install/lib/libvtkCommonTransforms-8.2.so.1: error adding symbols: DSO missing from command line
2021-09-02T18:44:08.6110158Z collect2: error: ld returned 1 exit status
2021-09-02T18:44:08.6125184Z Libs/Analyze/CMakeFiles/MeshFromDT.dir/build.make:340: recipe for target 'bin/MeshFromDT' failed
2021-09-02T18:44:08.6130193Z make[2]: *** [bin/MeshFromDT] Error 1
2021-09-02T18:44:08.6139037Z CMakeFiles/Makefile2:1075: recipe for target 'Libs/Analyze/CMakeFiles/MeshFromDT.dir/all' failed
2021-09-02T18:44:08.6143764Z make[1]: *** [Libs/Analyze/CMakeFiles/MeshFromDT.dir/all] Error 2
2021-09-02T18:44:08.6148631Z make[1]: *** Waiting for unfinished jobs....
```

Libs/Common/ShapeworksUtils.h
archanasri: I do think we need it for testing purposes

Applications/shapeworks/MeshCommands.cpp
archanasri: Done

Applications/shapeworks/Commands.cpp
archanasri: Done

akenmorris:
My guess is that it's an order of library linking issue like this:

https://newbedev.com/libpthread-so-0-error-adding-symbols-dso-missing-from-command-line

I've committed a change to move VTK to the end and we'll see if that fixes it.

akenmorris:
Looks like the problem was that Libs/Common had undeclared dependencies (ITK and VTK).

Libs/Mesh/PreviewMeshQC/FECVDDecimationModifier.cpp
archanasri: Done

Libs/Common/ShapeworksUtils.h
archanasri: Done

cchriste (approved):

cchriste:
I like the new rng in ShapeworksUtils, but it won't solve the issue of generating random numbers if threads are involved. Something to think about for the future. 

The use of shared ptrs and consts is also helping the code become more stable with fewer memory leaks. 👍

Issue #1197, opened 2021-04-06
Standardize 2D Supershapes Cohort Gen


jadie1:
Update the 2D supershapes code so that we can generate images and segmentations as we do for the ellipsoid and 3d supershapes

PR #1454, opened 2021-09-07
Contour image generation

jadie1:
This PR addresses issue #1197 - adding the option of seg and image generation for 2D supershapes generation.

To test:

- Add shapely to shapeworks conda environment
- Run pip install ShapeWorks/Python/ShapeCohortGenPackage/
- Generate 2D supershapes with images in python the same way we do 3D, for example:

> ```
> import ShapeCohortGen
> 
> out_dir = "Generated_Contours/"
> ss_generator = ShapeCohortGen.Supershapes2DCohortGenerator(out_dir)
> contourFiles = ss_generator.generate(3)
> segFiles = ss_generator.generate_segmentations()
> imageFiles = ss_generator.generate_images()
> ```


iyerkrithika21:
@jadie1 are the images expected to look this way?

jadie1:
@iyerkrithika21 thanks, it should work now

iyerkrithika21:
Image generation works fine now.
Should we include these functions in the shape cohort notebook?

jadie1:
> Should we include these functions in the shape cohort notebook?

I'm not sure, the ellipsoid joints are not in the notebook either, Maybe we should include a description of the other types of shapes you can generate but not walk through it. Since the process is the same regardless of the kind of shape, I think walking through just supershapes and ellipsoids is sufficient. What do you think?

iyerkrithika21:
> > Should we include these functions in the shape cohort notebook?
> 
> Maybe we should include a description of the other types of shapes you can generate but not walk through it. 
This sounds good

iyerkrithika21 (approved):

akenmorris (approved):

Issue #1448, opened 2021-09-02
Studio: Recent XML projects need to store path as well


akenmorris:
XML projects that show up in the recent projects in Studio generally can't be loaded because they rely on the current path in order to find files.  Studio needs to store the loading path as well.

akenmorris:
Completed in #1455

Issue #1401, opened 2021-07-28
Extra space in optimize params in studio


sheryjoe:
This is from Studio running on windows and maximized. Is the space between Number of particles and other params expected?

<img width="218" alt="image" src="https://user-images.githubusercontent.com/3253036/127360483-1111a58b-c1eb-4d32-8328-ba1ac6ef9c2f.png">

akenmorris:
I haven't been able to reproduce this so far.   I tried both ShapeWorks 6.0.0 and 6.1.0.  This is what the layout looks like for me on windows 10:

![studio_optimize_layout_win10](https://user-images.githubusercontent.com/1693349/127512853-a268ee1d-08ee-41a2-84be-185d7604f098.png)

akenmorris:
I thought perhaps this might happen if you load a project with multiple domains, which populates multiple particle boxes, and then load a single domain project, but for me this resulted in the same correct layout.

akenmorris:
I was able to reproduce this.  This happens when one of the other tabs has expanded the layout of the control panel.

<img width="399" alt="image" src="https://user-images.githubusercontent.com/1693349/131865171-bbc6d16f-fcfd-4b04-b08c-c58c7cf6ce21.png">

Issue #1359, opened 2021-06-28
Feature map color bar control


patkins:
The feature map color bar seems to adjust automatically, such that when two groups are compared that have different feature value ranges, the color bar is different for the two groups (0.6-7.2 for cam, 0.8-8.0 for control in the images below). This makes it difficult to compare feature values between groups. 
![image](https://user-images.githubusercontent.com/13870137/123690027-27e08000-d811-11eb-8df0-9f661a3ebff1.png)
![image](https://user-images.githubusercontent.com/13870137/123690003-1eefae80-d811-11eb-9df9-3b29dddda794.png)

I would prefer the ability to choose the bounds of the color map, but it should at least be constant (or based on the range from the input data) if it cannot be altered by the user.

akenmorris:
Completed in #1455

Issue #1284, opened 2021-06-10
Feature Request - Multi-domain transparency


RichLisonbee:
Multi-domain allow for users to change the transparency of the individual domains. This will allow the user to see the space between the different domains to check for overlap or check model correspondence. Additionally, the colors of the correspondence particles become monochromatic after the second domain is added. This makes it difficult to compare correspondence particles when evaluating. Can these be randomized in a larger spectrum colormap?

These feature requests were discussed in the 6/9/21 ShapeWorks meeting for reference.

Thanks!

![image](https://user-images.githubusercontent.com/69816397/121568220-b3b67780-c9dc-11eb-9fba-b528476b75e2.png)

PR #1455, opened 2021-09-07
Studio Updates

akenmorris:

#1452 - Opacity controls for each domain
#1284 - Feature Request - Multi-domain transparency
#1359 - Add support for manual feature scalar range
#1448 - Studio: Recent XML projects need to store path as well
#1159 - Studio: Allow passing in initial points
#1401 - Extra space in optimize params in studio 

@patkins, Support for initial points is limited to the running of optimization, there is no display of these points yet.  To use this features, simply add a column with the prefix "landmarks_file_<name>" and these points will be used as the initial points to the optimizer.

For example:

image

### DIFF
--- a/Libs/Optimize/OptimizeParameters.cpp
+++ b/Libs/Optimize/OptimizeParameters.cpp
@@ -282,6 +282,17 @@ bool OptimizeParameters::set_up_optimize(Optimize* optimize)
     throw std::invalid_argument("No subjects to optimize");
   }
 
+  // landmarks/point files
+  std::vector<std::string> point_files;
+
+  for (auto s : subjects) {
+    auto landmarks = s->get_landmarks_filenames();
+    point_files.insert(std::end(point_files), std::begin(landmarks), std::end(landmarks));
+  }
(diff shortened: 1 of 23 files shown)


iyerkrithika21:
> I also could not find the transparency controls for multi-domain models.

@patkins this option is available is when you click on the drop-down arrow next to the ISO button at the bottom of the Studio window

akenmorris:
Yeah, the opacity controls are on the little dropdown like the glyph size and detail.

For the naming if <file> is your name, then the landmarks column will need to be landmarks_file_file.  I specified "file" in the name because I intend to store landmarks directly in the spreadsheet for landmarks creating within studio so as to not write out N more files.  Sorry if it's confusion.  I would do something like:


| shape_labrum      | landmarks_file_labrum |
| ----------- | ----------- |
| ...      | ...       |
| ...   | ...        |

iyerkrithika21 (approved):
I was able to test all the changes, and they seem to be working fine.

patkins:
I now see the opacity and the landmarks file seems to be importing correctly.

Issue #1314, opened 2021-06-16
Unclosable empty "Visualization Toolkit - OpenGL" windows popping up when running certain notebooks


HeavenlyBerserker:
Seriously, unclosable. Aaaah!!! Attempted the close all command and they're still there. Also, why are they empty?

![Screenshot from 2021-06-16 09-15-42](https://user-images.githubusercontent.com/11202328/122247142-63c82c80-ce84-11eb-963f-d017ed0f29d9.png)
![Screenshot from 2021-06-16 09-16-58](https://user-images.githubusercontent.com/11202328/122247159-662a8680-ce84-11eb-88c7-61903d1ec661.png)

cchriste:
These used to appear with the old version of pyvista. So likely one of two things:
1) old version of pyvista is still installed (can query this from conda command line or through module in python)
2) old way of calling pyvista (can look around the notebooks and/or google to see the new way; also the command line can provide worthwhile information about deprecated usage)
3) windows is different from the other platforms (yeah, I know said "of two things")

HeavenlyBerserker:
I tried with the three newest versions of pyvista: 0.29.1, 0.30.1, and 0.31.1. The conda_install.sh uses 0.30.1.

0.29.1 exhibits slightly different behavior; it still opens empty windows but closes them immediately. None of these are very useful.

I'm running on Ubuntu 20.04.2 LTS, btw, so I shouldn't get the issues Windows users get. @jadie1 , you run on Linux right? Have you gotten this issue before?

jadie1:
Yes this has been happening since we added the notebooks, the windows only close when you close the notebook. I check into if pyvista is aware of the issue.

riddhishb:
I am trying to figure this out currently, any updates on this @HeavenlyBerserker that I should know about?

iyerkrithika21:
@riddhishb, do we need to open an issue with pyvista for this problem, or can we do something from our side?

Issue #1113, opened 2021-03-04
pyvista viewpoint different than in videos (windows only?)


cchriste:
this screenshot shows the difference:
<img width="940" alt="2021-03-04 (1)" src="https://user-images.githubusercontent.com/1260945/110043359-7e76fb00-7d04-11eb-8a30-4836cae0e6fd.png">

sheryjoe:
@jadie1 @iyerkrithika21 with the refactored notebooks, is this still an issue?

iyerkrithika21:
This is still an issue when we visualize the data in grids

cchriste:
Since it's still functional I propose we push this back to 6.2.

jadie1:
Per our discussion, I will add a note to the notebooks that the videos were captured on linux/mac and might look different on Windows.

PR #1470, opened 2021-09-16
Notebook Visualization Fix

riddhishb:
This commit basically removes the interactive flag which was a misleading nomenclature and replaced it with a notebook flag in the pyvista plotters.

[A] When the notebook flag is False (this is the default now), it generates a render window and the visualization is performed in this window, you can close this window (tested on Linux) without affecting the rest of the code.

[B] When the notebook flag is True the rendering process is still done via OpenGL which requires an empty window that will pop open, but the visualization itself will be inline in the notebook:
This causes two problems and the suggestion is to never use this.
- Either you cannot close the window or if you close it then the notebook hangs
- Due to communication between the renderer and notebook the interactivity is very laggy

I have changed the backend plot functions and notebooks accordingly, and by default, it uses option [A].

Testing needed for Windows and Mac -- I suggest running [this notebook](https://github.com/SCIInstitute/ShapeWorks/blob/master/Examples/Python/notebooks/tutorials/getting-started-with-exploring-segmentations.ipynb)

Fixes issue #1314 and other lagging problems.

### DIFF
--- a/Examples/Python/notebooks/tutorials/getting-started-with-data-augmentation.ipynb
+++ b/Examples/Python/notebooks/tutorials/getting-started-with-data-augmentation.ipynb
@@ -319,7 +319,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "Python 3 (ipykernel)",
    "language": "python",
    "name": "python3"
   },
--- a/Examples/Python/notebooks/tutorials/getting-started-with-exploring-segmentations.ipynb
+++ b/Examples/Python/notebooks/tutorials/getting-started-with-exploring-segmentations.ipynb
@@ -181,7 +181,7 @@
    "source": [
     "# define parameters that controls the plotter\n",
     "use_same_window = False # plot using multiple rendering windows if false\n",
-    "is_interactive  = True  # to enable interactive plots\n",
+    "notebook        = False # True will enable the plots to lie inline\n",
     "show_borders    = True  # show borders for each rendering window\n",
     "shade_volumes   = True  # use shading when performing volume rendering\n",
     "color_map       = \"viridis\" # color map for volume rendering, e.g., 'bone', 'coolwarm', 'cool', 'viridis', 'magma'\n",
@@ -195,7 +195,8 @@
     "sw.plot_volumes(shapeSegList,    \n",
     "             volumeNames     = shapeNames, \n",
     "             use_same_window = use_same_window,\n",
-    "             is_interactive  = is_interactive, \n",
+    "             is_interactive  = is_interactive,\n",
+    "             notebook        = notebook,\n",
     "             show_borders    = show_borders,  \n",
     "             shade_volumes   = shade_volumes, \n",
     "             color_map       = color_map,\n",
@@ -543,7 +544,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "Python 3 (ipykernel)",
    "language": "python",
    "name": "python3"
   },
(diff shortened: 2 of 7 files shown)


jadie1 (approved):
I tested all the notebooks on Linux and it seems to work well. Having the output in the notebook was a little more intuitive but I think this is a better user experience. 

If someone tests on windows can you see if this resolves issue #1113

cchriste:
> If someone tests on windows can you see if this resolves issue #1113

booting it up now...

cchriste:
> If someone tests on windows can you see if this resolves issue #1113

It does not.

cchriste:
It works fine, but the notebooks now have missing images. Maybe this is just a coincidence and they were already missing. 
These should be [re-]added.

133708565-ac58fb17-ffaa-4ee1-8ac3-10e444ac257a_LI

cchriste:
> I tested all the notebooks on Linux and it seems to work well. Having the output in the notebook was a little more intuitive but I think this is a better user experience.

The part I find frustrating is that multiple outputs cannot be simultaneously displayed. Once upon a time I thought this was possible...

cchriste (changes requested):
Add missing images.

riddhishb:
Yeah, that annoyed me too, but unfortunately, I cannot add these images, as they are not actually images it's interactive blocks either it's detached from a notebook or in there. and unfortunately, those square boxes can be removed as far as I know

cchriste:
> Yeah, that annoyed me too, but unfortunately, I cannot add these images, as they are not actually images it's interactive blocks either it's detached from a notebook or in there. and unfortunately, those square boxes can be removed as far as I know

Clicking on the cell shows its contents. Looks like an image missing from the sci.utah.edu/~shapeworks directory. So what needs to be done is for the screenshots to be [re-created and] re-uploaded to this location. Anyone know where that is?
In the meantime, a scan through the notebooks to identify and recapture the missing material will help find what needs to be re-uploaded and even say what it should be named. (Please don't simply delete these lines. I know cuz' I was tempted to do the same :). 

image

riddhishb:
> > Yeah, that annoyed me too, but unfortunately, I cannot add these images, as they are not actually images it's interactive blocks either it's detached from a notebook or in there. and unfortunately, those square boxes can be removed as far as I know
> 
> Clicking on the cell shows its contents. Looks like an image missing from the sci.utah.edu/~shapeworks directory. So what needs to be done is for the screenshots to be [re-created and] re-uploaded to this location. Anyone know where that is?
> In the meantime, a scan through the notebooks to identify and recapture the missing material will help find what needs to be re-uploaded and even say what it should be named. (Please don't simply delete these lines. I know cuz' I was tempted to do the same :).
> 
> image

alright let me fix this.

cchriste:
Is it possible the links from the SCI server are only accessible from a restricted set of URLs? I'm looking at PR #1472 and there are links there that are not accessible from my localhost (ex: https://sci.utah.edu/~shapeworks/doc-resources/gifs/ellipsoid_mode1.gif). Because it's a big update, I'm not sure where/if that link was simply not there before, or where to look on the SCI server to find these. 

Can someone (@sheryjoe @jadie1 @archanasri @iyerkrithika21 @akenmorris) please tell us where to find the SCI resources directory? Here are images showing the missing resources in the other PR I mentioned:
Screen Shot 2021-09-20 at 1 56 26 PM

Screen Shot 2021-09-20 at 1 46 48 PM
Screen Shot 2021-09-20 at 1 48 47 PM

riddhishb:
Yeah I am facing the same issue navigating to a site with the resource gives me permission error.
https://sci.utah.edu/~shapeworks/doc-resources/pngs/shape_boundary.png

riddhishb:
I checked all the notebooks, @cchriste I think this can be merged now.

Issue #825, opened 2020-11-25
Update shapeworks.sw2vtkMesh to export mesh vertices, edges, and polys rather than read/write to disk


sheryjoe:
Similar to shapeworks.Image, we want to convert a shapeworks mesh to itk or vtk data structure to be visualized by itkwidgets, which supports both itk:Mesh and vtk::PolyData

This also will be very handy when visualizing multiple samples in a grid view using pyvista.

For now, a workaround would be saving the mesh/image and re-reading it for visualization.

sheryjoe:
Update for sw.Image: 
- Converting to itk image can be done by extracting a numpy array from a sw image and use itk.GetImageFromArray
- Converting to vtk image can be done by extracting a numpy array from a sw image and use pyvista.wrap 

These helper functions will be soon demonstrated in our notebook tutorials. These helper functions can be included in shapeworks __init__.py 

I will provide more updates on sw.Mesh later.

sheryjoe:
For sw.Mesh, we need access to mesh vertices and faces to construct itkMesh and vtkPolyData in python.

sheryjoe:
pyvista wrap function can wrap vtkPolyData and trimesh (python) objects. If we can get the vertices and faces as arrays from sw.Mesh, we can construct vtkPolydata in python and use pv.wrap for visualization.

sheryjoe:
Please, update notebooks to replace sw2vtk sw2itk helper functions.

cchriste:
This may simply not be possible from c++, but these helpers should be in the pymodule.
For now, they will be moved into pymodule (by @jadie1 and @iyerkrithika21)

sheryjoe:
@jadie1 @iyerkrithika21 please address this issue as you resolve #1173 and #865.

archanasri:
I don't think we need toItk anymore since we are not going to use itkwidgets.

sheryjoe:
@archanasri is this still an issue?

archanasri:
> pyvista wrap function can wrap vtkPolyData and trimesh (python) objects. If we can get the vertices and faces as arrays from sw.Mesh, we can construct vtkPolydata in python and use pv.wrap for visualization.

In order to construct the mesh after obtaining the two arrays, we need to import vtk or import trimesh. Is this something we want to do?
importing itk always led to more issues

cchriste:
[...]

PR #1473, opened 2021-09-16
Constructing PolyData in utils.sw2vtk instead of read/write

archanasri:
This PR fixes the issue in #825. Previously the sw2vtk() used to take a sw mesh and convert it to vtk mesh by writing to disk and reading it. Now it creates a polydata by obtaining the mesh's vertices and faces.

Testing: Run jupyter notebooks that use utils.sw2vtk


cchriste:
This is missing inclusion of a mesh's fields and unit tests to verify it's working. 
For example, the dataset loaded with "getting started with meshes" example notebook has one field, `Normals`.  Here are the results of the conversion before:
Screen Shot 2021-09-16 at 5 53 24 PM

...and here are the results now:
Screen Shot 2021-09-16 at 5 53 32 PM

Everything else looks fine. 

Note: I also added a fix for a break in the sw2vtkImage function by a previous PR.

Examples/Python/notebooks/tutorials/getting-started-with-meshes.ipynb
```diff
@@ -38,14 +38,7 @@
...
    "source": [
-    "# let's import shapeworks library to test whether shapeworks is now set\n",
-    "try:\n",
-    "    import shapeworks as sw\n",
-    "except ImportError:\n",
-    "    import shapeworks as sw\n",
-    "    print('ERROR: shapeworks library failed to import')\n",
-    "else:\n",
-    "    print('SUCCESS: shapeworks library is successfully imported!!!')"
+    "import shapeworks as sw"
```
cchriste, line 41: 👍

cchriste (approved):
Great!

PR #1475, opened 2021-09-17
Resample tests for dts, timing test and doc update for adding new tests

archanasri:
This PR makes the following updates:

1. Update documentation for adding PythonTests and shapeworksTests
2. Add resample tests for binary distance transforms
3. Adding timing profiling test to an optimize test


cchriste:
There are a lot of good changes in this PR, but the description doesn't match. Examples include adding timing to one of the optimize tests (awesome, let's add the rest!) and lack of any additions to documentation. Can this be updated? 

I'm also unclear about why antialiasing and binarization is used for resampling of a distance transform. Perhaps an explanatory comment will be useful, since my understanding is it's already antialiased and I feel especially disinclined to binarize the result. It also adds a bunch of additional operations to a unit test. So if it's unnecessary let's remove it, and it's it's necessary an explanation will help people who look at this test as an example.

sheryjoe:
> I'm also unclear about why antialiasing and binarization is used for resampling of a distance transform. Perhaps an explanatory comment will be useful, since my understanding is it's already antialiased and I feel especially disinclined to binarize the result. It also adds a bunch of additional operations to a unit test. So if it's unnecessary let's remove it, and it's it's necessary an explanation will help people who look at this test as an example.

 antialiasing and binarization are only needed if one wants to resample a given segmenation. I am not sure about the context here, but why are distance transforms resampled in the first place?

archanasri:
> There are a lot of good changes in this PR, but the description doesn't match. Examples include adding timing to one of the optimize tests (awesome, let's add the rest!) and lack of any additions to documentation. Can this be updated?
> 
> I'm also unclear about why antialiasing and binarization is used for resampling of a distance transform. Perhaps an explanatory comment will be useful, since my understanding is it's already antialiased and I feel especially disinclined to binarize the result. It also adds a bunch of additional operations to a unit test. So if it's unnecessary let's remove it, and it's it's necessary an explanation will help people who look at this test as an example.

The title has been updated to reflect the changes this PR makes.
I agree that antialiasing and binarizing the distance transform is unnecessary. So I have updated the tests to only resample the distance transforms.

cchriste (approved):
I hope we'll add timing to other tests and use cases. Maybe create an issue!

PR #1476, opened 2021-09-20
Improved performance via compiler flags

### LINKED ISSUE #1378: Revisit compiler optimizations options
We are currently using RelWithDebInfo defaults, but it seems this disables some optimizations that it doesn't need to.

See:

https://gitlab.kitware.com/cmake/cmake/-/issues/20812

akenmorris:
Resolves #1378

In particular, we switch to Release mode and add debug symbols.  The problem with CMake's default RelWithDebInfo is that it turns off a lot of optimizations.

For Windows, we now use for compilation:

/FS /Zi /GL /MD /O2 /Ob3 /DNDEBUG /EHsc

And for linking:

/LTCG /DEBUG

This results in at least 20% speedup over regular RelWithDebInfo (what we were using previously) and 10% over Release.  The main speedup over Release comes from the global optimization (/GL) and whole program optimization (/LTCG)

I haven't been able to match gcc under linux on the same machine, but I'm not convinced that it's possible.  GCC on linux is still another 20% faster than this best result.  Based on the descriptions of MSVC's shortcomings that I've read, the ShapeWorks codebase is almost adversarial for MSVC.  Reportedly it has trouble with inlining deeply nested templates, which are all over the place in SW.

Other suggested options I tried:

/fp:fast  - This failed the unit tests due to the numerical differences
/arch:AVX2 - This should supposedly speed up Eigen, but we're not generally bound there.  This caused the automated tests to fail on GH, even though they worked for me locally.  Using this option didn't result in any speedup for me and in theory would restrict the CPUs that the binary would work on.


cchriste:
Am I just supposed to build RelWithDebInfo?
Are there any timings tests that will be easy to execute? (such as some just added in #1475)

cchriste:
This is awesome! A couple examples confirm things are more than 20% faster on Windows.
Just be sure to rebuild dependencies (have some coffee ☕️).

NOTE: `--tiny_test` wasn't sufficient to conduct these tests, so #1480 is now formally requesting the `--no_gui` option be added to RunUseCase.

Issue #903, opened 2021-01-16
Add version of Image::getArray (and Mesh::getField) that only wraps data in numpy.ndarray rather than copying it


cchriste:
It's inefficient to copy an image's data in order to modify it with numpy and then reconstruct the image from the copied data (which probably copies it again). Instead, we can use this version with a capsule, except our capsule will **not** free the data when the numpy array goes out of scope since it's actually owned by the itk image. 

This will make operations like the ones @jadie1 is performing much more efficient. 
The seeming danger is the image being free'd and the numpy array trying to be used afterwards. So add a note to the function's documentation (to be defined as a lambda in ShapeworksPython.cpp).

```c++
#include <iostream>
#include <pybind11/pybind11.h>
#include <pybind11/numpy.h>

namespace py = pybind11;

PYBIND11_PLUGIN(numpywrap) {
    py::module m("numpywrap");
    m.def("f", []() {
        // Allocate and initialize some data; make this big so
        // we can see the impact on the process memory use:
        constexpr size_t size = 100*1000*1000;
        double *foo = new double[size];
        for (size_t i = 0; i < size; i++) {
            foo[i] = (double) i;
        }

        // Create a Python object that will free the allocated
        // memory when destroyed:
        py::capsule free_when_done(foo, [](void *f) {
            double *foo = reinterpret_cast<double *>(f);
            std::cerr << "Element [0] = " << foo[0] << "\n";
            std::cerr << "freeing memory @ " << f << "\n";
            delete[] foo;
        });

        return py::array_t<double>(
            {100, 1000, 1000}, // shape
            {1000*1000*8, 1000*8, 8}, // C-style contiguous strides for double
            foo, // the data pointer
            free_when_done); // numpy array references this parent
    });
    return m.ptr();
}
```
```python
>>> import numpywrap
>>> z = numpywrap.f()
>>> # the python process is now taking up a bit more than 800MB memory
>>> z[1,1,1]
1001001.0
>>> z[0,0,100]
100.0
>>> z[99,999,999]
99999999.0
>>> z[0,0,0] = 3.141592
>>> del z
Element [0] = 3.14159
freeing memory @ 0x7fd769f12010
>>> # python process memory size has dropped back down
```
From https://stackoverflow.com/questions/44659924/returning-numpy-arrays-via-pybind11

[...]

PR #1494, opened 2021-09-29
Mesh API for Geodesic Distance computation

archanasri:
This PR consolidates the following:
1. Geodesic distance between two points on the mesh
2. Geodesic distance between a point (landmark) and all vertices on the mesh
3. Geodesic distance between a set of points (curve) and all vertices on the mesh

The distance is computed using the heat method in VtkMeshWrapper.

Python bindings have been added
MeshTests, PythonTests, and shapeworksTests have been added


cchriste (commented):
So far, so good. Once the conflict is fixed and it's passing tests I'll take a closer look.

Libs/Mesh/Mesh.cpp
```diff
@@ -53,8 +54,6 @@
 #include <vtkAppendPolyData.h>
 #include <vtkCleanPolyData.h>
 #include <vtkNew.h>
-#include <vtkPolyData.h>
```
cchriste, line 56: You'll wish you hadn't removed these when you've consolidated VtkMeshWrapper. Out of curiosity, what informed you they were no longer necessary?

Libs/Mesh/Mesh.cpp
```diff
@@ -537,16 +535,52 @@ int Mesh::closestPointId(const Point3 point)
...
+  vtkSmartPointer<vtkDoubleArray> distance = vtkSmartPointer<vtkDoubleArray>::New();
+  distance->SetNumberOfComponents(1);
+  distance->SetNumberOfTuples(numPoints());
+  distance->SetName("GeodesicDistanceToLandmark");
+
+  VtkMeshWrapper wrap(this->mesh, true);
+
+  for (int i = 0; i < numPoints(); i++)
+  {
+    distance->SetValue(i, wrap.ComputeDistance(getPoint(i), -1, landmark, -1));
```
cchriste, line 555: tl;dr: `wrap.ComputeDistance(landmark, 0, getPoint(i), -1)`

This doesn't take advantage the MeshWrapper caching for the first point (or any others). I suggest passing landmark w/ idx=0 for the first point. From my perusal of the vtkMeshWrapper caching system, this should only allocate one item in the underlying cache, used by ComputeFaceAndWeights, and which still exists even though it's noted to have no performance advantage. So we want to only send idx 0 to minimize the "unused-but-still-used" cache, but still enable the first point cache, which is handled in a different way (in both ComputeDistance and IsWithinDistance).

Libs/Python/ShapeworksPython.cpp
```diff
@@ -1116,10 +1116,60 @@ PYBIND11_MODULE(shapeworks_py, m)
...
+  .def("geodesicDistance",
+       [](Mesh &mesh, const std::vector<double> p) -> decltype(auto) {
+          auto array = mesh.geodesicDistance(Point({p[0], p[1], p[2]}));
+          const auto shape = std::vector<size_t>{static_cast<unsigned long>(array->GetNumberOfTuples()),
+                                                 static_cast<unsigned long>(array->GetNumberOfComponents()),
+                                                 1};
+          auto vtkarr = vtkSmartPointer<vtkDoubleArray>(vtkDoubleArray::New());
+          vtkarr->SetNumberOfValues(array->GetNumberOfValues());
+
+          // LOTS of copying going on here, see github #903
```
cchriste, line 1132: thanks for putting a comment pointing to the issue that should resolve all this unnecessary copying.

Libs/Python/ShapeworksPython.cpp
```diff
@@ -1116,10 +1116,60 @@ PYBIND11_MODULE(shapeworks_py, m)
...
+       "landmark"_a)
+
+  .def("geodesicDistance",
+       [](Mesh &mesh, const std::vector<std::vector<double>> p) -> decltype(auto) {
+          std::vector<Point> points;
+          for (int i=0; i<p.size(); i++)
+          {
+            points.push_back(Point({p[i][0], p[i][0], p[i][2]}));
+          }
+          auto array = mesh.geodesicDistance(points);
```
cchriste, line 1150: Please create a function for this section (down to line 1163) since it's identical to the lambda above:
`py::array vtkToNumpy(vtkArray)`. 

In fact, I suggest `vtkToEigen` since what we have is a nComponents x nPoints matrix, i.e., `Eigen::MatrixXd(double, nComponents, nPoints)`, because it's actually much simpler to write and pybind11 takes care of moving it back and forth. I just implemented it for itk arrays in EigenUtils.h/cpp in the as-yet-unpushed branch that fixes Transforms. I made a simple copy version for that, and if you do that you can still reference the "don't copy arrays passed between c++ and Python" issue above :). 

The more appropriate version should use `Eigen::Map<...>` [as described here](https://eigen.tuxfamily.org/dox/group__TutorialMapClass.html), and as used in the other functions in EigenUtils.h. 

I'm copying this comment to the issue #903 if you just want to make a simple wrapper of this duplicated code in these two lambdas for now since it works.

Libs/Python/ShapeworksPython.cpp
```diff
@@ -1116,10 +1116,60 @@ PYBIND11_MODULE(shapeworks_py, m)
...
+          return py::array(py::dtype::of<double>(),
+                         shape,
+                         vtkarr->GetVoidPointer(0));          // copy2
+       },
+       "computes geodesic distance between a set of points (curve) and all vertices on mesh",
+       "curve"_a)
+
+  .def("distance",
+       &Mesh::distance,
+       "computes surface to surface distance",
```
cchriste, line 1170: I suggest we tell the user in the help text the methods of computing distance that can be used.
I also recommend changing their Python enums to use CamelCase just as all the other enums and functions here. 
i.e., PointToPoint and CellToCell rather than POINT_TO_POINT and CELL_TO_CELL (around line 950).

Libs/Python/ShapeworksPython.cpp
```diff
@@ -1204,7 +1254,7 @@ PYBIND11_MODULE(shapeworks_py, m)
 
   //TODO: See github issue #966
   .def("setField",
-       [](Mesh &mesh, std::vector<double>& v, std::string name) -> decltype(auto) {
+       [](Mesh &mesh, std::string name, std::vector<double>& v) -> decltype(auto) {
```
cchriste, line 1257: Please change the GitHub issue to #1495. Also on line 1268 and anywhere else it's mentioned in this file. Details for why are in a previous comment. Thanks!

Libs/Python/ShapeworksPython.cpp
```diff
@@ -1213,11 +1263,11 @@ PYBIND11_MODULE(shapeworks_py, m)
          return mesh.setField(name, arr);
        },
        "sets the given field for points with array",
-       "array"_a, "name"_a)
+       "name"_a, "array"_a)
 
   //TODO: See github issue #966
   .def("setField",
-       [](Mesh &mesh, std::vector<std::vector<double>>& v, std::string name) -> decltype(auto) {
+       [](Mesh &mesh, std::string name, std::vector<std::vector<double>>& v) -> decltype(auto) {
```
cchriste, line 1270: I like the name/value being reversed for this function (both here and in Mesh).

Testing/PythonTests/geodesic.py
```diff
@@ -1,11 +1,40 @@
...
 from shapeworks import *
+import numpy as np
 
-def geodesicTest():
+def geodesicTest1():
   mesh = Mesh(os.environ["DATA"] + "/ellipsoid_0.ply")
   dist = mesh.geodesicDistance(10, 20)
 
-  return abs(dist - 1.0083) < 1e-4
+  return abs(dist - 1.64581) < 1e-4
```
cchriste, line 10: Can someone else please verify these updated distance results are correct? @sheryjoe @iyerkrithika21 @jadie1 Thanks!

Testing/data/geodesic.txt
```diff
@@ -1 +1 @@
-Geodesic Distance between two points: 1.00833
+Geodesic Distance between two points: 1.64581
```
cchriste, line 1: You don't have to fix it since it was already this way, but saving a text file to compare numeric output at least gets a frowny face here.

cchriste (changes requested):
Mostly looks good. Please see comments.

For bonus, measure the performance of the calls to compute the distance w/ MeshWrapper before and after swapping the arguments.

Libs/Mesh/Mesh.cpp
akenmorris: What's the "unused-but-still-used" cache?

cchriste: The function `GetTriangleForPoint` performs caching based on the provided index. It is called by `ComputeFaceAndWeights`. Since we are deliberately only passing -1 as the index for points other than the landmark (following my recommendation here, because first point has a different cache mechanism, explicitly associating the face id and barycentric weights with the first index passed), the mechanism in the `GetTriangleForPoint` function (an array of indices associated with the triangle in which they belong) will never be used. However, because it still allocates an array of points to associate with an index, using index 0 for the first point is the best way to limit this unnecessary allocation (short of reorganizing the mesh wrapper itself).

akenmorris: If you always pass idx=-1, the `particle_triangles_` cache won't ever be allocated.

cchriste:
Correct. But you still want to cache the first point, which is why you want
to swap the order of arguments and pass 0 as the first argument’s (the
source pt) cache index. The dest point can keep using -1, and should, since
according to the comments in the code, only caching the first point made a
difference.

Libs/Mesh/Mesh.cpp
akenmorris: > Correct. But you still want to cache the first point, which is why you want to swap the order of arguments and pass 0 as the first argument’s (the source pt) cache index. The dest point can keep using -1, and should, since according to the comments in the code, only caching the first point made a difference.

I agree that using '0' and landmark as the first point in this case is very slightly better, but I want to make sure the caches and the comments in the code are understood.  ComputeDistance has a separate **single** cached value of only the very last query such that `ComputeFaceAndWeights` and `GetTriangleForPoint` are not even called (nor is the `particle_triangles_` cache even used).  For optimal speed, Karthik was trying to avoid even the `IsInTriangle` check that happens when the `particle_triangles_` cache is used.  He didn't find that a secondary single value cache of the 2nd point was helpful because the optimizer never calls with the 2nd same point twice in a row.  This is true of the `Field Mesh::geodesicDistance(const Point3 landmark)` function as well.  But these comments are not referring to the cache in `GetTriangleForPoint`.  That cache is still heavily used by the optimizer for the 2nd point, but it will call `IsInTriangle` to confirm validity.

Keep in mind that if someone combines the Mesh and VtkMeshWrapper code, always using idx=0 and -1 will lead to incorrect results when a user's code makes repeated `Mesh::geodesicDistance` with different landmarks and these are always passed in as `idx=0` and the last query check matches even though the point's position is different.

This won't be a problem in this code now because it creates a new VtkMeshWrapper object for each call.  As a side note, have a look at the constructor of VtkMeshWrapper and see it is probably millions of times slower than ComputeDistance calls.  If you are concerned about performance here, you should keep around the VtkMeshWrapper.

cchriste: > Keep in mind that if someone combines the Mesh and VtkMeshWrapper code, always using idx=0 and -1 will lead to incorrect results when a user's code makes repeated Mesh::geodesicDistance with different landmarks and these are always passed in as idx=0 and the last query check matches even though the point's position is different.

Thank you. That's a very useful issue to identify since as written it will only work for the first computation of geodesic distance from a point and any other points will use invalid cache values. It should be changed to check the values of the point instead of its index. This addresses both issue at once: no `particle_triange_` cache will be allocated, and the omission of calling `GetTriangleForPoint` will remain. Reuse of idx will no longer accidentally result in an incorrect face id and barycentric coordinates being returned for the first cached point*.

*It's technically possible, but the same point will always return the same triangle and barycentric coordinates in that triangle for the same wrapped mesh, and there is no way to change which mesh is wrapped.

cchriste: The cut and pasted use of the "first point cache" is also troublesome, but can easily be addressed. 
The `particle_trangles_` cache is never used without providing idx, and (with the safety updates above) it will never be provided.

cchriste: The latest push dumps caching of the first point altogether. Turns out swapping the argument order speeds up computation by _more than 600%_, but it's due to internal caching of the vtkCellLocator (perhaps the same trick).

Libs/Python/ShapeworksPython.cpp
cchriste: Function created, and lots of duplicated code consolidated.

Libs/Python/ShapeworksPython.cpp
cchriste: done

Libs/Python/ShapeworksPython.cpp
cchriste: consolidated the many places this was duplicated into a helper function and added a comment there

cchriste:
This should be good to go. @akenmorris do you mind giving it a once over (once the tests pass; first round here will fail, but I just pushed the fix) just to be sure. Thanks!

cchriste:
Ah... happy Monday morning! I ran all the Python tests before I committed, but of course most of them don't actually verify their results (I'll do that now in a separate branch, just didn't want to rock the boat by combining tasks in one PR). On my machine it "just" fails (I ran it just now), but looks like we get a lucky seg fault in the actions. @archanasri and I are addressing this now and should have a fix shortly.

akenmorris:
Can you explain more about removing the geodesic last query cache?  The optimizer will now repeatedly call `ComputeFaceAndWeights` for the same point, which is a relatively expensive operation.  I ran `_thin_cavity_bean_` use case on master vs geodesic_exe branch and see about a 5% slowdown on geodesic_exe branch.

cchriste:
> Can you explain more about removing the geodesic last query cache? The optimizer will now repeatedly call `ComputeFaceAndWeights` for the same point, which is a relatively expensive operation. I ran `_thin_cavity_bean_` use case on master vs geodesic_exe branch and see about a 5% slowdown on geodesic_exe branch.

It just cached the first point to be queried, and all I did was clean it up (cache based on point coordinate instead of idx since as you mentioned, idx could be the same even for different points*). It didn't make a difference for computing geodesic distance, but it sounds like the thin cavity beam use case gains some benefit. I'll just put it back. The whole thing is a project begging for cleanup and simplification, and we've already got an issue to consolidate it all.

*it may be that the thin cavity beam use case is inappropriately using the same idx for different points inside the same wrapped mesh, in which case the removal of this cache fixed a bug. Cuz basically it seems that the vtk query already uses exactly this method for caching as evidenced by the incredible performance improvement when simply switching the argument order.

cchriste:
> Can you explain more about removing the geodesic last query cache? The optimizer will now repeatedly call `ComputeFaceAndWeights` for the same point, which is a relatively expensive operation. I ran `_thin_cavity_bean_` use case on master vs geodesic_exe branch and see about a 5% slowdown on geodesic_exe branch.

I compared the full tests and the `geodesic_exe` branch was actually _about 10% faster_ than `master` (neither had to download the dataset). I can't spend my time comparing for such a small gain/loss. If you want, I'll leave the first point query cache code in VtkMeshWrapper, but I'm fairly confident it's not doing anything, and also fairly confident that's exactly what `vtkCellLocator::FindClosestPoint` (the only notably function called by `VtkMeshWrapper::GetTriangleForPoint`) is doing given that reversing the order of arguments as I suggested in the first place resulted in an **over 600x speedup.**

akenmorris (changes requested):
Let's leave the last query cached value in VtkMeshWrapper for now.

PR #1497, opened 2021-09-30
[WIP] Particle check

akenmorris:
This is not meant to me merged but to demonstrate one way in which the pipeline outputs can be checked.  This adds a ParticleSystem::Compare call and shows it being used for left_atrium --tiny_test.  For this test, the results are the same across platforms.  The baseline results are stored in git.  I disagree with storing them in the portal for a variety of reasons.  For one thing, any user (not even at SCI) should be able to make modifications on their own branch that change the results and be able to update baseline values so that their own tests work.


sheryjoe:
> I disagree with storing them in the portal for a variety of reasons. For one thing, any user (not even at SCI) should be able to make modifications on their own branch that change the results and be able to update baseline values so that their own tests work. This should not require uploading data anywhere.

Agreed. I think the best option is the sci/shapeworks folder where we can control write/read permissions.

akenmorris:
> Agreed. I think the best option is the sci/shapeworks folder where we can control write/read permissions.

My suggestion is to store the baseline results in Git so that they are versioned along with the code as we do with other automated tests. Particle files are very small. If they are anywhere else then you have permission issues and someone will have to manually maintain bookkeeping to keep track of every version as the code changes.

akenmorris:
@sheryjoe , I'm having second thoughts about this approach of comparing particle positions to N-digit signifigance.  I worry it will be too arduous to re-run all use cases (even if automated) on all platforms to update baseline data.  For example, comparing the left atrium run between mac and linux, the particle positions are completely different:

```
9.89514 6.78884 20.6203
-32.4108 3.92866 6.55981
28.1908 3.09603 -23.1811
30.1621 -3.41728 -2.01533
...
```

vs

```
-13.1715 -6.35506 21.7747
-32.2528 3.81845 6.04235
6.79825 5.47302 -32.6691
24.1164 -6.95967 12.5086
...
```

But the resulting models are more or less "equivalent":

Mac:

mac

Linux:

linux

I'm wondering if we should perhaps instead use shape evaluation metrics to compare shape models. This would allow a single baseline for all platforms and also not require examining and updating baseline values for insignificant changes.  What do you think?

cchriste:
Is this perhaps due to using different (or simply not using) rng seeds? We
have added the ability to specify these in ShapeworksUtils, even on the
Python API side, for exactly this reason.

akenmorris:
Switching to a consistent RNG gives us consistent results on the same platform.  See #562.

The issue between platforms/compilers is regarding floating point math.  For example:

https://stackoverflow.com/questions/49471943/floating-point-arithmetic-varies-between-g-and-clang

https://gcc.gnu.org/wiki/FloatingPointMath

cchriste:
I am not convinced since these results are orders of magnitude more
different than floating point divergence could account for. But there are
numerous iterations and so I may be wrong. A graph of the values as the
computations evolve could help answer this question more clearly.

Please look at ShapeworksUtils since it uses Mersenne and provides a
centralized method to produce random numbers. One problem could be threaded
requests of random numbers for which we could provide thread-local
generators or another appropriate solution.

Also, if the solution for #562 is a duplicate of what we added to
ShapeworksUtils, we should probably consolidate these.

akenmorris:
Multithreaded RNG access was removed long ago.  Results are perfectly reproducible on the same platform/compiler regardless of the number of threads.

There's a lot of wiggle room in C++ floating point math.  We have hundreds of billions of floating point operations in an hour long optimization for the left atrium.  Unless the platforms/compilers are using identical instructions, FP rounding modes, etc, results are bound to diverge.

cchriste:
> Multithreaded RNG access was removed long ago. Results are perfectly reproducible on the same platform/compiler regardless of the number of threads.

👍 

> There's a lot of wiggle room in C++ floating point math. We have hundreds of billions of floating point operations in an hour long optimization for the left atrium. Unless the platforms/compilers are using identical instructions, FP rounding modes, etc, results are bound to diverge.

Can we use doubles?

akenmorris:
> Can we use doubles?

We use doubles for most things.  I meant floating point math in the sense of IEEE754 which covers all precision levels.

It's the same issue.

akenmorris:
I think it’s worth investigating how/when things diverge.

Issue #1506, opened 2021-10-09
Mesh warp crash


akenmorris:
The shape model here is terrible, but I encountered a crash in meshwarping that should be fixed:

[meshwarp_crash.zip](https://github.com/SCIInstitute/ShapeWorks/files/7315092/meshwarp_crash.zip)

PR #1508, opened 2021-10-13
Fix mesh warping crash #1506

akenmorris:
The problem was that while we check for already deleted/dead cells on the particle hit, if we are on an edge, then we need to check the neighbor cell as well before proceeding.  If it's deleted, then we need wait until the next iteration.

I've added another automated test containing the data that demonstrates the crash, confirmed that the test fails prior to the fix and succeeds after.

### DIFF
--- a/Testing/data/mesh_warp/mesh_warp3_baseline.vtk
+++ b/Testing/data/mesh_warp/mesh_warp3_baseline.vtk
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:15d54784ef6246c84584a054de7c0d11aa32c06b6832f23eee53c1dcdf5aec59
+size 323879


archanasri (approved):

Testing/MeshTests/MeshTests.cpp
```diff
@@ -757,6 +757,37 @@ TEST(MeshTests, warpTest2)
...
+  ParticleSystem particlesystem(paths);
+  Eigen::MatrixXd allPts = particlesystem.Particles();
+  Eigen::MatrixXd staticPoints = allPts.col(0);
+  Eigen::MatrixXd movingPoints = allPts.col(1);
+
+  int numParticles = staticPoints.rows() / 3;
+  staticPoints.resize(3, numParticles);
+  staticPoints.transposeInPlace();
+  movingPoints.resize(3, numParticles);
+  movingPoints.transposeInPlace();
```
cchriste, line 781: I'm unfamiliar with this ParticleSystem, so take this with a grain of NaCl, but it seems most of this resizing and transposition could be done inside the ParticleSystem. Not for this PR, but something to consider.

Testing/data/mesh_warp/mesh_warp3b.particles
```diff
@@ -0,0 +1,128 @@
...
+13.7168    0.327632    7.10369
+-2.78519    -8.99136    3.8792
+19.6565    0.369836    0.119015
+1.55236    9.84213    0.412358
+7.13004    5.06294    -7.69477
+-14.7782    -3.2087    5.69722
+2.87268    5.14811    4.19701
+-10.9736    4.16212    -7.08014
+19.578    -0.384228    0.843869
+-19.6512    -0.0573275    0.525887
```
cchriste, line 69: This particle seems a little suspicious... 🤨

akenmorris: I don't know what you mean, but these particles are garbage anyway, they just happened to produce the crash.

PR #1511, opened 2021-10-13
Don't upgrade pip to latest, it exposes us to breaking changes

akenmorris:
Stay on 20.2.3

### DIFF
--- a/install_shapeworks.bat
+++ b/install_shapeworks.bat
@@ -28,7 +28,7 @@ REM reactivate shapeworks environment
 call conda activate base
 call conda activate %CONDAENV%
 
-call pip install --upgrade pip
+call conda install --yes pip=20.2.3
 call pip install termcolor==1.1.0
 call pip install matplotlib==3.1.2
 call pip install hotelling==0.5.0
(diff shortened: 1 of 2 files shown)


akenmorris:
2021-10-14T00:21:36.0262547Z ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.


That's a nice message to see (windows)

akenmorris:
The actual error:
```

021-10-14T00:23:11.5517590Z     Uninstalling nbconvert-5.6.1:
2021-10-14T00:23:11.5647680Z       Successfully uninstalled nbconvert-5.6.1
2021-10-14T00:23:15.0597879Z ERROR: Could not install packages due to an EnvironmentError: [WinError 5] Access is denied: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\pip-uninstall-wjn89wac\\jupyter-nbconvert.exe'

```

I found this:

https://github.com/jupyter/nbconvert/issues/1476

"Looks like pip is trying to update a package by first uninstalling a package that was installed with conda. If you are in a conda environment, then use conda install nbconvert. Otherwise pip can break things which may require reinstalling conda in the worst case scenario."

PR #1565, opened 2021-11-03
Add Doxygen C++ API Reference to mkdocs

akenmorris:
This PR should add the Doxygen C++ API reference to the mkdocs documentation:

doxygen


cchriste:
How do I run the doxygen for this? Just using `mkdocs serve` it builds the docs but they just have placeholders for the APIs.

cchriste:
> There's nothing in "Getting Started with Documentation" which seems like the place it should be.

In fact, it looks like that whole section of the documentation needs to be updated since right now it still talks about having a separate branch for docs. (or maybe I'm still unclear about this)

akenmorris:
> How do I run the doxygen for this?

Doxygen is run when BUILD_DOCUMENTATION is enabled.

The conversion to mkdocs only happens in Support/deploy_docs.sh, which is only meant to be run by the GitHub Actions for deployment.  It also runs `Python/RunShapeWorksAutoDoc.py` to fill out the docs from the shapeworks executable.  If you want to run doxybook2 locally,  You can download the binary and run:

doxybook2 -i ${BUILD_DIR}/Documentation/Doxygen/xml -o docs/api -c docs/doxygen/doxybook2.config.json

This will overwrite the placeholders and you will see the doxygen with `mkdocs serve` locally.

akenmorris:
> separate branch for docs.

There used do be a protected `mkdocs` branch.  @sheryjoe, what happened to it?

cchriste:
> Doxygen is run when BUILD_DOCUMENTATION is enabled.

This correctly runs doxygen.

> The conversion to mkdocs only happens in Support/deploy_docs.sh, which is only meant to be run by the GitHub Actions for deployment. 

I think it's important to be able to run this locally.

> It also runs `Python/RunShapeWorksAutoDoc.py` to fill out the docs from the shapeworks executable. 

NOTE: this must be run from `$SRC/docs` as it has hard-coded paths to `../docs/...`

> If you want to run doxybook2 locally, You can download the binary and run:

This doesn't seem to work with a dev tree. I didn't try downloading the binary release, but I think it's important to be able to see changes without first pushing to GitHub.

> doxybook2 -i ${BUILD_DIR}/Documentation/Doxygen/xml -o docs/api -c docs/doxygen/doxybook2.config.json

This adds link

> This will overwrite the placeholders and you will see the doxygen with `mkdocs serve` locally.

`mkdocs serve` still just shows the placeholder pages.

`mkdocs build` creates a static site, but the links to the API docs are relative while everything else is the full path, and they don't work. Replacing them with the full path shows the page correctly. If this is the only way that we can produce and view docs locally, I think we should remove `mkdocs serve` from the documentation and probably add a script to perform all these steps (silently handling the case when `doxybook2` isn't available in the path). 

> I'll add this to the "Getting Started with Documentation" page.

👍

akenmorris:
>>    If you want to run doxybook2 locally, You can download the binary and run:

> This doesn't seem to work with a dev tree. I didn't try downloading the binary release, but I think it's important to be able to see changes without first pushing to GitHub.

I meant the binary of doxybook2, not ShapeWorks.

akenmorris:
I disagree with removing `mkdocs serve`.  It works really well for developing the documentation since when you save a markdown file, the page in your browser automatically updates.  It would be a pain to have to re-run `mkdocs build` to see your results.

I also recommend that most people work on the documentation without the doxygen files in place (e.g. use the placeholders).  Otherwise, `mkdocs serve` takes an extra 60-90 seconds and is really slow in general.

cchriste:
> I disagree with removing `mkdocs serve`. It works really well for developing the documentation since when you save a markdown file, the page in your browser automatically updates. It would be a pain to have to re-run `mkdocs build` and manually reload the page to see your results.
> 
> I also recommend that most people work on the documentation without the doxygen files in place (e.g. use the placeholders). Otherwise, `mkdocs serve` takes an extra 60-90 seconds and is really slow in general.

I installed doxybook2 and it runs just fine. The problem is the mkdocs.

I also completely agree that `mkdocs serve` is far easier than manually rebuilding the docs every time. It's just important to be able to see the APIs during development and it's not yet working. Suggesting we get rid of mkdocs serve was only as a last resort if the other can't be fixed.

akenmorris:
Ok, I’m not sure what the problem is. Are you getting an error somewhere?

cchriste:
> Ok, I’m not sure what the problem is. Are you getting an error somewhere?

Seeing the APIs during development and is not yet working.

akenmorris:
I've added a new script `Support/build_docs.sh` that should hopefully do these steps for you.

After build:

```
cd <shapeworks code root>
./Support/build_docs.sh
mkdocs serve
```

cchriste:
The script is handy, but seeing the APIs during development is still not working. The file exists, but the url doesn't work.
```
(sw62) cam@ananda:~/code/ShapeWorks/ShapeWorks$ find . -name "classColorSchemes.html"
./build_debug/Documentation/Doxygen/html/classColorSchemes.html
```
Screen Shot 2021-11-08 at 4 30 43 PM
.

akenmorris:
I've tweaked the relative path settings in Doxybook2 so that it should hopefully handle all cases now.

cchriste:
Yay! This is fantastic. Side question: how are groups designated? For example, there are multiple classes in the Mesh group (e.g., Mesh, MeshUtils, etc), but only Image in the Image group.

Next issues:

1. For some reason it's showing the entire shapeworks tree (maybe even all the classes and namespaces) twice in the Classes section.

2. Some files are changed by running the `build_docs.sh` script.
```
(sw62) cam@ananda:~/code/ShapeWorks/ShapeWorks$ git status
On branch mkdoxygen
Your branch is up to date with 'origin/mkdoxygen'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   docs/api/index_classes.md
	modified:   docs/api/index_files.md
	modified:   docs/api/index_groups.md
	modified:   docs/api/index_namespaces.md
```

sheryjoe:
@akenmorris is this part of 6.2?

akenmorris:
> how are groups designated? 

https://www.doxygen.nl/manual/grouping.html

https://github.com/SCIInstitute/ShapeWorks/blob/943c02894aabf44ce189b154771fac02c8082b62/Libs/Image/Image.h#L20

I just threw a couple in as an example.  Most classes are missing Doxygen tags altogether.

akenmorris:
> For some reason it's showing the entire shapeworks tree (maybe even all the classes and namespaces) twice in the Classes section.

I'm not seeing this on my end.  Do both links go to the same target?

akenmorris:
> Some files are changed by running the build_docs.sh script.

This should be fixed now, we don't really need the placeholders.

cchriste:
1. This warning is printed, with different source files _including_ api/Namespaces/namespaceshapeworks.md, a lot:
```
WARNING  -  Documentation file 'api/Files/SplashScreen_8h.md' contains a link to
            'api/Namespaces/namespaceshapeworks.md' which is not found in the documentation files.
WARNING  -  Documentation file 'api/Namespaces/namespaceShapeWorks.md' contains a link to
            'api/Namespaces/namespaceshapeworks.md' which is not found in the documentation files.
```

In particular, this is strange since there _isn't_ an all-lowercase version of this file, even after building it all and while `mkdocs serve` is running!
```
(sw62) cam@ananda:~/code/ShapeWorks/ShapeWorks/build_debug$ find .. -iname "namespaceshapeworks*"
../build_debug/Documentation/Doxygen/xml/namespaceShapeWorks.xml
../docs/api/Namespaces/namespaceShapeWorks.md
```

2. There is still one file that gets changed after building docs:
(btw, it's awesome that the docs get updated automatically even while being served, which restoring this file revealed.)
```
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   docs/tools/ShapeWorksCommands.md
```

3. Lots of things such as the shapeworks namespace still show up twice in the Classes page, along with things in one's local tree, as this screenshot illustrates:
Screen Shot 2021-11-10 at 8 20 09 AM

cchriste:
> > For some reason it's showing the entire shapeworks tree (maybe even all the classes and namespaces) twice in the Classes section.
> 
> I'm not seeing this on my end. Do both links go to the same target?

Yes they do.

akenmorris:
> modified:   docs/tools/ShapeWorksCommands.md

Yeah, this will happen if the shapeworks command has changed.

I guess when it gets updated like this, you should commit it.

akenmorris:
So the overview dox file was telling Doxygen that there was a "ShapeWorks" namespaces and the two obviously override each other on a case insensitive filesystem.  This should be fixed now and I've also added support for build directories that are inside the source code tree.

cchriste:
Fantastic! Everything is clean and working now. 
I thought I was being overly picky, but it's actually the succinct view of only what is there is _much_ more useful.

@sheryjoe If you agree, we can go ahead and merge this.

cchriste (approved):

cchriste:
I have a question: why do some pages include their relevant files (e.g., the page for Image class), while other do not (e.g., the page for GroomTool)?

akenmorris:
> why do some pages include links to their relevant files (e.g., the page for Image class), while other do not (e.g., the page for GroomTool)?

No idea.

PR #1578, opened 2021-11-10
Removing epsEqualN and equalNSigDigits

### LINKED ISSUE #1544: remove equalNSigDigits, replace with epsEqual
After some additional testing with @akenmorris, this function has been shown to be invalid for some tests and should be removed. While epsEqual requires more specific knowledge about the appropriate range of results to be compared, it is able to be used predictably.

To determine if this should be replaced sooner than later, I suggest simply calling epsEqual from this function to see which tests fail as its use may be masking other problems.

archanasri:
This PR resolves #1544


cchriste (approved):

akenmorris (approved):

akenmorris:
I think those values are close enough.  I would just adjust the threshold for now.  We can add an issue to investigate later, though it's probably just fine as is.

Issue #1490, opened 2021-09-28
Add FFC functionality for mesh domains


HeavenlyBerserker:
Should be simple using the same strategy of clipping the mesh that cutting planes use for mesh domains.

akenmorris:
Merged in #1614

PR #1614, opened 2022-01-04
Mesh FFC (by clipping)

HeavenlyBerserker:
Uses mesh clipping to add free form constraint support for meshes.

Tackles issue #1490.


Libs/Optimize/OptimizeParameterFile.cpp
```diff
@@ -521,6 +522,15 @@ bool OptimizeParameterFile::read_mesh_inputs(TiXmlHandle* docHandle, Optimize* o
           mesh.clip(plane);
         }
       }
+
+      std::cout << "ffcssize " << ffcs.size() << std::endl;
```
akenmorris, line 526: Let's remove the output unless verbose.

Testing/OptimizeTests/OptimizeTests.cpp
```diff
@@ -798,3 +798,79 @@ TEST(OptimizeTests, contour_domain_test)
...
+  for (size_t domain = 0; domain < num_doms; domain++) {
+    std::vector<itk::FixedArray<double, 3> > list;
+    for (auto k = 0;
+         k < app.GetSampler()->GetParticleSystem()->GetPositions(domain)->GetSize(); k++) {
+      list.push_back(app.GetSampler()->GetParticleSystem()->GetPositions(domain)->Get(k));
+    }
+    lists.push_back(list);
+  }
+
+  bool good = true;
```
akenmorris, line 846: This whole section seems to be copied and pasted from ffc_test.  It would be much better to write a utility function so that they could shared the same code.

akenmorris:
@HeavenlyBerserker , how many times have you copied and pasted the 50 lines of constraint checking code in `OptimizeTests.cpp` ?  7-8 times?

akenmorris (approved):

PR #1618, opened 2022-01-06
Mesh grooming in use cases

jadie1:
This PR adds mesh grooming to the ellipsoid_mesh use case. It also combines the femur, femur_mesh, and femur_cut use cases into one single use case (femur_cut) which does mesh grooming, grooms the corresponding images, and optimizes on meshes with a single cutting plane as a constraint.

The portal, verification data, and documentation have been updated accordingly.


archanasri:
transform.py needs to be improved. It does not test anything right now. Just runs createTransfrom and applyTransform functions.

PR #1619, opened 2022-01-07
ReconstructSurface API

archanasri:
This PR adds the ReconstructSurface API. It consolidates ReconstructSurface.cxx and ReconstructSamplesAlongPCAModes.cxx. It consolidates parts of Reconstruct.h/.cpp as well.

Tests have been added.


Libs/Common/Shapeworks.h
```diff
@@ -1,5 +1,8 @@
 #pragma once
 
+#include "Libs/Alignment/Transforms/itkThinPlateSplineKernelTransform2.h"
```
cchriste, line 3: Not sure if it's necessary, but I really like the fully paths being used for these includes since they are quite confusingly named and likely to be mistaken for actual itk includes. 👍

Libs/Common/Shapeworks.h
```diff
@@ -42,6 +46,10 @@ using IdentityTransform  = itk::IdentityTransform<double, 3>;
 using TransformPtr       = GenericTransform::Pointer;
 TransformPtr createTransform(const Matrix33 &mat, const Vector3 &translate = makeVector({0,0,0}));
 
+/// Transforms that can be used for ReconstructSurface
```
cchriste, line 49: This is a nice way to guide the use of a templated class while also providing a clear path to extending it.

Libs/Mesh/CMakeLists.txt
```diff
@@ -70,17 +70,17 @@ install(TARGETS Mesh
 #TARGET_LINK_LIBRARIES(GenerateFidsFiles ${ITK_LIBRARIES} Mesh tinyxml)
 #INSTALL(TARGETS GenerateFidsFiles RUNTIME DESTINATION bin)
 
-ADD_EXECUTABLE(ReconstructMeanSurface ReconstructMeanSurface.cxx)
-TARGET_LINK_LIBRARIES(ReconstructMeanSurface ${ITK_LIBRARIES} tinyxml Utils Analyze)
-INSTALL(TARGETS ReconstructMeanSurface RUNTIME DESTINATION bin)
+# ADD_EXECUTABLE(ReconstructMeanSurface ReconstructMeanSurface.cxx)
```
cchriste, line 73: If they're gone let's get rid of them in the cmake lists as well.

Libs/Python/ShapeworksPython.cpp
```diff
@@ -1328,6 +1329,100 @@ PYBIND11_MODULE(shapeworks_py, m)
        "return the variance accounted for by the principal components")
   ;
 
+  py::class_<ReconstructSurface<ThinPlateSplineTransform>>(m, "ReconstructSurface_ThinPlateSplineTransform")
```
cchriste, line 1332: These class names are really not long enough. How thin are the plates of the spline? What color is the surface? Are they using tractors or shovels?! </sarcasm>

cchriste (approved):

Libs/Common/Shapeworks.h
akenmorris: Is there any way to do this without adding these includes to Shapeworks.h?  These will now get included almost everywhere and impact compilation speed of each file.

Libs/Mesh/CMakeLists.txt
archanasri: Removed them

Libs/Common/Shapeworks.h
archanasri: I moved it to ReconstructSurface.h

akenmorris (approved):

PR #1629, opened 2022-01-12
Fix for #1483

### LINKED ISSUE #1483: Crop followed by resample results in an empty image, but crop->save->load->resample works fine
I think Image::crop should probably use itk::RegionOfInterestImageFilter instead of itk::ExtractImageFilter.

From https://itk.org/Doxygen/html/classitk_1_1ExtractImageFilter.html

"The output image produced by this filter will have the same origin as the input image, while the ImageRegion of the output image will start at the starting index value provided in the ExtractRegion parameter. If you are looking for a filter that will re-compute the origin of the output image, and provide an output image region whose index is set to zeros, then you may want to use the RegionOfInterestImageFilter. The output spacing is is simply the collapsed version of the input spacing."

It appears Image::resample makes some assumptions that are incorrect for the output of itk::ExtractImageFilter

```
Dims dims({ static_cast<unsigned>(std::floor(inputDims[0] * inputSpacing[0] / spacing[0])),
              static_cast<unsigned>(std::floor(inputDims[1] * inputSpacing[1] / spacing[1])),
              static_cast<unsigned>(std::floor(inputDims[2] * inputSpacing[2] / spacing[2])) });
  
  Point3 new_origin = origin() + toPoint(0.5 * (spacing - inputSpacing));  // O' += 0.5 * (p' - p)
```

To demonstrate, try the following python code:

```
from shapeworks import *

img = Image("seg.nrrd")
img.crop(img.physicalBoundingBox(0.5))
img.write("cropped.nrrd")

img.resample(1.0)
img.write("resampled.nrrd")

img = Image("cropped.nrrd")
img.resample(1.0)
img.write("resampled2.nrrd")
```

The resulting **resample.nrrd** is empty, but **resampled2.nrrd** is what the user expects.

Switching to itk::RegionOfInterestImageFilter gives both files being identical and what is expected.

[example.zip](https://github.com/SCIInstitute/ShapeWorks/files/7212190/example.zip)

I'm not sure the other implications of switching filters.  The region of interest filter doesn't have `SetDirectionCollapseToIdentity` that we are using.

archanasri:
This PR resolves #1483.

Image.crop has been updated to use RegionOfInterestImageFilter instead of ExtractImageFilter.
Added tests.

### DIFF
--- a/Libs/Image/Image.cpp
+++ b/Libs/Image/Image.cpp
@@ -742,14 +742,13 @@ Image& Image::crop(PhysicalRegion region, const int padding)
     throw std::invalid_argument("Invalid region specified (it may lie outside physical bounds of image).");
   }
 
-  using FilterType = itk::ExtractImageFilter<ImageType, ImageType>;
+  using FilterType = itk::RegionOfInterestImageFilter<ImageType, ImageType>;
   FilterType::Pointer filter = FilterType::New();
-  
+
   IndexRegion indexRegion(physicalToLogical(region));
   indexRegion.pad(padding);
-  filter->SetExtractionRegion(ImageType::RegionType(indexRegion.min, indexRegion.size()));
+  filter->SetRegionOfInterest(ImageType::RegionType(indexRegion.min, indexRegion.size()));
   filter->SetInput(this->image);
-  filter->SetDirectionCollapseToIdentity();
   filter->Update();
   this->image = filter->GetOutput();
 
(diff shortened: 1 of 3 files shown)


cchriste (approved):
Awesome. Thanks for the suggestions, Alan. And a great use of Jupyter notebooks to explore the solution, Archana!

akenmorris (approved):

Issue #1630, opened 2022-01-12
Add check and warnings for shape evaluation commands


iyerkrithika21:
- Shape evaluation commands querying metrics for the specified mode of variation should check if the entered mode is valid or not. Add appropriate warning for these purposes. 
- The evaluation use case should not proceed if no particle files are found.

iyerkrithika21:
Fixed with #1631

PR #1631, opened 2022-01-12
Mode of variation check for shape evaluation

iyerkrithika21:
Address #1630 

After the warnings this is the output if the 3rd mode is queried for tiny test with 3 samples: 
```
(shapeworks_new) iyerkrithika@krithika:~/ShapeWorks_new/ShapeWorks/Examples/Python> python RunUseCase.py ellipsoid_evaluate --tiny_test
Using shapeworks module from /home/sci/iyerkrithika/ShapeWorks_new/ShapeWorks/Python/shapeworks/shapeworks/__init__.py
Using shapeworks from /home/sci/iyerkrithika/ShapeWorks_new/ShapeWorks/build/bin/shapeworks

Compactness
-----------
Invalid mode of variation specified
Compactness value of the 3rd mode -  0.0

Generalization
--------------
Invalid mode of variation specified
Generalization value of the 3rd mode -  0.0

Specificity
--------------
Invalid mode of variation specified
Specificity value of the 3rd mode -  0.0

````

### DIFF
--- a/Examples/Python/ellipsoid_evaluate.py
+++ b/Examples/Python/ellipsoid_evaluate.py
@@ -27,6 +27,10 @@ def Run_Pipeline(args):
 
     # get the list of all the world particles
     particleFilesList = glob.glob(shape_models_dir+"/*world.particles")
+    if(len(particleFilesList)==0):
+        print(f'Shape model file at {shape_models_dir} empty. No particle files found',file=sys.stderr)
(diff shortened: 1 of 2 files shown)


akenmorris (changes requested):
I was thinking it should probably throw an exception like std::invalid_argument rather than just print a message and returning 0.

akenmorris (approved):

Issue #182, opened 2019-09-30
Add `shapeworks::Exception` for better error handling


akenmorris:
Most of our tools could use some better error handling. For example, right now if distance transforms don’t have matching origins, View2 and Studio will just dump out with “std::exception” with no indication what the problem is.

akenmorris:
I think this is a fairly useless issue to have on GitHib as it’s not specific enough and has no real way of being completed. (I’m the one that added it) better is too subjective.

cchriste:
This class should include `operator<<` that describes the reason for it being thrown.
See example usage in the comment for this commit: 
https://github.com/SCIInstitute/ShapeWorks/commit/1a56525e8e31fb1a9c2b297c5ad67f8973a0c3e7

Creating shapeworks-specific exceptions will help us handle errors better at multiple levels. This issue merely creates the class (and an associated unit test, preferably like the example in the comment).

PR #1632, opened 2022-01-15
Exception class

archanasri:


cchriste (approved):

cchriste:
Deriving from `std::runtime_error` seems appropriate. 

Since where to use this is somewhat ambiguous, I think this is sufficient to resolve #182. New issues can be created for where/how it should be used and if any other project-specific exceptions should be created.

akenmorris:
Looks fine to me. Should the class name be ShapeWorksException?  I think among all of our various styles, classes are at least all PascalCase, right?

cchriste:
> Looks fine to me. Should the class name be ShapeWorksException or just Exception (in shapeworks namespace?)? I think among all of our various styles, classes are at least all PascalCase, right?

I think we were trying to follow the style of C++ exceptions.

akenmorris (approved):

PR #1636, opened 2022-01-18
Fix #1635 - save procrustes transforms on `shapeworks optimize` command

### LINKED ISSUE #1635: shapeworks optimize command needs to save procrustes transforms upon completion
Right now the procrustes transforms are only filled in when run by Studio.

akenmorris:


jadie1 (approved):
Adds a Procrustes column to the spreadsheet as expected

PR #1661, opened 2022-02-12
Warp meshes with landmarks

nawazishkhan1-nk:
Added a new command `warp-mesh-with-landmarks` for transforming landmark particles along with mesh warp.


nawazishkhan1-nk:
Reference Mesh with Landmarks :
Screen Shot 2022-02-11 at 5 58 36 PM


Warped Mesh With  Landmarks:
Screen Shot 2022-02-11 at 5 57 57 PM

akenmorris (changes requested):
The changes in the PR need to be refactored quite a bit.  I see many functions that have been copied and pasted and then one version is slightly tweaked.  This leads to significant future development and maintenance costs as future changes must be correctly applied to both versions of these functions, or worse may mistakenly only be applied to one and not the other, leading to very confusing bugs.

The other big issue I see is that these changes appear to break the multi-threaded usage by studio of this class.  The `build_mesh` function with landmarks is calling a function that modifies the private member `landmarks_map_` with no mutex.  If multiple threads call the first `build_mesh` at the same time, this may crash.

ben2k:
Hi Nawazish,

It seems that you find the closest vertex from for each landmark to warp. It can create inaccuracies depending of the mesh density no ?  

 Why no apply the following algorithm to warp landmarks:

- Compute a ThinPlateSpine function between the source and target particles
- Evaluate source landmarks to obtain (target) warped landmarks
- Project landmarks on the target surface to obtain final locations

ben2k:
Hi Nawazish,

It seems that you find the closest vertex from for each landmark to warp. It can create inaccuracies depending of the mesh density no ?  

 Why no apply the following algorithm to warp landmarks:

- Compute a ThinPlateSpine function between the source and target particles
- Evaluate source landmarks to obtain (target) warped landmarks
- Project landmarks on the target surface to obtain final locations

nawazishkhan1-nk:
> Hi Nawazish,
> 
> It seems that you find the closest vertex from for each landmark to warp. It can create inaccuracies depending of the mesh density no ?
> 
> Why no apply the following algorithm to warp landmarks:
> 
>     * Compute a ThinPlateSpine function between the source and target particles
> 
>     * Evaluate source landmarks to obtain (target) warped landmarks
> 
>     * Project landmarks on the target surface to obtain final locations

Hi Benjamin,

I was also a bit concerned about the inaccuracy which might come from the mesh density after landmarks propagation. I discussed this with @sheryjoe  earlier this week in our meet and we decided to have a two-step process to warp the landmarks. 
1. In the first step, using the warp-mesh-with-landmarks with EOS-mesh as the reference mesh, EOS-particles (coming after fixed domain step) as reference points, landmark defined on EOS-mesh and target points will be the local points of all CT samples coming from the shape model. This will bring the EOS landmarks to all the CT samples' space using the biharmonic warp matrix. Coming back to your comment about closest vertex - Since we have the reference mesh as EOS-mesh, the closest point between the landmarks and the reference mesh will be on the mesh itself as the landmarks are defined on it. This is computed just to get the vertex ID corresponding to all landmarks which is invariant during the warping process. Using these mesh vertices IDs we can find the corresponding location of all the landmarks in the CT smaple space after applying the warp matrix.

2. The next step would be to project the warped landmarks in CT sample space back to the (actual) reconstructed meshes of the shape model using projectPointsOntoMesh.

I think this is much similar to what you are suggesting?

ben2k:
> > Hi Nawazish,
> > 
> > It seems that you find the closest vertex from for each landmark to warp. It can create inaccuracies depending of the mesh density no ?
> > 
> > Why no apply the following algorithm to warp landmarks:
> > 
> >     * Compute a ThinPlateSpine function between the source and target particles
> > 
> >     * Evaluate source landmarks to obtain (target) warped landmarks
> > 
> >     * Project landmarks on the target surface to obtain final locations
> 
> Hi Benjamin,
> 
> I was also a bit concerned about the inaccuracy which might come from the mesh density after landmarks propagation. I discussed this with @sheryjoe  earlier this week in our meet and we decided to have a two-step process to warp the landmarks. 
> 1. In the first step, using the warp-mesh-with-landmarks with EOS-mesh as the reference mesh, EOS-particles (coming after fixed domain step) as reference points, landmark defined on EOS-mesh and target points will be the local points of all CT samples coming from the shape model. This will bring the EOS landmarks to all the CT samples' space using the biharmonic warp matrix. Coming back to your comment about closest vertex - Since we have the reference mesh as EOS-mesh, the closest point between the landmarks and the reference mesh will be on the mesh itself as the landmarks are defined on it. This is computed just to get the vertex ID corresponding to all landmarks which is invariant during the warping process. Using these mesh vertices IDs we can find the corresponding location of all the landmarks in the CT smaple space after applying the warp matrix.
> 
> 2. The next step would be to project the warped landmarks in CT sample space back to the (actual) reconstructed meshes of the shape model using projectPointsOntoMesh.
> 
> I think this is much similar to what you are suggesting?

The landmarks defined on EOS mesh can be located everywhere on the surface (not necessary belonging to a vertex). May be we have to find the closest point on the closest triangle (point to cell) and store the point location in triangle coordinates (using two coefficients from which we can recover the point location on the surface knowing the three vertices of the face).  The mapping should be in this case the FaceID and two linear coefficients, not only the closest VerticeID to have an accurate warping (for landmark still far away to his closest vertex). What do you think ?

In mesh warping, vertices (and triangles) are added in order that each particle strictly belong to one vertex of the reference surface (because biharmonic warp matrix requires that control points belong to the topology I believe). I don't know if this the case with the warp-mesh-with-landmarks.

akenmorris:
I would recommend changing this implementation to use a similar approach to the particles.  For the particles, we insert a new vertex at the location of each particle and split the triangle that it lands on.  This can be done for landmarks as well so that they will always have an exact vertex associated with them.

ben2k:
> The changes in the PR need to be refactored quite a bit. I see many functions that have been copied and pasted and then one version is slightly tweaked. This leads to significant future development and maintenance costs as future changes must be correctly applied to both versions of these functions, or worse may mistakenly only be applied to one and not the other, leading to very confusing bugs.
> 
> The other big issue I see is that these changes appear to break the multi-threaded usage by studio of this class. The `build_mesh` function with landmarks is calling a function that modifies the private member `landmarks_map_` with no mutex. If multiple threads call the first `build_mesh` at the same time, this may crash.

@nawazishkhan1-nk perhaps you should add this feature in a Python binding to avoid copy/past of commands in Shapeworks. I did the work to add a binding for MeshWarper ([https://github.com/SCIInstitute/ShapeWorks/pull/1705])  allowing to write:

```
particle_sys = sw.ParticleSystem(local_point_files)
median_shape_idx = find_median_shape()

warping = sw.MeshWarper()
warping.computeWarping(meshes[median_shape_idx], particle_sys.ShapeAsPointSet(median_shape_idx))
print('Warpping has bad particle: {}'.format(warping.hasBadParticles()))

mesh_warped_0 = warping.buildMesh(particle_sys.ShapeAsPointSet(0))

warping.getReferenceModel().write(f"{self.eval_out_dir}/mean_mesh.vtk")
```

In your case just add other constructor with additional landmarks to wrap:

```
warping = sw.MeshWarper()
warping.computeWarping(median_meshes, median_particles, median_landmarks)
[warped_mesh, warped_landmarks] = warping.buildMesh(control_particles)
```

The algo used for landmarks warping is still to define, I'm talking about doing directly in Python (without calling the external exe tool shapeworks)

akenmorris:
> perhaps you should add this feature in a Python binding to avoid copy/past of commands in Shapeworks.

But we will eventually use this functionality in Studio to propagate landmarks for analysis.  It won't make sense to call out to Python and back to do this.

ben2k:
> > perhaps you should add this feature in a Python binding to avoid copy/past of commands in Shapeworks.
> 
> But we will eventually use this functionality in Studio to propagate landmarks for analysis. It won't make sense to call out to Python and back to do this.

Sorry: we add the feature in the Core C++ MeshWarper class, then we can use it both in Studio and in Python (via binding)

akenmorris:
I've fixed the conflicts and cleaned up the interface a bit.  The landmarks are now given with the mesh, but are optional rather than having a second call to set them.  And the size of the landmarks can be used anywhere to detect if they were set or not.

The only thing left to do as far as I can see is to add a unit test for this functionality.  Have a look at `MeshTests.cpp` `warpTest1`, `warpTest2`, `warpTest3` for examples.

akenmorris (changes requested):
Just needs unit tests for the landmark functionality

ben2k:
I will add the corresponding Python binding and some other tools needed for the project  on the top of this commit

akenmorris (approved):

PR #1662, opened 2022-02-13
Small fix for cohort generator and added documentation

iyerkrithika21:
1. Added clarification based on a question asked on discourse: https://shapeworks.discourse.group/t/p-value-feature-map/70
2. ShapeCohortGenPackage was failing because of the way shapeworks was imported:
`from shapeworks import *`
This was causing the `range` function to misbehave and hence this error: 
image

Changed the import statement to: `import shapeworks as sw`, and now the package works without any error.


jadie1 (approved):
Looks good, in general we should avoid `from - import *`
Thanks for fixing this @iyerkrithika21

PR #1665, opened 2022-02-15
Studio Image Support

akenmorris:
Provides basic image (3d volume) support in Studio.

2D/3D Mode
Shared Window and Level (Brightness and Contrast) or separate
X/Y/Z planes
Shape/plane intersection
Only show particles for current slice

More details to come...

image

Sample data here:

http://www.sci.utah.edu/~amorris/sample_image_volume2.zip


jadie1:
Hi Alan, I was testing this with your sample data. The data tab looks and works as expected for me. When I move to the Analyze tab this is what I am seeing:
image
I am guessing these are the particles for the current slice, but am I supposed to also be able to see the image slice as well?

akenmorris:
Sorry, that turned out to be an exceptionally bad sample project.  It contained an optimization with only two particles, so your mesh warp basically failed and all you saw was the two particles.  Try this:

http://www.sci.utah.edu/~amorris/sample_image_volume2.zip

The mean shape and any other derived shapes (PCA modes) won't have any image data, so don't expect to see a slice.

jadie1 (approved):
The new example works well! I messed around with it and didn't run into any issues.

PR #1690, opened 2022-03-18
Fix #1689 - Particle splitting needs to use domain transforms

### LINKED ISSUE #1689: Particle splitting needs to use domain transforms
There is a general problem in ShapeWorks that the particle splitting direction is not taking into account per domain transforms.  The problem is most clearly demonstrated when reflection input prefix transforms are supplied.  Watching closely in Studio, you can see that particles split in nearly opposite directions on domains that are reflected vs those that are not.

The problem also affects procrustes alignment as well.  A singular split direction is determined, but it is applied to local positions without respect for alignment.

Here is an example.  Watch the red and green particles on the top go the opposite direction:

https://user-images.githubusercontent.com/1693349/159063136-583d7efb-f3b8-4fb1-9339-966036e59702.mp4

akenmorris:
This PR fixes:

* #1689 

We now apply the per domain local inverse transforms to the shared split direction.  This fixes the reflection problem and should also improve splitting consistency any time that procrustes or input prefix transforms are used.

Screen Shot 2022-03-18 at 12 44 05 PM

### DIFF
--- a/Libs/Optimize/ParticleSystem/ContourDomain.cpp
+++ b/Libs/Optimize/ParticleSystem/ContourDomain.cpp
@@ -1,11 +1,12 @@
 #include "ContourDomain.h"
-#include <iostream>
-#include <fstream>
+
 #include <vtkDijkstraGraphGeodesicPath.h>
 #include <vtkDoubleArray.h>
 
+#include <fstream>
+#include <iostream>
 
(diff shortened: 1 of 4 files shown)


akenmorris:
Looks like something is up with the contour domains.

Issue #1575, opened 2021-11-09
Support versioned documentation pages


akenmorris:
It would be great if we could have a snapshot of the documentation for each release posted on the website somewhere.

sheryjoe:
This supports versioning on gh-pages
https://github.com/jimporter/mike

akenmorris:
Good find.

cchriste:
Selecting some pages from documentation hides the version selection dropdown.

akenmorris:
Done

PR #1704, opened 2022-03-23
Versioned docs

cchriste:
Part of the PR is to remove gh-pages so it can be constructed using the new versioning system. A copy of that branch called gh-pages_bak has been pushed as a backup.

#1575

### DIFF
--- a/Support/build_docs.sh
+++ b/Support/build_docs.sh
@@ -26,7 +26,8 @@ else
     exit 1
 fi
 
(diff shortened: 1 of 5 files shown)


akenmorris:
Cool

cchriste:
Not sure why OSX failed. It is coincidentally the build that regenerates documentation, but the error was during building of dependencies, something about a missing git repo for Eigen. I'm going to re-run it.

akenmorris:
Seems something is wrong with the dep cache file.  I've removed it and restarted.

Issue #1702, opened 2022-03-23
Mesh needs to invalidate locators and any other data structures when copying from another mesh


akenmorris:
It's possible to end up with a Mesh class that has a vtkPolyData that doesn't match the locators.  The `operator=` call performs a deep copy of the vtkPolyData but leaves the other data structures (e.g. point locators) as is.

```
  Mesh foo("mesh1.vtk");
  Mesh bar("mesh2.vtk");
  Mesh baz = foo;
  pt = baz.closestPointId(center);  // initializes point locator
  baz = bar;  // baz now has polydata from mesh2.vtk, but locator for mesh1.vtk
```

I think `operator=` should probably call `invalidateLocators`

PR #1707, opened 2022-03-25
ensure Mesh invalidates locators when appropriate

cchriste:
for copy ctors and assignment
also moves Mesh::read into independent class, protecting its undesired use by other Mesh functions
#1702


akenmorris (approved):
Sweet.

PR #1721, opened 2022-03-30
Versioned docs attempt two

cchriste:
Trying to push to origin/gh-pages since simply pushing to gh-pages didn't work.
Also switched themes to mkdocs material.

Here are some pictures, and you can also checkout `gh-pages-local` if you want to try it out yourself (before this finishes with shining stars).
swdocs-6 3-specific-page
swdocs-6 3-class
swdocs-6 3-toc-narrow-window
swdocs-6 2


akenmorris:
I don't see any change on the docs website.

Also, someone pointed out that the doxygen stuff is all gone after the recent mkdocs changes:

http://sciinstitute.github.io/ShapeWorks/6.2/api/Modules/index_groups.md

akenmorris:
Wow, it looks great!

cchriste:
> Wow, it looks great!

And the push from the GitHub action worked! 🎉 

I'm tidying up a few things, adding logos, options to toggle themes, etc. 
But for my next push I'm going to once again disable deployment from non-master branches.
So... trust me? 😱
The local testing server seems to show things just like the github.io, so it should be okay.

There is one thing bothering me. When switching to the old version (technically "latest"), the version selector showing up is hit and miss. One way is that navigating to `<docs_url>/6.2` shows the dropdown, but when going all the way to `6.2/index.html` it doesn't show up. It's not a deal breaker, just annoying.

akenmorris:
So this should already work as is when it's run on master, right?

As I mentioned before, that's my preferred operation rather than having every PR fight to overwrite each other.

akenmorris (approved):

cchriste:
Thank you, Alan!

akenmorris:
Looks great, Cameron, nice job!

Issue #539, opened 2020-06-04
Linear discriminant analysis for groups in Studio and SW commands


sheryjoe:
- Shireen will provide a Matlab implementation that uses an optimized shape model and provide a shape score that is computed by projecting samples on the vector connecting shape means.
- This could be implemented as an API in the Analyze library and exposed in Studio (and add SW command if needed).
- Put together a use case that show case this. Scapula and femurs data have groups info.

sheryjoe:
@akenmorris here is a python implementation and example data for this.
[lda.zip](https://github.com/SCIInstitute/ShapeWorks/files/5469950/lda.zip)

sheryjoe:
@akenmorris 

can the use change the place and orientation of the legend? the legend can hide the plot. can we at least add it outside the plot?

<img width="157" alt="image" src="https://user-images.githubusercontent.com/3253036/170120763-ed4444b2-7e7a-40a0-8b2f-0c16ee7b20e9.png">

akenmorris:
Have you tried resizing the panel?  It may be kind of a losing battle to try to improve the tiny sized version of it. 

I had tried to do the multi line title with jkqtplotter without success, but I can try again. May have to modify the library.

akenmorris:
If you resize the panel:

<img width="550" alt="image" src="https://user-images.githubusercontent.com/1693349/170153595-7dd3e549-c635-4c42-9587-2ea34bb3ea3a.png">

sheryjoe:
Nice. Can we add some help text for the user to resize the panel? not sure where this should fit.

sheryjoe:
this works.

akenmorris:
Complete in #1777

PR #1728, opened 2022-04-04
Studio LDA chart

akenmorris:
See #539

Screen Shot 2022-04-04 at 1 56 53 PM

This LDA chart runs automatically when groups are chosen.

image


iyerkrithika21 (approved):
Tried to load a couple of projects to visualize the LDA plots, and looks good.
image

Issue #1430, opened 2021-08-24
Generalize fixed domain use case


jadie1:
Fixed domain use case should be adapted (and renamed) to be a more general incremental use case with two modes:

- Fixed domain mode (current use case)
- Incremental mode where we are adding new shapes but the old shapes are not fixed (new shapes affect statistics)

jadie1:
This use case will use groomed (aligned) supershape meshes as data. 
The order of optimization will be selected using mesh surface-to-surface distance (most similar first) and optimization will be run 
 incrementally (adding a fixed given amount at a time).

PR #1736, opened 2022-04-13
Incremental use case

jadie1:
This PR adds a use case that demonstrates how to do incremental optimization on supershape meshes. It will close issue #1430.

There are two cases where incremental optimization is beneficial: 
1. When the cohort of shapes is very large, where optimizing particle position across all shapes at once would take a very long time. 
2. When the dataset is small but contains a large amount of shape variation. 

In the documentation, I added a demonstration of the benefit of the incremental approach for both cases. For the second one instead of adding a second smaller dataset, I showed results using the `--use_subsample` tag. I made a new use case doc folder for multistep optimization use cases and moved the fixed domains use case doc into it as well. 

To test this PR run the `incremental_supershapes` use case (full or tiny test) and check the documentation (docs/use-cases/multistep/incremental_supershapes.md).


akenmorris (approved):

jojxl0l (commented):
- [@@]()

PR #1738, opened 2022-04-14
FFC Mesh Augmented Lagrangian

HeavenlyBerserker:
Added functionality for free form constraints for the mesh domain using augmented lagrangian in the optimization. 

- Added flag to toggle between clipping/aug-lag modes
- Added a unit test for it
- Changed some parameters for the mesh clipping unit test as well

These were added for testing but later removed because they were temporary and will be addressed again later
- Added particle splitting debugging infrastructure, which will be used for correspondence splitting next. (Removed, can reference in the future)
- Readded tangent projection during particle splitting, but better than before (Removed, can reference in the future)

Next PR:
- Add studio use case for femora and documentation


macskle1 (approved):

Libs/Optimize/OptimizeParameters.cpp
```diff
@@ -215,6 +215,12 @@ int OptimizeParameters::get_verbosity() { return this->params_.get("verbosity",
 //---------------------------------------------------------------------------
 void OptimizeParameters::set_verbosity(int value) { this->params_.set("verbosity", value); }
 
+//---------------------------------------------------------------------------
+int OptimizeParameters::get_mesh_ffc_mode() { return this->params_.get("mesh_ffc_mode", 0); }
```
akenmorris, line 219: I believe this should be a **bool**, not an **int**.  I assume it's either on or off.

HeavenlyBerserker: Yes, how does bool work here? Does it read 0/1 or does it read true/false, or is it True/False like in python? I opted for int cause at least I know 0/1 works, but I agree that bool is better.

Libs/Optimize/ParticleSystem/VtkMeshWrapper.h
```diff
@@ -102,6 +106,7 @@ class VtkMeshWrapper : public MeshWrapper {
                      const Eigen::Vector3d &vector) const;
 
   vtkSmartPointer<vtkPolyData> poly_data_;
+  vtkSmartPointer<vtkPolyData> og_mesh_;
```
akenmorris, line 109: What's an OG Mesh?  OG = ?

Libs/Optimize/ParticleSystem/Sampler.h
```diff
@@ -548,6 +550,7 @@ class Sampler {
   std::vector<std::vector<SphereType>> m_Spheres;
   std::vector<FFCType> m_FFCs;
   std::vector<vtkSmartPointer<vtkPolyData>> m_meshes;
+  bool m_meshFFCMode;
```
akenmorris, line 553: Recommend default to `false`.  It looks like this would not be an issue for spreadsheet project as it will always call `SetMeshFFCMode`, however the XML input reader will only call this if the XML tag is present.  Otherwise the mode will be undefined and could either be on or off.

Libs/Optimize/ParticleSystem/TriMeshWrapper.h
```diff
@@ -52,6 +52,8 @@ class TriMeshWrapper : public MeshWrapper {
     return false;
   }
 
+  inline vtkSmartPointer<vtkPolyData> GetPolydata() const override {return vtkSmartPointer<vtkPolyData>();}
```
akenmorris, line 55: I think we need to delete TriMeshWrapper altogether.  Not for this PR/branch.

Libs/Optimize/ParticleSystem/VtkMeshWrapper.h
akenmorris: Original maybe?

Libs/Optimize/ParticleSystem/Sampler.cpp
```diff
@@ -295,7 +318,10 @@ void Sampler::AddMesh(std::shared_ptr<shapeworks::MeshWrapper> mesh) {
   if (mesh) {
     this->m_Spacing = 1;
     domain->SetMesh(mesh);
-    m_NeighborhoodList.back()->SetWeightingEnabled(!mesh->IsGeodesicsEnabled());  // disable weighting for geodesics
+//    vtkSmartPointer<vtkPolyData> polydataCopy = vtkSmartPointer<vtkPolyData>();
+//    polydataCopy->DeepCopy(mesh->GetPolydata());
```
akenmorris, line 322: Let's clean it up and remove dead commented code.

akenmorris (commented):
Looks good, just a couple more minor things.

Libs/Optimize/OptimizeParameters.cpp
akenmorris: This is implemented in `Variant::operator bool()` here:

https://github.com/SCIInstitute/ShapeWorks/blob/0ca7bde21d3089b2ec7737154fb7dfd565c73ff3/Libs/Project/Variant.cpp#L16

It should accept 0/1, true/false, True/False, etc.

HeavenlyBerserker:
@akenmorris I'm not sure why it's failing on Windows. Any idea?

akenmorris:
Looks good, I will figure out what's wrong with pip/conda on windows.  It's nothing to do with this branch.

Issue #1739, opened 2022-04-14
Studio: Landmark placement sensitive to scrolling 


patkins:
When using the place landmark tool in Studio, if you have more shapes than are currently shown on the screen, when you try to scroll up/down (with your mouse wheel or on screen) to view additional shapes, the landmark placement tool stops working or works intermittently (sometimes even moves landmarks of the initially viewed shapes and removes the visualization of the landmark). The shapes that were not visible prior to scrolling also sometimes turn red after scrolling (for the black background and gray shapes color scheme on linux, this did not happen on windows, but it could have been chance).  If you adjust the viewer to show all shapes, you can place landmarks and see all landmarks as expected.

patkins:
FYI The red shapes also occurs on windows.

akenmorris:
I think this is two separate issues.  Can you tell me more about the red shapes?  Are you displaying a scalar field on them, or are the regular ones just "orange" or whatever your default color is?

patkins:
The shapes are just the default color. I did a bit more fiddling with this and it seems that the red shapes are the ones that aren't visualized after scrolling. For this example, I started with 16 shapes visualized, added a new landmark, scrolled down one mouse click, then zoomed out twice. If you don't scroll after adding a landmark, the shapes don't turn red. Interestingly, if I scroll one click up from what is shown, the 'red' shapes are the shapes located in the location of the red shapes in this image (but include a different set of shapes). I can make a video if that's helpful, just let me know.
![image](https://user-images.githubusercontent.com/13870137/164285722-ead45295-8caf-491b-8b7e-15a67b7fe653.png)

patkins:
You have to add a landmark, but you don't have to place any yet. I've uploaded a quick video to /home/sci/atkins/forAlan that shows what is happening (note that at second 12 or 13, I scroll down one click on my mouse, which is important in recreating this). Hope this helps.

akenmorris:
Thanks!  The video made all the difference.  For whatever reason (haven't discovered yet), it only does this when opening the new shapes from within Studio like this.  When I open them from the command line as I usually do (e.g. studio *.nrrd), I don't get this behavior.  Thanks again!

akenmorris:
Nevermind, I can do it on the command line itself, but I was doing about 600 shapes there and for some reason it doesn't happen then.  But the smaller selection I was doing from the file chooser triggered it and so do using a similar amount from the command line, as does loading them from a project file.

[...]

Issue #1735, opened 2022-04-13
Studio: Abort Optimization does not always seem to stop the optimization


patkins:
I think this only occurs if optimization parameters have not been changed (and potentially only when done quickly after the start of optimization), but when trying to understand where particles are initializing, I  hit the 'Abort Optimize' button as soon as particle optimization starts and then view where the first two particles are located. If done repeatedly, the particles seem to keep optimizing in the background (even if the project is reopened), such that if you press 'Run Optimize' it will start more than 2 particles, depending on the time that has passed.

akenmorris:
I'll look into it.  

I'd also like to replace the behavior of the Run Optimize to be more like the DeepSSM Run button that disables temporarily so that you don't accidentally double click it and cancel.

Issue #1734, opened 2022-04-11
Studio: Particle coloring is not removed when optimization is rerun


patkins:
When running a model in Studio, if you have used the particle coloring options (pressing 1 or 2 when hovered over a particle to show distances or running good/bad particle tool), the coloring is maintained based on particle number (I think) even after rerunning the optimization. It seems like this should be cleared with the 'Run Optimization' button.

akenmorris:
Agreed.

akenmorris:
It looks like this was completed in February.  Can you confirm in a recent development build?

patkins:
I am using a recent dev build and definitely could not remove the good/bad particle coloring, the particles would be colored normally until the same number of particles were on the shape and then would only be green and red, but wouldn't correspond to anything relevant.

akenmorris:
Ah, the good/bad.  I thought you meant the '1' and '2'.

patkins:
Yeah, sorry. I didn't verify that issue still existed. But I did have a weird pattern of colors for the '1' option in the latest dev build, which I assumed was related. I'll see if I can recreate it and add additional info.

patkins:
I think the oddity I noted yesterday using the '1' option (where the particle color map shows the distance to the chosen particle) may have been if a particle is chosen during optimization. In this scenario, only some particles are colored and the colormap doesn't seem to correlate to distance. (It's possible that it is not intended to function in this scenario, but I figured I would let you know.)

Issue #1674, opened 2022-03-02
Studio: Add ability to export montage of a mode of variation as a single image


akenmorris:
Related: 
* #254

akenmorris:
Completed.

PR #1742, opened 2022-04-21
Studio Updates

akenmorris:
This PR addresses:

* #254 
* #1674 
* #547

Also fixes for:

* #1739 
* #1735
* #1734 

image

image

image

Difference to mean (for any shape, e.g. sample, pca mode position):

image

https://user-images.githubusercontent.com/1693349/164534628-343867bc-e4af-43e0-bec1-aa814708bc60.mp4


iyerkrithika21:
@akenmorris: I am unable to export the montage. Not sure if this is a problem with my build or system. 

The montage is generated and I am able to change things in the window but when I say export, I see this error on the command line, and the image is not saved.
```
inotify_add_watch(/home/sci/iyerkrithika/ShapeWorks_new/Testing_PRS) failed: (No space left on device)

```
image

akenmorris:
Does it bring up a file chooser when you click `Export`?  

I assume your disk is not actually full?  (though it said `No space left on device`)

iyerkrithika21:
Other things seem great to me. I was able to test most of the issues this PR is addressing. 
Could not test the issue with landmark scrolling. What would be the correct way to test it?

iyerkrithika21:
> Does it bring up a file chooser when you click `Export`?
> 
> I assume your disk is not actually full? (though it said `No space left on device`)

Yes, the file/folder selection does show up. I can enter a name and click on export. 
The disk is not full.

akenmorris:
> Could not test the issue with landmark scrolling. What would be the correct way to test it?

It's a bit tricky.  I think it'd be best merge and get a dev build for Penny to try.

akenmorris:
Here's something about that inode message:

https://askubuntu.com/questions/1088272/inotify-add-watch-failed-no-space-left-on-device

`The error message doesn't actually refer to the amount of storage space, but to the amount of inotify watches that are available for a given file system (see also here: https://github.com/guard/listen/wiki/Increasing-the-amount-of-inotify-watchers).`

Seems very weird.  Are the other export options working?

iyerkrithika21:
I tried exporting PCA particles, and that worked.

iyerkrithika21:
Exporting PCA scores didn't work 
I saw the link you shared, and there they suggest some changes for which I need root access. 
I might have to contact the support to try that. Meanwhile, @jadie1, can you try exporting the montage? 
It looks like a problem with my system rather than the build.

akenmorris:
Can you try again, I've made some changes.  I think on Linux it wasn't getting a default extension and it didn't know how to save it.  I've also added a message either in the statusbar (success) or an error dialog.

I think that message you saw on the console is just warning and nothing to do with it not saving.

iyerkrithika21:
image
Yup, that fixed it!

iyerkrithika21:
@akenmorris: 
1. All the exports work correctly. Export current mesh and export mesh scalars fail for peanut shared boundary use case. 
image
This might be because we don't have any reconstruction for the contour domains.
2. For other datasets, the mesh isn't saved unless we provide an extension for export mesh. Is this intentionally done?
3. Mesh scalars also requires an extension in the name; it does save the scalar as a text file without extension.

akenmorris:
Yes, the issue with the peanut shared boundary use case is that we don't have a contour reconstruction method yet.  At the moment it's probably better to have the export fail than to save an empty mesh for that domain.

akenmorris:
Anything else needed to merge?  @iyerkrithika21 ?

iyerkrithika21:
> Anything else needed to merge? @iyerkrithika21 ?

Looks good! 
- Saving mesh and scalar are working as expected now. 
- Skipping groom for contour also works

jadie1 (approved):
Everything I tried worked as expected!
image

iyerkrithika21 (approved):

Issue #1080, opened 2021-03-02
Add equivalent of VectorImage::evaluate for the Image and Mesh classes


cchriste:
`VectorImage::evaluate` takes a point in space and returns the value of the image at that point by interpolating nearby values.
- [x] add function to Image
- [x] add similar function to Mesh to interpolate a given field at any point _on the mesh_ (use Mesh::closestPoint) to get the point to interpolate)

Incidentally, but not necessary for this issue, these two classes (Image and VectorImage) are extremely similar. Their functionality is specific to the type of data stored in their voxels, and they aren't both (currently) expected to work in the same places. We might be able to consolidate them using polymorphism or templates, or to make one utilize the other, or leave them separate, but still try to reduce the amount of copied code between the two. Note that they currently differ somewhat in interface and offered functionality, but for the most part may be similar enough to combine.

cchriste:
Closed #1053 as duplicate. Some example use cases noted there include:

volumetric correspondences (e.g. diffusion images, multiple clinically-relevant metrics - cortical thickness, bone density ... etc) - related to Andy's project

These notably encourage combining the classes. One interface to rule them all!

cchriste:
> Closed #1053 as duplicate. Some example use cases noted there include:

Actually leaving that issue open since it is more general. Currently tagged as won't fix/consolidation/would be nice.

archanasri:
We already have it for mesh as Mesh::probeVolume()

cchriste:
Look at the inheritance hierarchy here:
https://itk.org/Doxygen/html/classitk_1_1InterpolateImageFunction.html

There are several ways to interpolate the image such as Linear, NearestNeighbor, and Gaussian.
We can provide a simple version or an enum allowing the user to specify their preference. 

What will this be used for? My suggestion is Linear as the simplest intermediate value unless there is a specific use case needing to be resolved more delicately.

cchriste:
We have decided to use linear interpolation, same as used by VectorImage. 
We can add an enum in order to support other types of interpolation in the future.

akenmorris:
@sheryjoe , why do we need this?

sheryjoe:
@akenmorris mainly for volumetric correspondences and sampling/probing imaging data at particles locations (Andy's project).

akenmorris:
Done.

PR #1747, opened 2022-04-22
Image::evaluate

akenmorris:
Address Image::evaluate part of:

* #1080 

This adds Image::evaluate using a similar approach as was done for VectorImage with the advantage/difference that the interpolator is created only when first used.

Also cleaned up Mesh and Image classes a little bit.  There's a lot of formatting change for the most part.

### DIFF
--- a/Applications/shapeworks/SharedCommandData.h
+++ b/Applications/shapeworks/SharedCommandData.h
@@ -16,7 +16,7 @@ struct SharedCommandData
   ParticleSystem particleSystem;
   Field field;
 
-  bool validImage() const { return image.image != nullptr; }
+  bool validImage() const { return image.itk_image_ != nullptr; }
   bool validMesh() const { return mesh != nullptr; }
   bool validParticleSystem() const {return particleSystem.N() >= 1 && particleSystem.D() >= 1; }
 };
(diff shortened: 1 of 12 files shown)


HeavenlyBerserker (approved):
Looks to me like just a bunch of variable renaming mostly. Is there any reason "auto" is better than having the explicit type? I find the code more readable and easier to edit if I know the variable types of things without the extra step of looking at the function return type.

Looks good to me as long as it passes the checks. Anything in particular you want me to test or check? If not, it looks good.

akenmorris:
> Is there any reason "auto" is better than having the explicit type?

There are a lot of reasons:

https://subscription.packtpub.com/book/programming/9781786465184/1/ch01lvl1sec5/using-auto-whenever-possible

I don't use it all the time that it can be used.  If you look at the changes I've been making, they are typically changing something like this:

`vtkSmartPointer<vtkSmoothPolyDataFilter> smoother = vtkSmartPointer<vtkSmoothPolyDataFilter>::New();`

to

`auto smoother = vtkSmartPointer<vtkSmoothPolyDataFilter>::New();`

Some static analysis tools will recommend this type of change directly.  It's repeating the type in the same line and is simply much more concise.

akenmorris:
> Anything in particular you want me to test or check?

No, it's pretty simple.

Issue #1521, opened 2021-10-21
Automated verification of DeepSSM use case


akenmorris:
We are using shape evaluation metrics to verify regular ShapeWorks pipelines.  How can we automate the verification of DeepSSM?

jadie1:
Addressed in PR 1748

Issue #1409, opened 2021-08-03
DeepSSM Use Case  - adapt to include full pipeline


jadie1:
We need to adapt the DeepSSM use case to include the full pipeline:

- Define a train/test split
- Groom the train and test images using shape
- Optimize shapeworks on train set only
- Augment data using train set
- Split train into train and validation sets and train the model
- Test the model on the test set - use surface to surface distance to analyze

Issue #1161, opened 2021-03-17
Pytorch CUDA kernel failure


riddhishb:
Upon the installation of the shapeworks with cuda environment on CLIO and BEAST servers, they have the following specs
NVIDIA-SMI 460.32.03    Driver Version: 460.32.03    CUDA Version: 11.2  

The pytorch and torch vison installs correctly with following
```
>>> import torch
>>> import torchvision
>>> torch.cuda.is_available()
True
>>> torch.__version__
1.7.1+cu110
``` 
But the computations are not supported
If I run a simple test such as 

`python -c "import torch;print(torch.max(torch.rand((30,30),device='cuda')))"`

it gives me the following error trace

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
RuntimeError: CUDA error: no kernel image is available for execution on the device
```

This also happens with a new (non shapeworks) Conda environment using PyTorch 1.8.0 installation from the website (https://pytorch.org/).

I think there is an issue with the NVIDIA drivers and Cuda compatibility, but I cannot find a core reference to the same. This is a discussion about the same error but none of their solutions worked (https://github.com/pytorch/pytorch/issues/49161)
It might be something that we might need to take up with SCI support but I would like a better idea of such an error.

riddhishb:
@jadie1 Would it be possible for you to take a look at this? 

Thanks

riddhishb:
Ok Update on this! This is most definitely some sort of compatibility issue, 

I tried the same test on cibcgpu3 which has TITAN V

that has the following version

NVIDIA-SMI 460.32.03    Driver Version: 460.32.03    CUDA Version: 11.2

and it works well there.

I am going to relay this to support.sci to see if they have anything to say about this.

sheryjoe:
Is this still an issue? @riddhishb @jadie1

sheryjoe:
@jadie1 @riddhishb is this still an issue?

riddhishb:
I think this is fine to ignore --> this is only an issue for really old GPU's it works fine for CIBC.

sheryjoe:
Add to the online DeepSSM documentation the minimum cuda and GPU requirements. Add link to the cuda and GPU table.

sheryjoe:
https://docs.nvidia.com/deploy/cuda-compatibility/

akenmorris:
Not a shapeworks repo issue.

PR #1748, opened 2022-04-26
Full DeepSSM Use Case

jadie1:
This PR addresses issue #1409 by adapting the DeepSSM use case to do the full pipeline of creating training data and grooming validation and test images using image-to-image registration instead of relying on shape. 

The documentation for the use case ([deep-ssm-femur](http://sciinstitute.github.io/ShapeWorks/6.3/use-cases/deep-learning-based/deep-ssm-femur.html)) has been updated to reflect these changes.

Issue #1521 by adding verification to the tiny test, comparing the resulting surface-to-surface distance error. Full verification is not required for this use case as it takes a significant amount of time to run and the tiny test does not skip any step.

The doc has also been updated to address issue #1161.


akenmorris (approved):
Looks good, let's just double check that the tests now work after the merge.

Issue #1417, opened 2021-08-16
Add scalar value export option for particles


patkins:
In ShapeWorksStudio, add an option to export scalar values for all particles (current option for exporting scalars is based on reconstructed mesh nodes instead of particles).

akenmorris:
What scalar values are you interested in here?  From the feature map?

patkins:
Yeah, the scalar values from the feature map.

sheryjoe:
@akenmorris this is a use case for #1080

patkins:
This seems to export data for each mesh node instead of each particle. Additionally, if you try to overwrite the data you end up with additional data in the files (but it seems to be a random number of additional rows).

sheryjoe:
@akenmorris we can enable users to choose whether to export on particles or surface meshes. the latter option is useful of visualization purposes... etc

patkins:
I would agree that the export of scalar values on the 'current mesh' is definitely valuable.

akenmorris:
I've fixed "Export All Subjects Particle Scalars" to export the particles rather than the mesh vertices.

<img width="292" alt="image" src="https://user-images.githubusercontent.com/1693349/168332700-568c512e-14e5-4f58-8dda-23ae51e24c15.png">

Do we also need "Export All Subjects Mesh Scalars"?

patkins:
I wouldn't think so, since the meshes come in with scalars. But would there be a need for volumetric inputs?

akenmorris:
I don't know if it would be useful to anyone.  I guess I can probably add it without too much trouble.

akenmorris:
Added a separate issue for it, #1763

PR #1752, opened 2022-04-29
Address #1417 - Add scalar value export option for particles

akenmorris:
Addresses:

* #1417 

image

Example output:

image

This also enhances the mesh scalar output to support multi-component scalars.


patkins:
I haven’t tried this yet, but my first thought is whether the user has to do this for each shape or if this generates a file for each shape? It may be helpful to also have the current mesh option to export values for group means or PCA modes.

akenmorris:
At the moment, it's just the currently displayed shape.  We have an option to export particles along PCA modes and we could extend the scalar display to do that as well.

For group means, at the moment, you'd have to view each group mean and export them.  Do you want a menu option that is "Export Group Mean Particle Scalars" ?

akenmorris:
Do you want this feature as well, export a table of particles by shapes?  Only one scalar?

patkins:
A single export for all input shapes would be very useful for some of our projects, but the format of the export (whether there is a file per feature value with all shapes or a file per shape with all features) does not matter. I think the format you showed above would work well for the file per shape option too and then it would be consistent across the board.

akenmorris:
@patkins , I've added another export for all subjects, all scalars:

image

This will write a file for every subject as you described.

akenmorris:
@patkins , you can find builds here if you want to try the export:

http://www.sci.utah.edu/~amorris/1752/

PR #1766, opened 2022-05-13
DeepSSM updates

jadie1:
This PR makes the following updates to the DeepSSM use case:

- Improved parameters for optimization and neural network training
- Applied cutting plane clipping when comparing predicted and true meshes for more accurate surface-to-surface distance analysis
- Improved documentation and added snapshots


akenmorris (approved):
Looks good.  I tested both the tiny test and full use case.

PR #1821, opened 2022-07-14
Landmarks warping map amend

ben2k:
Invert Key/Value since multiple landmarks can have the same reference vertex id.


akenmorris:
Looks ok to me.  @nawazishkhan1-nk , let me know if you have any feedback.

I'm going to resubmit it as a separate PR so that the GitHub actions can run.  I can't see how to re-run yours with the GHA secrets required.  We should probably have @ben2k added to the project.

akenmorris (approved):
Looks good to me.

nawazishkhan1-nk (approved):
LGTM too :+1:

Issue #1817, opened 2022-07-05
Studio slow with large meshes and free form constraints


akenmorris:
https://shapeworks.discourse.group/t/v6-3-loading-newly-created-project-very-slow/78

I think we should investigate dropping the Djikstra stuff.

Related:

* #1801

akenmorris:
Fixed in 6.3.1

PR #1825, opened 2022-07-21
Remove shortest path

akenmorris:
Addresses:

* #1817

Long path to get here.  First I discovered that the dijkstra path was inserting points in the wrong order and skipping the last one.  This was generating some really odd paths for the clip.  After fixing this, I also added a check to see if the next point was a neighboring vertex (e.g. one edge away), if it is, then the dijkstra part can be skipped.  This state can be seen here:

98bf22c191f965f531e2747f142232adebb14b31

After that, I tried skipping the dijkstra part altogether.  All of the OptimizeTests still pass and it's much faster using with studio, so I can't see any reason to keep it there.


HeavenlyBerserker (approved):
This will help Studio in the patch but will test thoroughly over the following week.

PR #1833, opened 2022-08-05
OptimizeTests improvement: Converted all optimize tests to xlsx and fixed minor aug lag bug

HeavenlyBerserker:
This branch moves all optimize tests to the Testing/data/optimize directory. It also converted all xml optimize tests into xlsx. Made sure they all ran correctly after modifying OptimizeTests.cpp. 

I wrote a basic xml_to_xlsx python converter that uses shapeworks::project. I did not include this file, but I have it if anybody would like it included. It does the basic transition and the user can adjust the xlsx manually afterward for more challenging use cases.

Added infrastructure in the GaussSeidel optimizer to recognize whether it's in initialization mode or optimization mode. This will allow us to more easily test and manage correspondence fixes.

Found some bugs that Alan and I fixed, which made FFCs better and more correct. In particular, augmented lagrangian parameter mu is initialized to 0 now, instead of 1.

Fixed correspondence issues in constraint tests easily by observing particle movement after converting to xlsx.


akenmorris (approved):
Looks good.

PR #1835, opened 2022-08-06
MeshWarper: add Python method MeshWarper.getGoodParticlesIndices() to…

ben2k:
… return the indexes of good particles.


akenmorris (approved):
Looks fine to me.

Issue #1711, opened 2022-03-28
ShapeWorks / ITK interoperability


akenmorris:
This is mostly a documentation issue.

Document how to build ShapeWorks against an alternate version of ITK.

Provide examples of how to use C++ ShapeWorks API as part of another ITK program.

See also #1710

Issue #1710, opened 2022-03-28
ShapeWorks CMake Module and Example


akenmorris:
Update ShapeWorks build infrastructure to create an exportable CMake module so that a user can do:

find_package(ShapeWorks)

And use the C++ ShapeWorks API in their own C++ programs.

Provide some simple examples of how to use ShapeWorks C++ API (primarily optimization).

See also #1711

akenmorris:
Merged with #1889

PR #1889, opened 2022-10-03
Add example of using ShapeWorks as a third party library

akenmorris:
Adds everything necessary to `make install` ShapeWorks for use with `find_package(ShapeWorks)`.  Also provide an example and documentation.

Addresses:

* #1711 
* #1710 

Unfortunately, I had to move a lot of files around and restructure the `#include`s so that client code could use our headers in the installed version of shapeworks.


HeavenlyBerserker:
How do I test this? Can I make install it in place or do I need to install it to usr/local? Does it change the paths in my system?

The includes were not supposed to be that way anyways, so changing them is good. However, now it's more difficult to follow their source.

akenmorris:
I would use `-DCMAKE_INSTALL_PREFIX` to set the install to some temporary directory.  Don't install it to your system (don't use sudo).

HeavenlyBerserker:
I get the following error when running `(shapeworks) hxu@hxu-SYS-5038AD-I:~/Documents/ShapeWorks/build$ cmake -DCMAKE_CXX_FLAGS=-g -DCMAKE_PREFIX_PATH=/home/hxu/Documents/ShapeWorks/dependencies/install -DUSE_OPENMP=ON -DBuild_Studio=1  -Wno-dev -Wno-deprecated -DCMAKE_BUILD_TYPE=Release /home/hxu/Documents/ShapeWorks -DCMAKE_INSTALL_PREFIX=/home/hxu/Documents/shapeworks_install`

> CMake Error in dependencies/install/geometry-central/src/CMakeLists.txt:
>   Target "geometry-central" INTERFACE_INCLUDE_DIRECTORIES property contains
>   path:
> 
>     "/home/hxu/Documents/ShapeWorks/dependencies/install/geometry-central/src/../include"
> 
>   which is prefixed in the source directory.

The whole log is here
[itk_example error.txt](https://github.com/SCIInstitute/ShapeWorks/files/9728275/itk_example.error.txt)

akenmorris:
@HeavenlyBerserker , should be ready to try again.

Issue #1787, opened 2022-06-21
Multi-level PCA Analysis


nawazishkhan1-nk:
Integrate Multi-level PCA Analysis work to ShapeWorks

- [x] Studio + Backend Analyze
- [ ] Use-Cases + Tests
- [x] Documentation

akenmorris:
Completed and merged.

PR #1929, opened 2022-11-04
Studio and Backend for Multi-Level Analyze

nawazishkhan1-nk:
#1787


akenmorris:
Some feedback, (and I can help fix some of these issues):


https://user-images.githubusercontent.com/1693349/201018116-855f7d0a-63fe-4793-bd1d-277401594415.mp4


1. The text layout is bouncing around as the animation goes back and forth.

2. The animation slider timers aren't turning off with the multi-level tab.  Notice that in the 'master' branch studio, the PCA animation timer turns off if you leave the PCA page.  The multi-level one doesn't turn off and the PCA one also doesn't turn off when going to the multi-level page.

cut-off

3. This may only be on MacOS, but the "Group" tab is now cut off.

4. It would be great to see more extensive git commit messages.  Also the PR description should ideally describe what's being added/changed in the PR. 

5. This is more of a general Studio criticism that we should address with another issue, but when I load a project without multiple domains, the Multi-Level Analysis tab is disabled, but as a naive user, I have no idea why.  I think we should consider adding tooltips to anything that would be greyed out explaining what type of data is needed to enable it.

nawazishkhan1-nk:
Closed with #1962

PR #1934, opened 2022-11-14
Fix #1933 - Conda environment

### LINKED ISSUE #1933: install_shapeworks.sh : Unable to specify alternate conda env when using bash shell
`source install_shapeworks.sh testenv` results in a conda env named "python=3.9.13" instead of "testenv".

Works with zsh, but fails with bash.

akenmorris:
Fixes:

* #1933 

Array access was working for zsh for some reason, but not bash.


iyerkrithika21 (approved):
Creates a new env correctly now

Issue #1790, opened 2022-06-22
Incremental Shape Models


nawazishkhan1-nk:
- Build use cases for incremental shape models where shapes are added incrementally one by one by finding the one closest to existing distribution represented by current trained model.

PR #1936, opened 2022-11-15
Incremental use case update

jadie1:
This PR addresses issue #1790 and #1745.

It adds options to determine the order via random sorting, closest to median sorting, and closest to distribution sorting. 
It also adds options to specify the initial model size and incremental batch size.
The documentation has been updated appropriately. 

To test run:
```
python RunUseCase.py incremental_supershapes
```


jadie1:
The use case test is failing because I added interaction to the use case. It requires input for the sorting method, initial model size, and batch size. I did this instead of adding parameters to RunUseCase.py because these options are super specific to this use case and all of the other RunUseCase.py tags are general.

@akenmorris and @nawazishkhan1-nk Do you think it would be better to add these three tags to RunUseCase.py or hard code them in the incremental_supershapes.py and add comments and documentation that clarify that these can be changed?

akenmorris:
I would suggest that we use the "--interactive" flag that we've used in the past.  When it's specified, the script can ask for user input, otherwise it should use some defaults.

jadie1:
This should be set to merge :) @akenmorris @nawazishkhan1-nk

akenmorris (approved):
Looks great.

PR #1962, opened 2022-12-02
Multi-Level Analysis (Studio + Backend)

nawazishkhan1-nk:
This PR addresses issue #1787. 
It adds the Multi-Level Component Analysis feature (Backend + Studio) to ShapeWorks. 
- Studio changes reflected in PCA Panel of Analysis Tab
Screen Shot 2022-12-02 at 10 36 10 AM

- This is visible in Studio only if a multiple-domain model is loaded
- Backend changes done in `ParticleShapeStatistics.cpp`
- Documentation for this added feature in Analysis Tab: `docs/studio/studio-analyze.md`


Libs/Particles/ParticleShapeStatistics.cpp
```diff
@@ -208,6 +208,154 @@ int ParticleShapeStatistics::ImportPoints(std::vector<Eigen::VectorXd> points, s
   return 0;
 }
 
+//---------------------------------------------------------------------------
+void ParticleShapeStatistics::SetNumberOfParticlesAr(std::vector<int> num_particles_ar)
```
akenmorris, line 212: What does "Ar" stand for?

nawazishkhan1-nk: It means 'Array'. This holds the number of particles for each domain which is needed to compute multi-level statistics. I just added this suffix so that it won't be confused with num_particles already been used.

Libs/Particles/ParticleShapeStatistics.cpp
```diff
@@ -208,6 +208,154 @@ int ParticleShapeStatistics::ImportPoints(std::vector<Eigen::VectorXd> points, s
...
+//---------------------------------------------------------------------------
+void ParticleShapeStatistics::SetNumberOfParticlesAr(std::vector<int> num_particles_ar)
+{
+  this->m_num_particles_ar = num_particles_ar;
+}
+
+//---------------------------------------------------------------------------
+int ParticleShapeStatistics::ComputeMultiLevelAnalysisStatistics(std::vector<Eigen::VectorXd> points, unsigned int dps)
+{
+  unsigned int num_points = (int)(points[0].size() / (3 * dps)); // M_total
```
akenmorris, line 220: num_points is computed, but not used.

Studio/src/Analysis/AnalysisTool.cpp
```diff
@@ -65,6 +66,11 @@ AnalysisTool::AnalysisTool(Preferences& prefs) : preferences_(prefs) {
   connect(ui_->pcaAnimateCheckBox, SIGNAL(stateChanged(int)), this, SLOT(handle_pca_animate_state_changed()));
   connect(&pca_animate_timer_, SIGNAL(timeout()), this, SLOT(handle_pca_timer()));
 
+  //MCA animation  
+  connect(ui_->mcaLevelBetweenButton, SIGNAL(clicked()), this, SLOT(mca_between_radio_toggled()));
```
akenmorris, line 70: I know there's a lot in the code that haven't been updated, but we should prefer the newer style Qt signals/slots that are compile time checked vs runtime.

Libs/Particles/ParticleShapeStatistics.cpp
```diff
@@ -208,6 +208,154 @@ int ParticleShapeStatistics::ImportPoints(std::vector<Eigen::VectorXd> points, s
...
+      z_rel_pose_objective(k * VDimension, i) = z_rel_pose_centred(k, i * VDimension);
+      z_rel_pose_objective(k * VDimension + 1, i) = z_rel_pose_centred(k, i * VDimension + 1);
+      z_rel_pose_objective(k * VDimension + 2, i) = z_rel_pose_centred(k, i * VDimension + 2);
+    }
+  }
+  m_pointsMinusMean_for_rel_pose.resize(m_dps * VDimension, m_N);
+  m_pointsMinusMean_for_rel_pose.fill(0.0);
+  m_mean_rel_pose = z_rel_pose_objective.rowwise().mean();
+  m_pointsMinusMean_for_rel_pose = z_rel_pose_objective.colwise() - m_mean_rel_pose;
+}
```
akenmorris, line 297: warning: non-void function does not return a value [-Wreturn-type]

Studio/src/Analysis/AnalysisTool.cpp
```diff
@@ -724,6 +875,22 @@ void AnalysisTool::handle_group_animate_state_changed() {
   }
 }
 
+//---------------------------------------------------------------------------
+int AnalysisTool::get_mca_level() {
+  bool vanilla_pca = ui_->vanillaPCAButton->isChecked();
+  bool between = ui_->mcaLevelBetweenButton->isChecked();
+  bool within = ui_->mcaLevelWithinButton->isChecked();
+  if (vanilla_pca){
+    return 1;
```
akenmorris, line 884: We should really be using an enum rather than magic numbers, 1, 2, 3

Studio/src/Analysis/AnalysisTool.cpp
```diff
@@ -144,6 +154,14 @@ std::vector<Shape::Point> AnalysisTool::get_group_difference_vectors() {
   }
   return vecs;
 }
+//---------------------------------------------------------------------------
+void AnalysisTool::mca_vanilla_pca_radio_toggled() { Q_EMIT mca_update(); }
```
akenmorris: One nice thing with Qt is that signals can be directly connected to signals.  So we needn't write slots that just emit another signal.

Studio/src/Interface/ShapeWorksStudioApp.cpp
```diff
@@ -1754,6 +1771,18 @@ void ShapeWorksStudioApp::compute_mode_shape() {
   visualizer_->display_shape(analysis_tool_->get_mode_shape(pca_mode, pca_value));
 }
 
+//---------------------------------------------------------------------------
+void ShapeWorksStudioApp::compute_mca_mode_shape() {
```
akenmorris, line 1775: It seems redundant to have `compute_mca_mode_shape` and `compute_mode_shape` as the former will do what the latter does when the mode is `vanilla`.

akenmorris (changes requested):
I've mostly address the comments that I've made, but I added them anyway to help explain the changes.

I'd like to merge the pca/mca stuff in ShapeWorksStudioApp class further, but I also don't even like how the old code asks the AnalysisTool for the mode and value and then passes them back in.  The AnalysisTool should be handling all that.  I will restructure this.

PR #1980, opened 2023-01-09
Fixes for FFC

akenmorris:
Modifies `more_boundaries_debug` branch in several ways

* Merge master
* Fix isosurface generation back to 0 (zero level set)
* Replace flood fill algorithm
* Invert in/out usage for FFC
* Change query point choosing to disallow boundary points
* Fix boundary contour to 1.0 instead of 0.5 for reproducible boundaries

To reduce the PR diff, merge master into `more_boundaries_debug`


HeavenlyBerserker (commented):
Ok

Issue #1984, opened 2023-01-14
Implement in/out field constraint storage option


akenmorris:
Current boundary/query method does not support topologies where after an individual boundary cut, the query point is both inside and outside of the cut such as when a chunk of a donut is removed.

Flood fill techniques seem to work, but are fraught with problematic edge cases.

To solve this, we will store the in/out painted field directly which will support any topology and also provide exact reproducibility.  The only downside is some additional disk space usage, but it should be fairly minimal.

akenmorris:
Merged with #1986

PR #1986, opened 2023-01-20
FFC field definition

akenmorris:
Addresses

* #1984 

Previous constraints files are converted to the new format

New constraints also writes the old boundary/query which previous ShapeWorks can use (provided it would have worked before, e.g. not donut).

image(8)


HeavenlyBerserker (approved):
Looks good to me. I tried all the constraints stuff and optimize tests. I tried the femur and ellipsoid Studio use cases and even used the mesh_ffc_mode to enable aug lag for meshes. Everything seems to work just fine.

As a QoL improvement, we should display some informative text of what's going on when constraints are being processed after clicking optimize.

PR #2013, opened 2023-03-07
Refactor Optimize Library [after 6.4 release]

### LINKED ISSUE #160: rename 207 ITK-prefixed files to no longer have that prefix since they are not part of ITK

akenmorris:
Major refactoring of the entire Optimize library.  Got rid of lots of templates, removed the itk prefix and namespace, organized into folders.  Lots of other stuff.  Should resolve #160

There should be no behavior changes.


HeavenlyBerserker (approved):
Looks good to me. I went over all the constraint tests visually and can't see much difference from Master. It seems like correspondence issues persist in both.

PR #2036, opened 2023-04-03
Fix 2033 mesh to dt

### LINKED ISSUE #2033: Mesh to DT produces artifacts in some cases
The current algorithm for determining a point being inside or outside a mesh relies on a dot product with the normal of the closest face.  While this is fast, it can be incorrect in some cases:

image

image

The distance values are correct, but the sign is wrong.

We should try this instead:

https://vtk.org/doc/nightly/html/classvtkSelectEnclosedPoints.html

Example data attached:

[problemDTConv.zip](https://github.com/SCIInstitute/ShapeWorks/files/11093288/problemDTConv.zip)

akenmorris:
Fix #2033 by using vtkSelectEnclosedPoints to determine if points are inside or outside.

* #2033 

image

Also parallelized mesh to DT, patch bugs in VTK


HeavenlyBerserker (approved):
Looks good to me. Works for the dataset I first discovered the issue on. Can confirm that it is a lot faster too.

Doesn't matter that is not completely robust. I think it's good enough.

akenmorris:
> Doesn't matter that is not completely robust. I think it's good enough.

It turns out that it is simply returning a different, but still correct cell.  Quite often the closest point on a mesh is at a vertex, and any cell that the vertex belongs to would be valid.  The SMP enabled versions can simply turn different cells.

Screen Shot 2023-04-15 at 8 05 34 PM

Screen Shot 2023-04-15 at 8 04 17 PM

Issue #2055, opened 2023-04-27
Spatiotemporal SSM - ShapeWorks Studio Support


nawazishkhan1-nk:
### Feature Summary
Add support (Backend optimization as well as Studio) for the disentangled optimization technique for spatiotemporal statistical shape modeling. ([Referenced Paper](https://arxiv.org/abs/2209.02736))

PR #2056, opened 2023-04-27
Spatiotemporal SSM - Disentangled Approach

nawazishkhan1-nk:
[WIP] - Do Not Merge

## Summary
This PR addresses issue #2055.

## Changes Made
- Non-breaking change: Added a new class in Libs/Optimize/Function  implementing the correspondence function for disentangled spatiotemporal SSM.
- Made relevant additions (initialization, etc) to Sampler, Optimize and OptimizeParameter
- Added a checkbox in the ShapeWorks Studio Optimize panel and made appropriate connections to use the method in the optimization step. 

## Testing Done
- Tested the new optimization technique on a sample multi-domain project where each domain is assumed to be a separate time point

## Known Issues
- None at this time.

## Screenshots
Screen Shot 2023-04-27 at 7 51 15 AM


akenmorris:
1. We need a small automated test case using this (e.g. OptimizeTests).

2. It looks like the checkbox is visible and enabled (can be toggled) whether it's a multi-domain model or not.  This may be confusing for novice users as we don't want to overwhelm them with options that aren't relevant to their data.

I think it's fine as a first step to just have only this optimization setting, but in the future we may want more of a data level option for indicating that these are time points rather than joint anatomies.  This could enable other visualization modes (e.g. add a time point selector, or separate time points along the X axis).  We could do this either with a flag that the user provides or by extending the input types (e.g. instead of `shape_<name>` column, we have `shape_<name>_time_0`, `shape_<name>_time_1` and then when these are recognized, the spatiotemporal options become visible, or enabled.  @sheryjoe

akenmorris:
Can the checkbox be called "Spatiotemporal SSM"?  I'm not sure the "disentangled" word adds much understanding.

akenmorris:
@nawazishkhan1-nk , do you have an update on the PR?

nawazishkhan1-nk:
> @nawazishkhan1-nk , do you have an update on the PR?

I think this PR is good to go considering the backend optimization. The only feature that has to be worked on is exposing a better way to input time points for each subject in the Data view of Studio.

akenmorris (changes requested):
1. Can you add a small automated test to OptimizeTests that uses this feature?  Otherwise, I am concerned that future changes/refactoring could break this feature and we would never know.

2. The checkbox in Studio should only be enabled if the data is applicable.  The tooltip should also indicate this.  If you show a checkbox to a user, expect them to ask you what it does.

3. mkdocs documentation update with this feature?

akenmorris:
Be advised that I had to convert the inverse covariance matrix from VNL to Eigen:

https://github.com/SCIInstitute/ShapeWorks/pull/2118

akenmorris:
@nawazishkhan1-nk , I've fixed this branch for the changes in #2118.

Can you point me to a (hopefully small) example so that I confirm the result are unchanged (and to confirm future refactoring)?

akenmorris:
I am merging this now so that I don't have to spend a bunch of time resolving conflicts as I make changes to these same files.  I will create a separate issue for the automated test.

Issue #1136, opened 2021-03-10
DeepSSM: Data Augmentation Parallelization


jadie1:
User should be able to specify how many processes to use in deep_ssm use case. We also need to verify this works as expected on Windows.

akenmorris:
For Studio (embedded python), some info on the challenges here:

https://stackoverflow.com/questions/60328557/using-python3-with-pybind11-from-c-with-multiprocessing-process?noredirect=1&lq=1

jadie1:
Pushing to 6.3 as this will require refactoring for Studio

akenmorris:
@jadie1, what's the status on this issue?

jadie1:
We should resolve this for the next stage of DeepSSM in Studio #1887

jadie1:
@akenmorris Does PR #2089 resolve this?

akenmorris:
> @akenmorris Does PR #2089 resolve this?

Partly in that you can use the multiprocessing package now within Studio.  The fix was only needed for the pytorch dataloader though (as far as I know).  Were we already using multiprocessing for the data augmentation?

jadie1:
We have multiprocessing for data augmentation implemented; it parallelizes the image generation part to speed it up in `Python/DataAugmentationUtilsPackage/DataAugmentationUtils/DataAugmentation.py.` It just uses the Python multiprocessing lib, but if I remember correctly, it was throwing an error on Windows.

akenmorris:
Ok, it should work then.  I will check on windows as well.

akenmorris:
I've fixed this and enabled parallel data augmentation in Studio in #2093

jadie1:
@akenmorris Can we close this issue?

akenmorris:
Yep.  Merged in #2093

PR #2093, opened 2023-06-13
Fix windows python multiprocessing

akenmorris:
Fix Python multiprocessing on windows.

Also enable parallel data augmentation in Studio (see #1136)


jadie1 (approved):
I tested the data augmentation multiprocessing in studio and python and it is working. I just added one fix for a typo in the documentation for the DeepSSM in Studio example.

PR #2129, opened 2023-08-03
1881 deepssm implement variants and losses

zahidemon:
Main changes:

1. Added **TL-Net** as model architecture and associated training and validation functions.
2. Added **focal loss** as loss function options.
3. Added different **learning rate scheduler** options.
4. Added documentation for the newly added variants.


zahidemon:
To test the added variants in the DeepSSM use case, please make the following changes in the config file:

1. Set the `config["loss"]["function"]` parameter as the loss function to train the model. Available options are: `MSE` and `Focal.` Default value: `MSE`
2. Set the `config["tl_net"]["enabled"]` parameter as True to train the model using TL-Net. If the value is False, the model will be trained with Vanilla/Base DeepSSM. Default value: `False`

@akenmorris @jadie1

akenmorris (approved):
I tested this and while it took a long time to run, it worked.

akenmorris:
@jadie1 , any comments?

jadie1 (approved):
Looks good and ran for me. I added a couple of lines to send the PCA scores to the device. This ensures that training with PCA MSE loss on a GPU does not throw an error.

PR #2139, opened 2023-08-24
Free Form Constraint propogation to other shapes

### LINKED ISSUE #1517: Reduce constraint annotation burden
Currently, we assume the user defines sample-wise cutting planes or free form constraints for optimization. We could reduce this burden by allowing the user to define constraints on a representative subset of the data then appropriately transforming those constraints for each sample. Would need a use case to demonstrate this.

akenmorris:
* #1517 

Resolve #1517 by adding an option to copy free form constraints from one shape to the others using the alignment transforms.

Right click on the FFC in the table and select "Copy constraint to other shapes":

Screen Shot 2023-08-24 at 12 45 46 PM

The free form constraint is applied to the other shapes:

Screen Shot 2023-08-24 at 12 45 58 PM


HeavenlyBerserker (approved):
Looks good to me. Tested it on femurs and ellipsoids, including painting multiple areas on multiple shapes. All looks to be functional!

PR #2154, opened 2023-10-21
Global reference domain for offline analysis

akenmorris:
This PR fixes the alignment reference domain in the offline analysis.  Previously, the computed shapes would overlap:

Screen Shot 2023-10-21 at 12 30 14 AM

Now we use the global alignment:

Screen Shot 2023-10-21 at 12 37 36 AM

This PR also fixes the output location of the mean shape and particles (source vs target)


JakeWags (approved):
Looks good. Seems like it'd be simple enough to change the Alignment type based on a command argument if we decide to go that path.

Issue #1887, opened 2022-10-02
DeepSSM in Studio Phase 2


jadie1:
Phase 2 is adding training optimization and image grooming to Studio for DeepSSM. 
Currently to run DeepSSM in Studio you load groomed images and particle pairs. Ideally we'd like to start with ungroomed images and shapes and create the training and testing data. This involves:

- Optimizing training particles
- Optimizing testing particles using fixed domains
- Grooming images
   - Resampling
   - Applying transforms (shape grooming alignment and procrustes if used in training optimization)
   - Cropping (using bounding box that fits world particles plus some padding)

For more detail see the [DeepSSM use case](http://sciinstitute.github.io/ShapeWorks/latest/use-cases/deep-learning-based/deep-ssm-femur.html), but note in Phase 2 we will skip step 7 and prep the testing images in the same way as the training. In Phase 3 we will do true inference and groom testing images using image-to-image registration.

zahidemon:
@akenmorris, I have been talking to @jadie1 about this issue. We already have the Python codes for this phase 2. We just need to figure out how to incorporate them into the studio. We can discuss this more next week in the dev meeting and a separate meeting if you'd like to discuss this earlier than that.

jadie1:
Based on today's discussion, here is a tentative plan:

Define split before opening studio (use case steps 1 and 2)
1. Load training data into studio - data tab
2. Groom training meshes - groom tab (use case step 3)
3. Optimize training particles - optimize tab (use case step 4) 
4. Analyze training particles - analyze tab
5. Load validation data - fixed domain tab 
6. Groom validation meshes using training ref- fixed domain tab (use case step 8)
7. Optimize validation particles using fixed domains - fixed domain tab
8. Train and validation image grooming step 1: apply mesh alignment transforms to images (reflection, alignment, Procrustes)  - image grooming tab or deepssm tab (use case step 5)
9.  Train and validation image grooming step 2: find bounding box using train meshes +transforms, use it to crop all images - image grooming tab or deepssm tab (use case step 5) 
10. Run data augmentation using training images and world particles - deepssm tab (use case step 6)
11. Create data loaders - deepssm tab (use case step 9)
12. Train DeepSSM - deepssm tab (use case step 10)
13. Predict validation particles and analyze accuracy (use case step 11)

[...]

Issue #977, opened 2021-02-16
DeepSSM in Studio


sheryjoe:
Parent issue for including DeepSSM in Studio. 

- [x] Phase 1: Similar to current python deep_ssm use case, images are assumed to be pre-groomed, test shapes are optimized with training.

> - [x] #1371
> - [x] #1372
> - [x] #1373
> - [x] #1374
> - [x] #1375

- [x] Phase 1b: #1881 Integrating variants and loss that were worked on for the journal extension.

- [x] Phase 2: #1887 Image grooming within studio, test shapes are held out from optimization and optimized separately in a fixed domain model

- [x] Phase 3: True inference, new images are registered against groomed image model

akenmorris:
Phase 1 is complete. Moving this issue to 6.3 for phase 2.

jadie1:
Just updating the use case in 6.3, moving the next phase of DeepSSM in Studio to 6.4

akenmorris:
Done #2181

PR #2181, opened 2024-01-04
Studio DeepSSM Phase 2/3

akenmorris:
* #1887
* #977

Refactor deep_ssm.py script into DeepSSMUtils, switch to using ShapeWorks Project files directly for all steps, allowing it to be used more easily by Studio.  The results are similar to before but not identical because some steps are a little bit different.  For example, Grooming using ShapeWorks Groom library uses center of mass rather than just the mesh bounding box center.

Implement DeepSSM Phase 2 and 3 in Studio.  Single prep button performs all steps except augmentation, training, test.  Does not require that any grooming or optimization has been done.

image

I had started by replicating all the steps from the phase 2 deep_ssm.py script, but then realized I would either need to integrate C++ Elastix into ShapeWorks, or call out to Python for that step.  At that point I realized it would be a better idea to simply call out to Python for almost everything.  This necessitated refactoring deep_ssm.py script into DeepSSMUtils using ShapeWorks project objects rather than lists of filenames.  It utilizes the new `exclude` functionality in ShapeWorks projects to mark the validation and test subjects as excluded for the first grooming and optimization steps.  Then the training data is marked as `fixed` and the validation subjects are marked un-excluded for grooming and optimization.  The image grooming steps are mostly the same, but now read/write from the project file.  I added the image registration transforms as extra columns in the table for later use.

This approach should allow the Python, Studio and Cloud versions of DeepSSM to share the vast majority of code.

This PR also supports Phase 3 by allowing subjects to have only an image.  These subjects will be assigned as test separately from the train/val/test split assignment.

You can also add images to a completed DeepSSM model and manual mark them as `test` and re-run only the testing phase.  The new images just won't have a distance map or average error (as there is no ground truth mesh)

image


akenmorris:
@jadie1 @zahidemon @JakeWags , the docs have been updated with the new LA dataset.  Have a look.  I believe this is ready to merge.

JakeWags (approved):
Use case and implementation look good

zahidemon (approved):
Ran the whole pipeline without any errors. The results look good within a decent runtime.

Issue #2318, opened 2024-08-27
DeepSSM Image-based Registration Problems


iyerkrithika21:
There are few issues to the image-based registration (because we are assuming  we don't have access to ground truth surfaces during inference) process followed for evaluation: 
1. Function: `get_image_registration_transform`
- Translation transform crashes here: `itk_transform = SimpleITK.TranslationTransform()`
- The code should be changed to:
```
    # Get transform matrix
    parameter_map = result_transform_parameters.GetParameterMap(0)
    transform_params = np.array(parameter_map['TransformParameters'], dtype=float)
    if transform_type == 'rigid':
        itk_transform = SimpleITK.Euler3DTransform()
    elif transform_type == 'similarity':
        itk_transform = SimpleITK.Similarity3DTransform()
    elif transform_type!= 'tranlation':
        print("Error: " + transform_type + " transform unimplemented.")
    itk_transform.SetParameters(transform_params)
    itk_transform_matrix = np.eye(4)
    itk_transform_matrix[:3,:3] = np.array(itk_transform.GetMatrix()).reshape(3,3)
    if transform_type != 'translation':
        itk_transform_matrix[-1,:3] = np.array(itk_transform.GetTranslation())
   else:
       itk_transform_matrix[-1,:3] = transform_params.T
```
2. Should we expose other parameters for image registration to the users? This registration might require some parameter tuning for other complex anatomies. 

3. 'rigid' is hardcoded [here](https://github.com/SCIInstitute/ShapeWorks/blob/1aab2271e1b8ed09a821d3ef1ca02f40886fe381/Python/DeepSSMUtilsPackage/DeepSSMUtils/__init__.py#L72), so it always applies 'rigid' transformation even when others transform types are passed. 

4. As per ITK documentation:
- Rigid (itk::Euler3DTransform) applies a rotation and translation to the space given 3 euler angles and a 3D translation. Rotation is about a user specified center.
- Similarity (itk::Similarity3DTransform) applies a rotation, translation and isotropic scaling to the space
[...]

PR #2319, opened 2024-08-27
2318 image registration

akenmorris:
* #2318 

Fix translation transform
Fix rigid transform being hard-coded


iyerkrithika21 (approved):
Thanks, Alan! 
This looks good!

